Issue #19, opened 2015-08-04
Setup client tests for all platforms


inikulin:
We need to add Mac Safari, iOS Safari, Android browsers

inikulin:
ref #5

churkin:
@miherlosev is already working on this

inikulin:
I just see that the code was fixed, but CI env was not setup for this platforms (also, I don't see mentions of the MacOS Safari). So I'll keep this issue open for now, till we'll not have this things done.

miherlosev:
\cc @inikulin 
I think you can close issue.
Was added 
IPhone (IOS 7.1), Mac Safari 8.0, IPad 8.1
Android 5.1 (Sauce labs not contains image for Android 5.0)

inikulin:
Ok. Great job though.

PR #32, opened 2015-08-11
add safari 8.0 and higher sauce labs images for client tests

miherlosev:
cc @inikulin @churkin 
Part of #19 

Change qunit server ports from 1335, 1336 to 2000,2001.
Safari 8.0 + allows only specified ports for localhost connecting .
For more details see - https://docs.saucelabs.com/reference/sauce-connect/#can-i-access-applications-on-localhost-


inikulin:
From what I see changing port in the sever tests wasn't a requirement, but LGTM anyway

churkin:
lgtm

PR #40, opened 2015-08-14
src/client/utils/service.js refactoring (close #31)

### LINKED ISSUE #31: src/client/utils/service.js is misleading
I think it has a really misleading name. If it's just a bunch of generic helper funcs then let's create separate files for them, like we do for the server code.

miherlosev:


inikulin:
lgtm

PR #41, opened 2015-08-15
Client tests moved to gulp-qunit-harness

AlexanderMoskovkin:
I've moved client tests to [gulp-qunit-harness](https://github.com/AlexanderMoskovkin/gulp-qunit-harness).
The main benefits:
1) Get rid of all qunit-server stuff (server, templates, magic dirs like 'data', 'sandbox') from the project
2) Get rid of Saucelabs setup scripts
3) Run qunit with a gulp module

All of the test files should have the `-test.js` suffix like in the server tests.
Please see [Usage](https://github.com/AlexanderMoskovkin/gulp-qunit-harness#usage) and [QUnit tests](https://github.com/AlexanderMoskovkin/qunit-harness#qunit-tests) articles before review.

/cc @inikulin @miherlosev @churkin @LavrovArtem @helen-dikareva @AndreyBelym what do you think about this? We will use this in other our repositories with qunit-testing.

p.s. I suppose Saucelabs iOS tests will be failed with this commit, i'll fix it


AlexanderMoskovkin:
FPR

Gulpfile.js
```diff
@@ -106,32 +106,30 @@ gulp.task('server-tests', ['rebuild'], function () {
         }));
 });
 
-gulp.task('client-tests', ['rebuild'], function () {
-    gulp.watch(['./src/**', './test/client/fixtures/**'], ['rebuild']);
 
-    require('./test/client/server.js').start();
+var CLIENT_TESTS_SETTINGS = {
```
inikulin: I think we should move all settings object to the top of the file

test/client/before-test.js
```diff
@@ -1,4 +1,7 @@
 (function () {
+    $('script').addClass('script-TC2b9a6d');
+    $('link').addClass('ui-stylesheet-TC2b9a6d');
```
inikulin: I'm not sure if this correct, we make all scripts and links on page 'shadowed'. I think it's much better to explicitly specify which scripts and links we would like to hide.

test/client/fixtures/sandboxes/cookie-test.js
```diff
@@ -84,7 +84,7 @@ test('get/set', function () {
 asyncTest('path validation', function () {
     var iframe = document.createElement('iframe');
 
-    iframe.src = UrlUtil.getProxyUrl('/data/cookie-sandbox/validation.html');
+    iframe.src = UrlUtil.getProxyUrl(window.QUnitGlobals.getResourceUrl('../../data/cookie-sandbox/validation.html', 'cookie-sandbox/validation.html'));
```
inikulin: I don't get it, what second parameter of the `window.QUnitGlobals.getResourceUrl` do?

inikulin: It's also better to introduce var here to avoid this long line.

inikulin:
\r-

test/client/fixtures/sandboxes/cookie-test.js
AlexanderMoskovkin: see [here](https://github.com/AlexanderMoskovkin/qunit-harness#get-test-resource). It's neccessary for the cookies test. We need an iframe with url like `.../cookie-sandbox/iframe.html`. For me the best way is use:
`window.QUnitGlobals.getResourceUrl(<fileName>, 'cookie-sandbox/iframe.html')`. With the second parameter we set the resource url

test/client/before-test.js
AlexanderMoskovkin: We 'hide' all of the 'service' scripts/css on the page before hammerhead initialized, otherwise hammerhead process `qunit-harness` resources (qunit, jquery etc). I think this is correct, because there are no other scripts on the page in this moment (scripts that we should not 'shadow').

We can specify the scripts and links but we will have to correct that  when some changes in the qunit-harness implemented

test/client/fixtures/sandboxes/cookie-test.js
inikulin: So, the second parameter is the URL alias, right? But what is the default URL then (without specifying this alias) ?

test/client/before-test.js
inikulin: Let's add the comment which will clarify this behavior then. It's not quite clear currently.

test/client/fixtures/sandboxes/cookie-test.js
AlexanderMoskovkin: yes.
The default is:
`<hostname>/test-resource?filePath=<filePath>&base=<curTestFilePath>`
In my example it will be:
`<hostname>/test-resource/cookie-sandbox/iframe.html?filePath=<filePath>&base=<curTestFilePath>`

test/client/before-test.js
AlexanderMoskovkin: ok

AlexanderMoskovkin:
FPR
@inikulin we've described the getResourceUrl function in the [api topic](https://github.com/AlexanderMoskovkin/qunit-harness#get-test-resource)

AlexanderMoskovkin:
FPR (merged with upstream)

inikulin:
\r- I've asked for rebase, not merge. Now we have huge unnecessary diffs

AlexanderMoskovkin:
It was necessary to merge with upstream. There was a commit to the upstream repo with qunit tests changes after this pr was opened. I had to merge this with upstream and make some changes to fix qunit tests. 
Where do you see huge unnecessary diffs? I open the [Files changed 55](https://github.com/superroma/testcafe-hammerhead/pull/41/files) tab in this pr and see only my changes to move qunit to gulp-qunit-harness

AlexanderMoskovkin:
FPR

inikulin:
LGTM

AlexanderMoskovkin:
thanks

PR #45, opened 2015-08-18
IE11 throws an error when the second parameter of the Blob function is undefined (fix #44)

### LINKED ISSUE #44: IE11 throws an error when the second parameter of the Blob function is undefined
The issue occurs because we override the Blob function. If the overridden function is called with one parameter, it calls the original function with two parameters where the second parameter is undefined.

```javascript
new Blob(["a"])
```
Text of the error: "Could not complete the operation due to error 800a138f."

LavrovArtem:
/cc @inikulin @miherlosev

### DIFF
--- a/src/client/sandboxes/dom/window.js
+++ b/src/client/sandboxes/dom/window.js
@@ -113,14 +113,24 @@ export function override (window, overrideNewElement) {
 
     if (window.Blob) {
         window.Blob = function (parts, opts) {
-            var type = opts && opts.type && opts.type.toString().toLowerCase();
+            // NOTE: IE11 throws an error when the second parameter of the Blob function is undefined (GH-44)
+            // If the overridden function is called with one parameter,
+            // then we need to call the original function with one parameter as well.
+            switch (arguments.length) {
+                case 0:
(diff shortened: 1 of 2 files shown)


LavrovArtem:
FPR

inikulin:
It requires rebase now

LavrovArtem:
FPR

LavrovArtem:
FPR

LavrovArtem:
FPR

inikulin:
LGTM

inikulin:
iOS tests timeouted

AlexanderMoskovkin:
:white_check_mark: Tests for the commit b157b2ecfa1711c060b0c51c221859ebb3b87058 have passed. See [details](https://travis-ci.org/testcafe-build-bot/testcafe-hammerhead/builds/76113515).

Tests are ok now. We've found the issue with travis and saucelabs and will fix it in the gulp-qunit-harness. The problem appears when more than one travis tasks started in the one time.
So, @inikulin @miherlosev you can review the pr

inikulin:
LGTM then

PR #2, opened 2015-07-31
add hammerhead src to repository

churkin:
\cc @inikulin @LavrovArtem @miherlosev

Issue #1, opened 2015-07-30
Location wrapper returns case-sensitivity port, hostname and protocol in the href field (T272552) 


churkin:
Proxied page - `http://localhost:1401/!2/http://Example.com/`.

`location.href` should returns `http://example.com` instead of `http://Example.com`

inikulin:
Is this behavior consistent across browsers? Can't find it in [URL spec] (https://url.spec.whatwg.org/#dom-urlutils-hostname) (`Location` interface is inherited from `URLUtils` per [living DOM spec]()https://html.spec.whatwg.org/multipage/browsers.html#location).

churkin:
Browser sends requests with host, port, protocol in the lower case

inikulin:
All browsers?

churkin:
I have checked IE, FF, Chrome

inikulin:
OK then

inikulin:
fixed via 76d52329aa79b460df3f46288fdb21cfa16d4df2

PR #49, opened 2015-08-19
Active window watching mechanism implemented

helen-dikareva:
It's necessary to fix the T235186 TestCafe bug

/cc @miherlosev @LavrovArtem @inikulin 

@VasilyStrelyaev please check the note in the src/client/sandboxes/event/active-window.js file


VasilyStrelyaev:
lgtm

helen-dikareva:
FPR

miherlosev:
I think need test's coveradge for the following scenarious:
1)initially active is iframe #1 window then activate the iframe #2 window
2)initially active is iframe window then activate the top window (activate by clicking on shadow ui element).

AlexanderMoskovkin:
>I think we need tests that cover the following scenarious:
> initially active is iframe...

The top window is active always while the page is loading and our scripts initializing. What the reason of testing scenarios when an iframe is active at test start?

> (activate by clicking on shadow ui element)
This tests - unit tests for the active-window module. The Shadow UI has no relation to this

AlexanderMoskovkin:
ok, we have discussed it personally. We'll add two tests:
- for the 'active-window' module to test the scenario focus iframe1 -> focus iframe2
- for the focus-blur module to check that activeWindow.set is not called when an iframe is focused and focus called for the shadow ui element

src/client/sandboxes/event/active-window.js
```diff
@@ -0,0 +1,61 @@
...
+            }, activeWindow);
+        }
+        catch (err) {
+            //NOTE: the error appears in IE when the corresponding iframe is removed
+            return void 0;
+        }
+    }
+}
+
+export function init () {
```
inikulin: Since we have init() function I think it should be better to call this ActiveWindowTracker and make it class instead of set of functions (moreover it has internal state)

inikulin:
What's going on here?

helen-dikareva:
PR closed due to wrong merging

PR #55, opened 2015-08-25
Fix an incorrectly overridden CanvasRenderingContext2D.drawImage function

### LINKED ISSUE #54: The CanvasRenderingContext2D.drawImage function is overridden in a wrong way
The function does not take into account that the first parameter can be not only an `` element, but also a `<canvas>` or `<video>`. [Specification](https://html.spec.whatwg.org/multipage/scripting.html#dom-context-2d-drawimage).

LavrovArtem:
/сс @inikulin @AlexanderMoskovkin

### DIFF
--- a/src/client/sandboxes/dom/window.js
+++ b/src/client/sandboxes/dom/window.js
@@ -1,6 +1,6 @@
 /*global history, navigator*/
 import { isMozilla } from '../../utils/browser';
-import { isCrossDomainWindows } from '../../utils/dom';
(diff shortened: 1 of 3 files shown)


inikulin:
lgtm overall, but I think we should rename commit. Now it's looks like you are introducing the wrong override, instead of fixing it =)

LavrovArtem:
FPR

AlexanderMoskovkin:
lgtm

PR #73, opened 2015-09-02
eslint: added "no-undef" rule

LavrovArtem:

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,6 +23,7 @@
     "space-before-blocks": [2, "always"],
     "prefer-const": 2,
     "no-path-concat": 2,
+    "no-undef": 2,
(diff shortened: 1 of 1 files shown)


AlexanderMoskovkin:
Please wait a little bit. We've found an issue with eslint after updating to the new eslint version, @AndreyBelym will create an issue soon. Maybe we'll use a common solution

AndreyBelym:
OK, I've submitted an issue #74

AlexanderMoskovkin:
@LavrovArtem See https://github.com/superroma/testcafe-hammerhead/issues/74
We should not fix just `no-undef` rule because we've lost a lot of others. I suppose we can close this pr and @AndreyBelym will fix this

PR #84, opened 2015-09-09
The 'Illegal invocation' error after svg element focused (close #82)

### LINKED ISSUE #82: The "Illegal invocation" error after svg element focused
The issue is reproduced in Google Chrome version >=45 only.

Add the following markup on the page:
```html
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1">
    <rect id="rect" width="300" height="300" fill="red" tabIndex="1"></rect>
</svg>

<script>
    var rect = document.getElementById('rect');

    rect.focus();
    alert(document.activeElement.id);
</script>
```

**Expected**: the `alert` dialog with text `"rect"`
**Actual in the hammerhead playground**: the `Uncaught TypeError: Illegal invocation` error

In fact the error appears because of the following:
```js
var NativeMethods = {
    focus: document.createElement('div').focus
};

var rect = document.getElementById('rect');
NativeMethods.focus.call(rect);    //throws an error
```

churkin:
\cc @AlexanderMoskovkin @miherlosev @inikulin

### DIFF
--- a/src/client/sandboxes/dom/dom.js
+++ b/src/client/sandboxes/dom/dom.js
@@ -167,7 +167,7 @@ function onDocumentCleaned (window, document) {
             NativeMethods.refreshDocumentMeths(document);
 
         if (needToUpdateNativeElementMeths)
-            NativeMethods.refreshElementMeths(document);
+            NativeMethods.refreshElementMeths(document, window);
 
         // T239109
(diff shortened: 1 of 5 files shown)


AlexanderMoskovkin:
lgtm

miherlosev:
lgtm

PR #120, opened 2015-09-21
Use three-character unique value as a session id (close #116)

### LINKED ISSUE #116: Use three-character unique value as a session id

churkin:
\cc @inikulin @LavrovArtem

### DIFF
--- a/src/session.js
+++ b/src/session.js
@@ -3,16 +3,12 @@ import UploadStorage from './upload/storage';
 import COMMAND from './command';
 import mustache from 'mustache';
 import read from './utils/read-file-relative';
+import uuid from 'node-uuid';
 import { EventEmitter } from 'events';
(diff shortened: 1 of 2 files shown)


LavrovArtem:
lgtm

inikulin:
lgtm

src/session.js
```diff
@@ -30,6 +26,11 @@ export default class Session extends EventEmitter {
         };
     }
 
+    static _generateSessionId () {
+        // NOTE: #116 (https://github.com/superroma/testcafe-hammerhead/issues/116)
+        return uuid.v4().substr(0, 3);
```
inikulin: @churkin forgot to mention: do not use full link in comments, because we will move from the superroma account.
Use `testcafe-hammerhead#16` format instead

Issue #112, opened 2015-09-17
Transform instance-level constants to the ctor properties


churkin:
`constructor () {`
`this.UNCAUGHT_JS_ERROR = 'uncaughtJSError';`

to

`WindowSandbox.UNCAUGHT_JS_ERROR_EVENT = 'uncaughtJSError';`

inikulin:
It's will be not in proto in this case, but in ctor. I'll edit the title

churkin:
We decided not to implement it because of the inconvenience of use

PR #121, opened 2015-09-22
Get all constant values to a identical type, use static methods (close #111)

### LINKED ISSUE #111: Get all constant values to a identical type

churkin:
\сс @inikulin @LavrovArtem @miherlosev


src/const.js
```diff
@@ -5,24 +5,22 @@
...
 
-CONST.UPLOAD_SANDBOX_HIDDEN_INPUT_NAME = 'upload-89fc3-info';
+CONST.UPLOAD_SANDBOX_HIDDEN_INPUT_NAME = '@hammerhead|upload-info-hidden-input@';
 
-CONST.HOVER_PSEUDO_CLASS_ATTR = 'data-1b082a6cec-hover';
+CONST.HOVER_PSEUDO_CLASS_ATTR = 'data-hammerhead-hovered';
 
 CONST.XHR_CORS_SUPPORTED_FLAG   = 0x10;
-CONST.XHR_REQUEST_MARKER_HEADER = 'x-tc-xm-cd46977f';
+CONST.XHR_REQUEST_MARKER_HEADER = '%hammerhead|xhr-request-marker-header%';
```
inikulin: Why not `@` prefix and postfix here?

src/const.js
```diff
@@ -5,24 +5,22 @@
 
 var CONST = {};
 
-CONST.DOM_SANDBOX_OVERRIDE_DOM_METHOD_NAME = 'tc_odm_e01a2ff5';
-CONST.DOM_SANDBOX_PROCESSED_CONTEXT        = 'tc-spc-c1208da2';
-CONST.DOM_SANDBOX_STORED_ATTR_POSTFIX      = '_e01a2f_stored';
-CONST.EVENT_SANDBOX_WHICH_PROPERTY_WRAPPER = 'tc-e01a2f-which';
+CONST.DOM_SANDBOX_OVERRIDE_DOM_METHOD_NAME = '@hammerhead-override-dom@';
```
inikulin: What is the purpose of this constant? It's naming is quite unclear

inikulin: And why it not follows format `@hammerhead|...@`?

src/const.js
churkin: We can not use '%' because the const values used in the url attributes and can not be decoded/ecoded.

inikulin: You mean `@`? It's used in HTTP header name and not URL AFAIK and I see that `@` is disallowed there: http://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2
We can get rid of prefix and postfixes then and just keep name it `hammerhead|xhr-request-marker-header`. What do you think?

src/const.js
churkin: uses in the regex

inikulin: @churkin What do you mean?

inikulin:
tests fail

inikulin:
BTW what about #112? Will this PR close it as well?

churkin:
I'll close #112. We decided not to do it because of the inconvenience of use

inikulin:
ok

inikulin:
Requires rebase now

inikulin:
@churkin Seems like it really hangs up in IE 10 and 9

churkin:
FPR

miherlosev:
lgtm

LavrovArtem:
lgtm

inikulin:
lgtm

PR #128, opened 2015-09-25
fix iframe proxing (top level window uses AMD module loader) (fix #127)

### LINKED ISSUE #127: 'almond' AMD module loader crashes jsParsingTools module inside Iframe without src
'almond' AMD module loader - https://github.com/jrburke/almond

Page for reproduce without 'almond':
```javascript
<html>
	<head>
		<!--<script src="https://raw.githubusercontent.com/jrburke/almond/latest/almond.js"></script>-->
		<script>
			var define;
			define = function () {};
	    	define.amd = {}
		</script>
	</head>
	<body>
		<h1>Test</h1>
		<script>
			iframe.id = "iframe";
		</script>
	</body>
</html>
```

After opening page in playground I get the script error:
'Uncaught TypeError: Cannot read property 'parse' of undefined'

miherlosev:
/cc @inikulin @churkin 
The test name was checked by @MargaritaLoseva

### DIFF
--- a/test/client/fixtures/sandbox/iframe-test.js
+++ b/test/client/fixtures/sandbox/iframe-test.js
@@ -145,3 +145,20 @@ asyncTest('ready to init event must not raise for added iframe(B239643)', functi
     }, 100);
 });
 
+asyncTest('the AMD module loader damages proxing an iframe without src (GH-127)', function () {
+    var amdModuleLoaderMock = function () {};
+
+    amdModuleLoaderMock.amd = {};
+    window.define = amdModuleLoaderMock;
+
+    var iframe = document.createElement('iframe');
+
+    iframe.id = 'test_iframe_unique_id_jlsuie56598o';
+    iframe.addEventListener('load', function () {
+        ok(this.contentWindow.Hammerhead);
+        delete window.define;
+        iframe.parentNode.removeChild(iframe);
+        start();
+    });
+    document.body.appendChild(iframe);
+});
--- a/src/processing/js/parsing-tools.js
+++ b/src/processing/js/parsing-tools.js
@@ -30,12 +30,6 @@ var acornExports = {};
 //
 // [dammit]: acorn_loose.js
 // [walk]: util/walk.js
-
-    (function (root, mod) {
-        if (typeof exports == "object" && typeof module == "object") return mod(exports); // CommonJS
-        if (typeof define == "function" && define.amd) return define(["exports"], mod); // AMD
-        mod(root.acorn || (root.acorn = {})); // Plain browser env
-    })(this, function (exports) {
         "use strict";
 
         exports.version = "0.4.1";
@@ -1915,10 +1909,9 @@ var acornExports = {};
             return finishNode(node, "Identifier");
         }
 
-    });
-}).call(acornExports);
+    })(acornExports);
 
-jsParsingTools.parse = acornExports.parse || acornExports.acorn.parse;
+jsParsingTools.parse = acornExports.parse;
 // ------------------------------------------
 
 // --------------- Code Gen -----------------


test/client/fixtures/sandbox/iframe-test.js
```diff
@@ -43,6 +43,24 @@ test('event should not raise if a cross-domain iframe is appended', function ()
     $iframe.remove();
 });
 
+asyncTest('the AMD module loader damages proxing an iframe without src', function () {
```
inikulin: It should be in `regression` section and test title should inlclude github issue number (in format `(GH-{num})`)

inikulin:
It has so much insertions/deletions because of reformatting, right?

\r-

miherlosev:
>>It has so much insertions/deletions because of reformatting, right?
Yes

miherlosev:
Commit contains fix for unstable test (test/client/fixtures/sandbox/event/focus-blur-change-test.js)
FPR

test/client/fixtures/sandbox/event/focus-blur-change-test.js
```diff
@@ -770,12 +770,22 @@ if (window.HTMLInputElement.prototype.setSelectionRange) {
...
+                    attemptCount++;
+                    var isInputFocused = document.activeElement === input2;
+
+                    if (focusRaised === needFocus && isInputFocused === needFocus ) {
+                        expect(0);
+                        window.clearInterval(intervalId);
+                        startNext();
+                    }
+                    else if (attemptCount > maxAttemptCount)
+                        ok(false, 'The check for calling focus failed');
```
inikulin: @VasilyStrelyaev Check spelling pls

VasilyStrelyaev: The check actually succeeds - it's the result that is unexpected, so maybe:
'The element's focus state doesn't match the expected one.'

miherlosev:
FRP
/cc @LavrovArtem

inikulin:
Maybe we can wrap this code with attempts into some kind of function in `test-utils` or smthg like this, so it can be reused later?

inikulin:
Let's discuss it on Monday with @AlexanderMoskovkin and @churkin

miherlosev:
ping @AlexanderMoskovkin , @churkin

AlexanderMoskovkin:
You can implement this in the `before-test.js` file or create the `test-utils.js` file and add it to the `CLIENT_TESTS_SETTINGS` in the `Gulpfile.js`.
I suppose we can use this function in the following way:
```js
//the util function
window.checkWithInterval = function (predicate, done, interval, timeout) {
...
}

//test
asyncTest('test', function () {
    function checkSomething () {
        return a === b && c === d;
    }
    function done (success) {
        if(!success)
            ok(false, 'our check are failed');
    
        start();
    }
    
    window.checkWithInterval(checkSomethink, done, 100, 1000);
});
```

As an alternative we can implement some build-in function in the `qunit-harness`. Something like:
```js
//test
asyncTest('test', function () {
    window.QUnitGlobals.assert(function (ok, notOk, equal, notEqual) {
         ok(someValue);
    }, interval, timeout);
});
```
In this case qunit-harness can automatically make assertions during timeout and finish the test if assertions successed or timeout expired

inikulin:
@AlexanderMoskovkin What I don't like in current implementation is what we don't use assertions, so it's not clear what checked in the file. The good implementation will be to add `attempts` parameter to the test and restart test on qunit level if it fails (some kind of quarantine mode but for qunit). I suppose we should modify qunit-harness

AlexanderMoskovkin:
Created an issue for the `qunit-harness` repo: https://github.com/AlexanderMoskovkin/qunit-harness/issues/6

inikulin:
@miherlosev Let's keep things as is right now, but create separate issue for switching tests you modified to the AlexanderMoskovkin/qunit-harness#6 mechanism once it will be ready.

miherlosev:
Commit contains only bug fix.
Separate issue for unstable tests - https://github.com/superroma/testcafe-hammerhead/issues/136

miherlosev:
FPR

miherlosev:
ping @inikulin @churkin

churkin:
Restore the original `parsing-tools.js` code format please

churkin:
lgtm

inikulin:
lgtm

PR #132, opened 2015-09-25
Add comment (close #115)

### LINKED ISSUE #115: How uploading works on the Linux/MacOS platforms?
In the Hammerhead we are using windows-specific code:

`const FAKE_PATH_STRING         = 'C:\\fakepath\\';`

We should investigate how it works on other platforms.

LavrovArtem:


inikulin:
Seems like bot is still broken \cc @AlexanderMoskovkin

AlexanderMoskovkin:
The bot is ok. It's the `saucelabs environment error` on some virtual machines. It's a problem on the saucelabs side, we can't fix it

inikulin:
@AlexanderMoskovkin But, shouldn't bot restart tests in that case?

AlexanderMoskovkin:
Failed tests were restarted three times.

inikulin:
Which delay was useb beetween restarts?

AlexanderMoskovkin:
It was about 6 minutes between each rerun. If it's necessary we can increase the delay or reruns count when the saucelabs environment error appeared

inikulin:
Yeah, I think we should use progressive delays as well.

AlexanderMoskovkin:
ok

inikulin:
lgtm

churkin:
lgtm

PR #144, opened 2015-09-30
Hammerhead should remove service data from local storage on the first session page load (close #100)

### LINKED ISSUE #100: Hammerhead should remove service data from local storage on the first session page load
The client `Transport` module saves failed service messages to the localStorage and resend them on the next page load. The localStorage key is the session ID.
We have a problem with the following situation:
1) Start the app and create a hammerhead `Session` (with `id` = 1)
2) Some service message was stored
3) The application process was finished (the next page load did not occur)
4) Start a new application and create a new hammerhead `Session` (it have the same `id` = 1)
5) As a result on the first page load `Transport` get the stored service messages by the `Session` id and  resend them.
I suppose, hammerhead should clear all previous service data from local storage on the first session page load

/cc @churkin @inikulin

churkin:
\cc @AlexanderMoskovkin @inikulin @LavrovArtem

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -95,6 +95,10 @@ class Hammerhead {
         }
     }
 
+    static _cleanLocalStorageServiceData (sessionId, window) {
+        window.localStorage.removeItem(sessionId);
+    }
+
     on (evtName, handler) {
         var eventOwner = this._getEventOwner(evtName);
 
@@ -110,10 +114,16 @@ class Hammerhead {
     }
 
     start (initSettings, win) {
-        if (initSettings)
+        win = win || window;
+
+        if (initSettings) {
             settings.set(initSettings);
 
-        this.sandbox.attach(win || window);
+            if (initSettings.isFirstPageLoad)
+                Hammerhead._cleanLocalStorageServiceData(initSettings.sessionId, win);
+        }
+
+        this.sandbox.attach(win);
     }
 }
 
--- a/src/client/task.js.mustache
+++ b/src/client/task.js.mustache
@@ -2,6 +2,7 @@ window.Hammerhead.start({
     cookie                   : '{{{cookie}}}',
     sessionId                : '{{{sessionId}}}',
     serviceMsgUrl            : '{{{serviceMsgUrl}}}',
+    isFirstPageLoad          : {{{isFirstPageLoad}}},
     ie9FileReaderShimUrl     : '{{{ie9FileReaderShimUrl}}}',
     crossDomainProxyPort     : '{{{crossDomainPort}}}',
     referer                  : '{{{referer}}}'
--- a/src/session.js
+++ b/src/session.js
@@ -17,10 +17,11 @@ export default class Session extends EventEmitter {
 
         this.uploadStorage = new UploadStorage(uploadStoragePath);
 
-        this.id         = Session._generateSessionId();
-        this.cookies    = new Cookies();
-        this.proxy      = null;
-        this.injectable = {
+        this.id            = Session._generateSessionId();
+        this.cookies       = new Cookies();
+        this.proxy         = null;
+        this.pageLoadCount = 0;
+        this.injectable    = {
             scripts: ['/hammerhead.js'],
             styles:  []
         };
@@ -46,15 +47,20 @@ export default class Session extends EventEmitter {
         if (withPayload)
             payloadScript = isIFrame ? this._getIFramePayloadScript() : this._getPayloadScript();
 
-        return mustache.render(TASK_TEMPLATE, {
+        var taskScript = mustache.render(TASK_TEMPLATE, {
             cookie:               cookies.replace(/'/g, "\\'"),
             sessionId:            this.id,
+            isFirstPageLoad:      this.pageLoadCount === 0,
             serviceMsgUrl:        serverInfo.domain + '/messaging',
             ie9FileReaderShimUrl: serverInfo.domain + '/ie9-file-reader-shim',
             crossDomainPort:      serverInfo.crossDomainPort,
             payloadScript:        payloadScript,
             referer:              referer
         });
+
+        this.pageLoadCount++;
+
+        return taskScript;
     }
 
     _getIFramePayloadScript () {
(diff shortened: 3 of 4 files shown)


src/client/index.js
```diff
@@ -92,6 +92,10 @@ class Hammerhead {
         }
     }
 
+    _resetServiceData (sessionId, window) {
```
inikulin: Method sould be `static` `_resetServiceData`->`cleanLocalStorageServiceData`

src/client/task.js.mustache
```diff
@@ -2,6 +2,7 @@ window.Hammerhead.start({
     cookie                   : '{{{cookie}}}',
     sessionId                : '{{{sessionId}}}',
     serviceMsgUrl            : '{{{serviceMsgUrl}}}',
+    needToResetServiceData   : {{{needToResetServiceData}}},
```
inikulin: `needToResetServiceData` -> `isFirstPageLoad`

src/session.js
```diff
@@ -17,10 +17,11 @@ export default class Session extends EventEmitter {
...
         this.uploadStorage = new UploadStorage(uploadStoragePath);
 
-        this.id         = Session._generateSessionId();
-        this.cookies    = new Cookies();
-        this.proxy      = null;
-        this.injectable = {
+        this.id               = Session._generateSessionId();
+        this.cookies          = new Cookies();
+        this.proxy            = null;
+        this.firstPageRequest = true;
```
inikulin: You can call it `pageLoadCount`

inikulin: And make it a number

src/session.js
```diff
@@ -40,20 +41,24 @@ export default class Session extends EventEmitter {
     }
 
     getTaskScript (referer, cookieUrl, serverInfo, isIFrame, withPayload) {
-        var cookies       = this.cookies.getClientString(cookieUrl);
-        var payloadScript = '';
+        var cookies          = this.cookies.getClientString(cookieUrl);
+        var firstPageRequest = this.firstPageRequest;
+        var payloadScript    = '';
+
+        this.firstPageRequest = false;
```
inikulin: Just increment `pageLoadCount` here

src/session.js
```diff
@@ -40,20 +41,24 @@ export default class Session extends EventEmitter {
...
-            cookie:               cookies.replace(/'/g, "\\'"),
-            sessionId:            this.id,
-            serviceMsgUrl:        serverInfo.domain + '/messaging',
-            ie9FileReaderShimUrl: serverInfo.domain + '/ie9-file-reader-shim',
-            crossDomainPort:      serverInfo.crossDomainPort,
-            payloadScript:        payloadScript,
-            referer:              referer
+            cookie:                 cookies.replace(/'/g, "\\'"),
+            sessionId:              this.id,
+            needToResetServiceData: firstPageRequest,
```
inikulin: And here you will get:
```js
isFirstPageLoad: this.pageLoadCount === 0
```

test/client/fixtures/transport-test.js
```diff
@@ -11,7 +11,8 @@ var requestIsAsync = false;
 Settings.get().serviceMsgUrl = '/service-msg/100';
 
 function reqisterAfterAjaxSendHook (callback) {
-    callback = callback || function () {};
+    callback = callback || function () {
+        };
```
inikulin: Formatting broken here

inikulin:
\r-

churkin:
FPR

test/client/fixtures/transport-test.js
```diff
@@ -81,7 +81,7 @@ asyncTest('queuedAsyncServiceMsg', function () {
             var expectedCompleteMsgReqs = [10, 500, 200, 300, 200];
 
             deepEqual(completeMsgReqs, expectedCompleteMsgReqs);
-            Transport.asyncServiceMsg = savedAsyncServiceMsgFunc;
+            Transport.asyncServiceMsg   = savedAsyncServiceMsgFunc;
```
inikulin: Unnecessary extra space. Seems like it's formatting bug. Just add a newline to avoid it (actually it's always better to separate assertions from the other code blocks as well as variable assignments)

src/session.js
```diff
@@ -49,6 +50,7 @@ export default class Session extends EventEmitter {
         return mustache.render(TASK_TEMPLATE, {
             cookie:               cookies.replace(/'/g, "\\'"),
             sessionId:            this.id,
+            isFirstPageLoad:      this.pageLoadCount++ === 0,
```
inikulin: Avoid using unary operators which modify state within binary expressions (I wonder if we have eslint rule for this). So it should be:
```js
this.pageLoadCount++;
...
isFirstPageLoad: this.pageLoadCount === 1
``` 
Actually, it's my bad, AFAIR it was my suggestion to use comparison with 0. This variant looks semantically correct.

inikulin:
\r-

churkin:
FPR

src/session.js
```diff
@@ -46,15 +47,20 @@ export default class Session extends EventEmitter {
...
             sessionId:            this.id,
+            isFirstPageLoad:      this.pageLoadCount === 0,
             serviceMsgUrl:        serverInfo.domain + '/messaging',
             ie9FileReaderShimUrl: serverInfo.domain + '/ie9-file-reader-shim',
             crossDomainPort:      serverInfo.crossDomainPort,
             payloadScript:        payloadScript,
             referer:              referer
         });
+
+        this.pageLoadCount++;
```
inikulin: This is the completely opposite to what I've suggested in previous comment: at the first increment `this.pageLoadCount`, then compare it with `1` not zero.

inikulin: It semantically correct and help avoid introducing additional `taskScript` var

inikulin:
\r-

src/session.js
AlexanderMoskovkin: As for me, the `isFirstPageLoad` variable means that there were no page loads in this session, and the `this.pageLoadCount === 0` check is ok for me

inikulin: ok then, let's keep it as is

inikulin:
lgtm

AlexanderMoskovkin:
lgtm

LavrovArtem:
lgtm

PR #145, opened 2015-09-30
Consider extending EventEmitter in sources (close #95)

### LINKED ISSUE #95: Consider extending EventEmitter in sources
Currently we use `EventEmitter` as class field with the wrapped methods, e.g.: https://github.com/superroma/testcafe-hammerhead/blob/master/src/client/sandbox/base.js

I think we should extend `EventEmitter` in this cases instead. 
\cc @AlexanderMoskovkin I think we need to do this is `testcafe` as well.

BTW currently `src/client/utils/event-emitter` using ES5 syntax, use ES6 classes instead (review other parts of code and check if we have old syntax somewhere else).

churkin:
\cc @AlexanderMoskovkin @inikulin @LavrovArtem

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -1,6 +1,7 @@
 /*eslint-disable no-native-reassign*/
 import Sandbox from './sandbox';
 import CodeInstrumentation from './sandbox/code-instrumentation';
+import EventEmitter from './utils/event-emitter';
 import settings from './settings';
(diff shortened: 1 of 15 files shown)


AlexanderMoskovkin:
lgtm

inikulin:
lgtm

LavrovArtem:
lgtm

PR #146, opened 2015-09-30
Remove module.exports from client code (close #130)

### LINKED ISSUE #130: module.exports in client code
https://github.com/superroma/testcafe-hammerhead/blob/master/src/client/utils/browser.js#L37 - I'm not sure if this works. Anyway, we should get rid of this approach, which replaces exports.

LavrovArtem:


src/client/index.js
```diff
@@ -38,7 +38,9 @@ class Hammerhead {
         // Methods
         this.getOriginElementAttributes = CodeInstrumentation.getAttributesProperty;
         this.upload                     = this.sandbox.upload.upload;
-        this.get                        = require;
+
+        // Used in client tests
```
inikulin: Wrong. It's used in tests, we need to describe how it used as well. Discuss text with @VasilyStrelyaev

src/client/utils/browser.js
```diff
@@ -22,40 +22,56 @@ function calculateBrowserAndVersion (userAgent) {
     };
 }
 
+
+export var isIE11;
```
inikulin: I doubt user agent information changes for the different windows of the **same** browser. Why can't we just get rid of `init` and `export default` object with all this properties?

inikulin:
\r-

miherlosev:
I think we need rename 'IsMozilla' -> 'IsFirefox'.
Because Mozilla is a company name, Firefox is a browser name.

inikulin:
@miherlosev +1

src/client/utils/browser.js
churkin: \cc @helen-dikareva

helen-dikareva: we can do it only one time for browser and init() is redundant

churkin: It looks like it is used for the tests. I propose to pass window through the constructor

churkin: `class BrowserUtils  {`
`constructor(window) { ... }`
`}`
...
`export default new BrowserUtils(window);`

miherlosev: > we can do it only one time for browser and init() is redundant @miherlosev

It is used in TestCafe. Later I will provide more details.

miherlosev: TestCafe-phoenix requires initialized BrowserUtil before Hammerhead.init() was called.
https://github.com/superroma/testcafe-phoenix/blob/master/src/client/core/util/text-selection.js#L131

inikulin: @churkin I don't like what we turn pure utilities into the class object just because of tests, it's this kind of thing which nearly shouts in your face: "We need to rewrite those tests! They screw up our production code!"

inikulin: @miherlosev So? Isn't this is an argument for the utility object which is initialized on `require`?

miherlosev: 1)remove from browser.js `init();`
2)add to Hammerhead constructor browserUtil.init(window)

browserUtils doesn't initialized on `require`

inikulin: > 2)add to Hammerhead constructor browserUtil.init(window)
WHY???

inikulin:
Any update on this?

src/client/index.js
```diff
@@ -38,7 +38,10 @@ class Hammerhead {
         // Methods
         this.getOriginElementAttributes = CodeInstrumentation.getAttributesProperty;
         this.upload                     = this.sandbox.upload.upload;
-        this.get                        = require;
+
+        // We should provide a function to retrieve modules,
+        // since the client tests do not support the ECMAScript 6 modules syntax
+        this.get = require;
```
inikulin: It's not because it doesn't support, but because hammerhead will be bundled into the single file and we will not have access to the internal modules by default

inikulin:
\r- Just a minor comment issue

miherlosev:
rename 'IsMozilla' -> 'IsFirefox' in corresponding comments

churkin:
lgtm

LavrovArtem:
FPR @inikulin @miherlosev

miherlosev:
lgtm

inikulin:
lgtm

PR #162, opened 2015-10-05
The URL attribute is set to an empty string on the client for the second time (T295078) (fix #159)

### LINKED ISSUE #159: The URL attribute is set to an empty string on the client for the second time (T295078)
That's why Firefox reloads the iframe and removes Hammerhead from it.

The Firefox scenario:

* an iframe without a src is being processed on the server
```html
<iframe src="" src-hammerhead-stored-value=""></iframe>
```
* ```window.Hammerhead``` is initialized in the iframe on the client
* ```src``` is again set to an empty string on the client
* the iframe is reloaded (i.e. ```window.Hammerhead``` is removed)
* finally, we try to execute the task script with eval() and get the ```TypeError: Hammerhead is undefined error```

LavrovArtem:
@inikulin @churkin

### DIFF
--- a/src/processing/dom/index.js
+++ b/src/processing/dom/index.js
@@ -397,7 +397,7 @@ export default class DomProcessor {
         if (urlReplacer && pattern.urlAttr) {
             var storedUrlAttr     = this.getStoredAttrName(pattern.urlAttr);
             var resourceUrl       = this.adapter.getAttr(el, pattern.urlAttr);
-            var processedOnServer = !!this.adapter.getAttr(el, storedUrlAttr);
+            var processedOnServer = this.adapter.hasAttr(el, storedUrlAttr);
 
             // NOTE: page resource URL with proxy URL
             if ((resourceUrl || resourceUrl === '') && !processedOnServer) {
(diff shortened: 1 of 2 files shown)


inikulin:
lgtm

churkin:
I do not see the test scenario

inikulin:
@churkin @LavrovArtem Seems like I'm missing context. Can you issue more details in #159: repro example, expected and actual behaviors

inikulin:
Odd, we are starting to get `module not found` errors in some PRs, \cc @AlexanderMoskovkin @churkin

inikulin:
Guys, please try to reproduce npm install fails with node 10 locally (you can use [nvm](https://github.com/coreybutler/nvm-windows) for windows to switch versions). If it reproduces locally, seems like some of our transitive dependencies requires node 12 now. If so we need to abandon node 10 support, which requires some additional job and benefits: native promises, classes, generators, conts, built-in keep-alive user agent, etc.

LavrovArtem:
FPR

inikulin:
Lgtm

churkin:
lgtm

PR #176, opened 2015-10-07
Focus-blur sandbox QUnit test doesn't work in Safari (close #171)

### LINKED ISSUE #171: Focus-blur sandbox QUnit test doesn't work in Safari
I've found a `hammerhead`-related qunit test in the `testcafe-phoenix` repo and it fails in Safari now.
We should move it to the `testcafe-hammerhead` and fix it.

The test title contains the 'on recording' string, but the problem was not related to `recorder` but to `hammerhead`.

```js
asyncTest('B237723 - Error on http://phonejs.devexpress.com/Demos/?url=KitchenSink&sm=3 on recording', function () {
    runAsyncTest(
        function () {
            var iframeSrc = window.QUnitGlobals.getResourceUrl('../../../data/focus-blur-sandbox/iframe.html');
            var $iframe   = $('<iframe></iframe>')
                .addClass(TEST_ELEMENT_CLASS)
                .attr('src', iframeSrc)
                .appendTo('body');

            var errorRaised = false;

            $iframe.load(function () {
                try {
                    $iframe[0].contentWindow.focusInput();
                }
                catch (err) {
                    errorRaised = true;
                }

                ok(!errorRaised, 'error is not raised');
                startNext();
            });
        }, 2000);
});
```

//data/focus-blur-sandbox/iframe.html
```html
<!DOCTYPE html>
<html>
<head>
    <title></title>
    <meta charset="utf-8">
    <script src="/hammerhead.js" class="script-hammerhead-shadow-ui"></script>
</head>
<body>
<script type="text/javascript">
    Hammerhead.get('./utils/url').OriginLocation.get = function () {
        return 'http://origin_iframe_host';
    };

    Hammerhead.start({
        crossDomainProxyPort: 2000
    });

    window.focusInput = function () {
        document.getElementById('input').focus();
    };
</script>

<input id="input"/>
</body>
</html>

```

The `ok(!errorRaised, 'error is not raised');` assertion is success but the test failed with the `TypeError: undefined is not an object (evaluating 'Object.prototype)`. It seems the test fails when `MessageSandbox` try to send a message to the top window from the removed iframe

churkin:
\сс @inikulin @LavrovArtem

### DIFF
--- a/src/client/sandbox/event/message.js
+++ b/src/client/sandbox/event/message.js
@@ -4,7 +4,7 @@ import nativeMethods from '../native-methods';
 import * as originLocation from '../../utils/origin-location';
 import { formatUrl, getCrossDomainProxyUrl, isSupportedProtocol } from '../../utils/url';
 import { parse as parseJSON, stringify as stringifyJSON } from '../../json';
-import { isIE9 } from '../../utils/browser';
+import { isIE9, isSafari } from '../../utils/browser';
 import { isCrossDomainWindows } from '../../utils/dom';
 import { isObjectEventListener } from '../../utils/event';
 import { Promise } from 'es6-promise';
@@ -172,12 +172,19 @@ export default class MessageSandbox extends SandboxBase {
         if (!this._isIframeRemoved() && (isIframeWithoutSrc || !isCrossDomainWindows(targetWindow, this.window) &&
                                                                targetWindow[this.RECEIVE_MSG_FN])) {
             //NOTE: postMessage delay imitation
-            nativeMethods.setTimeout.call(this.topWindow, () =>
+            nativeMethods.setTimeout.call(this.topWindow, () => {
+                // NOTE: We leave the capability to communicate with a removed frame. Unfortunately, this
+                // cannot be done in Safari, because it restricts operations with removed windows (GH-171).
+                var isUnreachableIframe = isSafari && !this.window.top;
+
+                if (!isUnreachableIframe) {
                     targetWindow[this.RECEIVE_MSG_FN]({
-                        data:   parseJSON(stringifyJSON(message)), // Cloning message to prevent this modification
+                        // NOTE: Cloning message to prevent this modification
+                        data:   parseJSON(stringifyJSON(message)),
                         source: this.window
-                    })
-                , 10);
+                    });
+                }
+            }, 10);
 
             return null;
         }
--- a/test/client/before-test.js
+++ b/test/client/before-test.js
@@ -23,9 +23,7 @@
         }
     };
 
-    Hammerhead.start({
-        sessionId: 'sessionId'
-    });
+    Hammerhead.start({ sessionId: 'sessionId' });
 
     window.overrideDomMeth = window[INTERNAL_PROPS.overrideDomMethodName];
 
@@ -39,7 +37,6 @@
     window.setProperty   = window[jsProcessor.SET_PROPERTY_METH_NAME];
     window.callMethod    = window[jsProcessor.CALL_METHOD_METH_NAME];
     window.getLocation   = window[jsProcessor.GET_LOCATION_METH_NAME];
-
 
     var globals = window.QUnitGlobals;
 
--- a/test/client/data/event-sandbox/focus-blur-sandbox.html
+++ b/test/client/data/event-sandbox/focus-blur-sandbox.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title></title>
+    <meta charset="utf-8">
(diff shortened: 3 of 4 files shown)


inikulin:
requires rebase now

churkin:
Ready to review

src/client/sandbox/event/message.js
```diff
@@ -169,12 +169,16 @@ export default class MessageSandbox extends SandboxBase {
         if (!this._isIFrameRemoved() && (isIFrameWithoutSrc || !isCrossDomainWindows(targetWindow, this.window) &&
                                                                targetWindow[this.RECEIVE_MSG_FN])) {
             //NOTE: postMessage delay imitation
-            nativeMethods.setTimeout.call(this.topWindow, () =>
+            nativeMethods.setTimeout.call(this.topWindow, () => {
+                // NOTE: GH-171
+                if (!isSafari || this.window.top) {
```
inikulin: This is cryptic. It took a while for me to decifer, what we shouldn't send message if we missing `window.top` in Safari. I suggest the following:
```js
// NOTE: add comment here about situation when it may happen with GH-issue reference
var isDetachedWindow = isSafari && !this.window.top;

if (!detachedWindow)
```
BTW, should we check for Safari here? We haven't encountered it in other browsers, but wouldn't it be better to use feature detection instead of browser check here (in case it will appear in other browser)?

inikulin:
\r-

src/client/sandbox/event/message.js
churkin: We want to leave the opportunity to communicate with a removed frame. This may be necessary in some cases. Unfortunately, we can not implement it in the Safari.

churkin:
FPR

src/client/sandbox/event/message.js
```diff
@@ -172,12 +172,19 @@ export default class MessageSandbox extends SandboxBase {
         if (!this._isIframeRemoved() && (isIframeWithoutSrc || !isCrossDomainWindows(targetWindow, this.window) &&
                                                                targetWindow[this.RECEIVE_MSG_FN])) {
             //NOTE: postMessage delay imitation
-            nativeMethods.setTimeout.call(this.topWindow, () =>
+            nativeMethods.setTimeout.call(this.topWindow, () => {
+                // NOTE: We leave the capability to communicate with a removed frame. Unfortunately, this
+                // cannot be done in Safari, because it restricts operations with removed windows (GH-171).
+                var isIframeRemoved = !this.window.top;
+
+                if (!isSafari || !isIframeRemoved) {
```
inikulin: Why we need this complicated condition once again?

src/client/sandbox/event/message.js
```diff
@@ -172,12 +172,19 @@ export default class MessageSandbox extends SandboxBase {
         if (!this._isIframeRemoved() && (isIframeWithoutSrc || !isCrossDomainWindows(targetWindow, this.window) &&
                                                                targetWindow[this.RECEIVE_MSG_FN])) {
             //NOTE: postMessage delay imitation
-            nativeMethods.setTimeout.call(this.topWindow, () =>
+            nativeMethods.setTimeout.call(this.topWindow, () => {
+                // NOTE: We leave the capability to communicate with a removed frame. Unfortunately, this
+                // cannot be done in Safari, because it restricts operations with removed windows (GH-171).
+                var isIframeRemoved = !this.window.top;
```
inikulin: It's should be `isRemovedIframe`. Anyway, `removed` is not the best word here, `detached` is better.

src/client/sandbox/event/message.js
inikulin: Why not 
```js
var isUnreachableIframe = isSafari && !this.window.top;
```

src/client/sandbox/event/message.js
inikulin: Moreover, `removed` iframe is already used here: https://github.com/churkin/testcafe-hammerhead/blob/i171/src/client/sandbox/event/message.js#L104

But it has quite different meaning. And still think we should get rid of Safari check. Seems like removed and reachable iframe will still have `window.top`

inikulin:
\r-

src/client/sandbox/event/message.js
churkin: I singled the '!this.window.top' expression because i think it can be unclear

inikulin: Actually it's much cleaner, because iframe is unreachable only in Safari (as you said), so it makes sense to include it in condition. Because instead I will wonder why we have additional condition for Safari and what does it mean. With this condition I will see what iframes can be unreachable in Safari browser.

churkin:
FPR

inikulin:
lgtm

LavrovArtem:
lgtm

PR #186, opened 2015-10-12
Use `es6-promisify` instead of our own promisify implementation (close #164)

### LINKED ISSUE #164: Use `es6-promisify` instead of our own promisify implementation
Note on `es6-promisify`: there is a difference between it and `Promise.denodeify` (and our own `promisify` implementation).

If node-style function has callback with conventional  two arguments `(error, data)`, everything is identical: Promises returned from both `es6-promisify` and `Promise.denodeify` resolve `data`, and reject `error`. But if callback is called with more arguments, like callback of [request](https://www.npmjs.com/package/request) function - `(error, response, body)`, then function promisified with `es6-promisify` will resolve array `[response, body]`, unlike `Promise.denodeify` and our own `promisify`, that resolves just `response`.

miherlosev:
/cc @inikulin @churkin


inikulin:
lgtm

churkin:
lgtm

PR #191, opened 2015-10-14
get rid of shared const (close #68)

### LINKED ISSUE #68: Get rid of shared const.
Split `const.js` into the semantically separated groups of constants and put them in the appropriate places. Remove shared const term from the code BTW. As an addition: rename `service-msg-cmd` to `commands` (like with `test-run` in testcafe package) and move it to the `/session` dir with the `session.js` and `cookie.js` (`session.js` being an `index.js` there).

miherlosev:
/cc @inikulin @churkin


src/processing/dom/index.js
```diff
@@ -152,6 +152,8 @@ export default class DomProcessor {
         var replacer = (resourceUrl, resourceType, charsetAttrValue) => urlReplacer(resourceUrl, resourceType, charsetAttrValue, baseUrl);
         var $all     = $('*');
 
+        var ELEMENT_IS_PROCESSED = 'hammerhead|element-is-processed';
```
inikulin: `const` here

src/client/dom-processor/client-dom-adapter.js
```diff
@@ -99,7 +99,7 @@ export default class ClientDomAdapter extends BaseDomAdapter {
     }
 
     isTopParentIframe (el) {
-        var elWindow = el[DOM_SANDBOX_PROCESSED_CONTEXT];
+        var elWindow = el[INTERNAL_ATTRS.processedContext];
```
inikulin: Ok, as I can see, here we have property, not attribute. So, seems, like we need to split constants in two groups: internal properties and internal attributes. We can put them in `processing/dom/internal/properties.js' and `processing/dom/internal/attributes.js' respectively.

src/client/sandbox/node/index.js
```diff
@@ -10,7 +10,7 @@ export default class NodeSandbox extends SandboxBase {
         super();
 
         this.raiseBodyCreatedEvent       = this._onBodyCreated;
-        document[CONST.DOCUMENT_CHARSET] = parseDocumentCharset();
+        document[INTERNAL_ATTRS.documentCharset] = parseDocumentCharset();
```
inikulin: Formatting is broken here

inikulin:
\r-

src/processing/dom/index.js
miherlosev: This place will be fixed in separate issue - https://github.com/superroma/testcafe-hammerhead/issues/194

src/client/dom-processor/client-dom-adapter.js
miherlosev: Internal properties:
    overrideDomMethodName:     'hammerhead|override-dom-method',
    processedContext:          'hammerhead|processed-context',
    whichPropertyWrapper:      'hammerhead|which-property-wrapper',
    documentCharset:           'hammerhead|document-charset',
Internal attributes:
    hoverPseudoClass:          'data-hammerhead-hovered',
    focusPseudoClass:          'data-hammerhead-focused',
    storedAttrPostfix:         '-hammerhead-stored-value',
    uploadInfoHiddenInputName: 'hammerhead|upload-info-hidden-input-name'

src/processing/dom/index.js
inikulin: I don't see why it's a reason to not write good **new** code

src/client/dom-processor/client-dom-adapter.js
inikulin: Looks good

miherlosev:
FPR

inikulin:
lgtm

inikulin:
@miherlosev What was changed after my review?

miherlosev:
as say @churkin 
we need to change order for `import` as follow:
at first - constants
at second - classes
then utils and separate methods (`import` as {...} from)

inikulin:
@miherlosev Guys don't do this, let's not complicate rules. Currently the rule is simple, let's keep it:
at first external or built-in deps, then internal deps.

churkin:
lgtm

PR #196, opened 2015-10-14
Add a ShadowUI suffix to css rules for the injectable stylesheets (close #129)

### LINKED ISSUE #129: Implement API to add a ShadowUI suffix to css rules for the injectable stylesheets
Hammehread should have API to process css files (add a suffix to css rules).

It's necessary to fix the https://github.com/superroma/testcafe-phoenix/issues/54 issue

LavrovArtem:
/cc @inikulin @churkin

### DIFF
--- a/src/proxy/router.js
+++ b/src/proxy/router.js
@@ -11,7 +11,6 @@ function buildRouteParamsMap (routeMatch, paramNames) {
         params[paramName] = routeMatch[i + 1];
         return params;
     }, {});
-
 }
 
 // Router
@@ -31,8 +30,11 @@ export default class Router {
         else {
             var isStatic = typeof handler !== 'function';
 
-            if (isStatic)
+            if (isStatic) {
+                this._processStaticContent(handler);
+
                 handler.etag = md5(handler.content);
+            }
 
             this.routes[`${method} ${route}`] = {
                 handler:  handler,
@@ -91,6 +93,10 @@ export default class Router {
         }
 
         return false;
+    }
+
+    _processStaticContent () {
+        throw new Error('Not implemented');
     }
 
     // API
--- a/src/shadow-ui/create-shadow-stylesheet.js
+++ b/src/shadow-ui/create-shadow-stylesheet.js
@@ -1,0 +1,34 @@
+import css from 'css';
+import SHADOW_UI_CLASS_NAME from './class-name';
+
+const ID_OR_CLASS_RE          = /#[a-zA-Z0-9_-]+|\.-?[a-zA-Z0-9_][a-zA-Z0-9_-]*/g;
+const ADD_POSTFIX_REPLACEMENT = '$&' + SHADOW_UI_CLASS_NAME.postfix;
+
+function transformSelector (selector) {
+    return selector.replace(ID_OR_CLASS_RE, ADD_POSTFIX_REPLACEMENT);
+}
+
+function addUIClassPostfix (rules) {
+    var ruleStack = rules.slice();
+
+    while (ruleStack.length) {
+        var rule = ruleStack.pop();
+
+        if (rule.type === 'rule' && rule.selectors)
+            rule.selectors = rule.selectors.map(transformSelector);
+
+        if (rule.rules)
+            ruleStack = ruleStack.concat(rule.rules);
+    }
+}
+
+export default function createShadowStylesheet (cssCode) {
+    var ast = css.parse(cssCode, { silent: true });
+
+    addUIClassPostfix(ast.stylesheet.rules);
+
+    return css.stringify(ast, {
+        indent:   0,
+        compress: false
+    });
+}
--- a/test/server/proxy-test.js
+++ b/test/server/proxy-test.js
@@ -3,6 +3,7 @@ var request            = require('request');
 var expect             = require('chai').expect;
 var express            = require('express');
 var Promise            = require('es6-promise').Promise;
+var read               = require('read-file-relative').readSync;
 var Proxy              = require('../../lib/proxy');
 var Session            = require('../../lib/session');
 var COMMAND            = require('../../lib/session/command');
@@ -500,6 +501,25 @@ describe('Proxy', function () {
             request(proxy.openSession('http://127.0.0.1:2000/with-auth', session), function (err, res, body) {
                 expect(res.statusCode).eql(401);
                 expect(body).to.be.empty;
+                done();
+            });
+        });
+    });
+
+    describe('Shadow UI', function () {
+        it('Should process shadow ui stylesheet', function (done) {
+            var src      = read('/data/shadow-ui-stylesheet/src.css').toString();
+            var expected = read('/data/shadow-ui-stylesheet/expected.css').toString();
+
+            proxy.GET('/testcafe-ui-styles.css', {
+                contentType:          'text/css',
+                content:              src,
+                isShadowUIStylesheet: true
+            });
+
+            request('http://127.0.0.1:1836/testcafe-ui-styles.css', function (err, res, body) {
+                expect(body.replace(/\r\n|\n/g, '')).equal(expected.replace(/\r\n|\n/g, ''));
+
                 done();
             });
         });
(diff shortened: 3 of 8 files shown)


src/proxy/router.js
```diff
@@ -31,8 +31,12 @@ export default class Router {
         else {
             var isStatic = typeof handler !== 'function';
 
-            if (isStatic)
+            if (isStatic) {
+                if (handler.isShadowUIStylesheet)
```
inikulin: I have feeling what it's not on it's place in route and it should be done in proxy. Maybe we can make method `_processStaticContent` which will do nothing in router, but will check flag and convert css in proxy

src/shadow-ui/prepare-stylesheet.js
```diff
@@ -0,0 +1,27 @@
+import css from 'css';
+import SHADOW_UI_CLASS_NAME from './class-name';
+
+const ID_AND_CLASS_REG_EXP = /#[a-zA-Z0-9_-]+|\.-?[a-zA-Z0-9_][a-zA-Z0-9_-]*/g;
```
inikulin: We using `_RE` postfix for regular expressions now

src/shadow-ui/prepare-stylesheet.js
```diff
@@ -0,0 +1,27 @@
...
+const ADD_POSTFIX_PATTERN  = '$&' + SHADOW_UI_CLASS_NAME.postfix;
```
inikulin: Seems like it should be `ID_AND_CLASS_REPLACEMENT`

src/shadow-ui/prepare-stylesheet.js
```diff
@@ -0,0 +1,27 @@
...
+
+
+function addUIClassPostfix (rules) {
+    rules.forEach(rule => {
+        if (rule.type === 'rule' && rule.selectors && rule.selectors.length)
+            rule.selectors = rule.selectors.map(selector => selector.replace(ID_AND_CLASS_REG_EXP, ADD_POSTFIX_PATTERN));
+
+        if (rule.rules)
+            addUIClassPostfix(rule.rules);
```
inikulin: We need to avoid recursion here (in general, it's always a good rule). You can use stack technique here:
```js
// NOTE: make a shallow copy of array here, to avoid modification
// of the original `ast.stylesheet.rules`
var ruleStack = rules.slice();

while (ruleStack.length) {
   var rule = ruleStack.pop();
    // Do processing here
    ...

    if (rule.rules)
        ruleStack = ruleStack.concat(rule.rules);
}
```

src/shadow-ui/prepare-stylesheet.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+}
+
+export default function prepareStylesheet (text) {
```
inikulin: I don't think this func name fits well. Maybe `convertToShadowStylesheet` (we need to change filename as well)? \cc @VasilyStrelyaev

inikulin: Argument should be name `css`

test/server/router-test.js
```diff
@@ -155,5 +156,43 @@ describe('Router', function () {
         expect(resMock.content).to.be.empty;
     });
 
+    it('Should prepare shadow ui stylesheet', function () {
+        var router   = new Router();
+        var src      = fs.readFileSync('test/server/data/shadow-ui-stylesheet/src.css').toString();
```
inikulin: Use `read-file-relative` util

inikulin:
\r-

src/shadow-ui/prepare-stylesheet.js
VasilyStrelyaev: hmm... both are ok, but I prefer `create...`

LavrovArtem:
FPR

src/proxy/router.js
```diff
@@ -93,6 +95,10 @@ export default class Router {
         return false;
     }
 
+    _processStaticContent () {
+
```
inikulin: Add comment here: `Do nothing =)`. So it will be clear what this is a virtual method and not piece of abandoned code. Smile at the end is the absolute semantic requirement in this sentence, BTW :hamburger:

src/shadow-ui/create-shadow-stylesheet.js
```diff
@@ -0,0 +1,32 @@
+import css from 'css';
+import SHADOW_UI_CLASS_NAME from './class-name';
+
+const ID_OR_CLASS_RE          = /#[a-zA-Z0-9_-]+|\.-?[a-zA-Z0-9_][a-zA-Z0-9_-]*/g;
+const ADD_POSTFIX_REPLACEMENT = '$&' + SHADOW_UI_CLASS_NAME.postfix;
+
+
+function addUIClassPostfix (rules) {
+    var ruleStack       = [].concat(rules);
```
inikulin: Why not  `rules.slice()`?

src/shadow-ui/create-shadow-stylesheet.js
```diff
@@ -0,0 +1,32 @@
...
+    var processSelector = selector => selector.replace(ID_OR_CLASS_RE, ADD_POSTFIX_REPLACEMENT);
```
inikulin: Avoid creating functional expressions, just do it inline

src/shadow-ui/create-shadow-stylesheet.js
```diff
@@ -0,0 +1,32 @@
...
+
+    while (ruleStack.length) {
+        var rule = ruleStack.pop();
+
+        if (rule.type === 'rule' && rule.selectors && rule.selectors.length)
+            rule.selectors = rule.selectors.map(processSelector);
+
+        if (rule.rules && rule.rules.length)
```
inikulin: No need for the length check here

src/proxy/router.js
LavrovArtem: Maybe ```throw new Error('Not implemented');``` like in ```base-dom-adapter.js```

test/server/proxy-test.js
```diff
@@ -505,6 +506,25 @@ describe('Proxy', function () {
         });
     });
 
+    describe('Shadow ui content', function () {
```
inikulin: Call it `Shadow UI` without `content`

inikulin:
Just minor issues left. \r-

src/shadow-ui/create-shadow-stylesheet.js
LavrovArtem: eslint throw error ```Don't make functions within a loop```

inikulin: Well, then make it a function outside the `addUIClassPostfix` scope.

inikulin: And call it `transformSelector`

LavrovArtem:
FPR

src/shadow-ui/create-shadow-stylesheet.js
```diff
@@ -0,0 +1,34 @@
...
+    return selector.replace(ID_OR_CLASS_RE, ADD_POSTFIX_REPLACEMENT);
+}
+
+function addUIClassPostfix (rules) {
+    var ruleStack = rules.slice();
+
+    while (ruleStack.length) {
+        var rule = ruleStack.pop();
+
+        if (rule.type === 'rule' && rule.selectors && rule.selectors.length)
```
inikulin: Arghh, missed that. Do we really need `length` check here?

inikulin:
LGTM

src/proxy/router.js
```diff
@@ -93,6 +95,10 @@ export default class Router {
         return false;
     }
 
+    _processStaticContent () {
+        // Do nothing =)
+    }
```
churkin: In the dom processor we are using 'throw new Error('Not implemented');' for the abstract methods.

src/proxy/router.js
inikulin: Makes sense, but we need to patch it in the tests then.

churkin:
lgtm

PR #200, opened 2015-10-16
FIle uploading doesn't work (close #195)

### LINKED ISSUE #195: FIle uploading doesn't work
Reproduced in playground.
You can check it on the [ASPxUploadControl demo page](https://demos.devexpress.com/ASPxFileManagerAndUploadDemos/UploadControl/MultiFileSelection.aspx). Just try to upload any file with hammerhead playground.

churkin:
Content length should be recalculated after upload info injecting.

\сс @inikulin @AlexanderMoskovkin

### DIFF
--- a/src/request-pipeline/index.js
+++ b/src/request-pipeline/index.js
@@ -90,6 +90,13 @@ var stages = {
 function createReqOpts (ctx) {
     var bodyWithUploads = injectUpload(ctx.req.headers['content-type'], ctx.reqBody);
 
+    // NOTE: First, we should rewrite the request body, because the 'content-length' header will be built based on it.
+    if (bodyWithUploads)
+        ctx.reqBody = bodyWithUploads;
+
+    // NOTE: All headers, including 'content-length', are built here.
+    var headers = headerTransforms.forRequest(ctx, this);
+
     return {
         url:         ctx.dest.url,
         protocol:    ctx.dest.protocol,
@@ -98,9 +105,9 @@ function createReqOpts (ctx) {
         port:        ctx.dest.port,
         path:        ctx.dest.partAfterHost,
         method:      ctx.req.method,
-        headers:     headerTransforms.forRequest(ctx, this),
         credentials: ctx.session.getAuthCredentials(),
-        body:        bodyWithUploads || ctx.reqBody
+        body:        ctx.reqBody,
+        headers:     headers
     };
 }
 
--- a/test/server/proxy-test.js
+++ b/test/server/proxy-test.js
@@ -1,15 +1,15 @@
+var Promise            = require('es6-promise').Promise;
 var fs                 = require('fs');
 var request            = require('request');
 var expect             = require('chai').expect;
 var express            = require('express');
-var Promise            = require('es6-promise').Promise;
 var read               = require('read-file-relative').readSync;
+var COMMAND            = require('../../lib/session/command');
+var XHR_HEADERS        = require('../../lib/request-pipeline/xhr/headers');
 var Proxy              = require('../../lib/proxy');
 var Session            = require('../../lib/session');
-var COMMAND            = require('../../lib/session/command');
 var DestinationRequest = require('../../lib/request-pipeline/destination-request');
 var requestAgent       = require('../../lib/request-pipeline/destination-request/agent');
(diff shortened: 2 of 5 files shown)


src/request-pipeline/index.js
```diff
@@ -90,6 +90,11 @@ var stages = {
 function createReqOpts (ctx) {
     var bodyWithUploads = injectUpload(ctx.req.headers['content-type'], ctx.reqBody);
 
+    if (bodyWithUploads)
```
inikulin: We need comment here about length recalculation. Because, initially it was done this way, but then removed, because it's not clear from code why we can't do it inline

inikulin:
\r-

churkin:
FPR

inikulin:
@churkin Just figured out what we need a regression test for this.
\r-

churkin:
FPR

test/server/proxy-test.js
```diff
@@ -604,5 +608,25 @@ describe('Proxy', function () {
...
+                method:  'POST',
+                body:    src,
+                headers: {
+                    'content-type': 'multipart/form-data; boundary=separator',
+                    'accept':       'text/html,application/xhtml+xml,application/xml;q=0.9,*!/!*;q=0.8'
+                }
+            };
+
+            request(options, function (err, res, body) {
+                expect(body.indexOf('%content-length:' + expected.length + '%')).gt(-1);
```
inikulin: Why not strict equality here?

inikulin: Or even better we can send back modified body, and compare it to the `expected`.

test/server/utils/new-line-replacer.js
```diff
@@ -0,0 +1,3 @@
+exports.replace = function (content) {
```
inikulin: Why not call it `replaceNewLines` and just export the whole function?:
```js
module.exports = function(content) {
    return new Buffer(content.toString().replace(/\r\n|\n/gm, '\r\n'));
};
```

inikulin: Looking at this now, it shouldn't be even named `replaceNewLines` it's too generic for this function, because:
 - it works with buffers only
 - it replaces with '\r\n`
Both are specific for formdata, so it should be called `normalizeFormDataNewLines`

inikulin: But the best thing will be to not extract utilities at all, just keep this short functions in both files.

inikulin:
\r-

churkin:
FPR

test/server/utils/new-line-replacer.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function replaceNewLines (content) {
```
inikulin: As I stated previously let's not make utils for this single method, Just keep it inside tests. It's naming is still misleading, because it's very specific to the tests and form data.

inikulin:
\r-

src/request-pipeline/index.js
```diff
@@ -90,6 +90,13 @@ var stages = {
 function createReqOpts (ctx) {
     var bodyWithUploads = injectUpload(ctx.req.headers['content-type'], ctx.reqBody);
 
+    // NOTE: First, we should rewrite the request body, because the 'content-length' header will be built based on it.
+    if (bodyWithUploads)
```
inikulin: \cc @VasilyStrelyaev check comment text, pls

churkin: checked

test/server/proxy-test.js
```diff
@@ -604,5 +611,25 @@ describe('Proxy', function () {
...
+                method:  'POST',
+                body:    src,
+                headers: {
+                    'content-type': 'multipart/form-data; boundary=separator',
+                    'accept':       'text/plain;q=0.9,*!/!*;q=0.8'
+                }
+            };
+
+            request(options, function (err, res, body) {
+                expect(body).eql(expected.length.toString());
```
inikulin: Once again: why not send back the body and compare it with `expected`? Why we compare just length and not content?

inikulin:
@churkin Tests failed due to ES6 syntax in tests

churkin:
FPR

inikulin:
lgtm

AlexanderMoskovkin:
lgtm

PR #204, opened 2015-10-20
Create API which will provide infrastructure for the cross-lib interaction (close #18)

### LINKED ISSUE #18: Create API which will provide infrastructure for the cross-lib interaction
Think how hammerhead can be safely exposed to TestCafe scripts: provide non-enumerable window property which will contain all script getters (without setters). You can call it `app.js`. Each module will take it's reference of the app from the `window`.
E.g. in `test-cafe/ui.js`:

```js
var hammerhead = window['@TestCafe@'].getLib('hammerhead');
window['@TestCafe@'].registerLib('test-cafe-ui', exports);
```
\cc @AlexanderMoskovkin 

ref #5

churkin:
\cc @inikulin @AlexanderMoskovkin 

Guys, what do you think about this decision?

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -132,6 +132,11 @@ class Hammerhead {
 
 var hammerhead = new Hammerhead();
 
-window.Hammerhead = hammerhead;
+Object.defineProperty(window, '%hammerhead%', {
+    enumerable:   false,
+    configurable: false,
+    writable:     false,
+    value:        hammerhead
+});
 
 export default hammerhead;
--- a/src/client/sandbox/cookie.js
+++ b/src/client/sandbox/cookie.js
@@ -10,7 +10,7 @@ import { queuedAsyncServiceMsg } from '../transport';
 export default class CookieSandbox extends SandboxBase {
     _getSettings () {
         var windowSettings = this.window !== this.window.top && !isCrossDomainWindows(this.window, this.window.top) ?
-                             this.window.top.Hammerhead.get('./settings') : settings;
+                             this.window.top['%hammerhead%'].get('./settings') : settings;
 
         return windowSettings.get();
     }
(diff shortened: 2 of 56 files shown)


src/client/index.js
```diff
@@ -132,6 +132,11 @@ class Hammerhead {
 
 var hammerhead = new Hammerhead();
 
-window.Hammerhead = hammerhead;
+Object.defineProperty(window, '@hammerhead@', {
```
inikulin: We use `%name%` in testcafe

inikulin:
\r-

src/client/index.js
churkin: We are using `@hammerhead@` in the src attribute values. '%' is a URI reserved char:
https://tools.ietf.org/html/rfc3986#section-2.1
This is potentially dangerous, for example, our qunit tests fall while trying to decode url. I suggest using '@' char.

churkin: @inikulin what do you say?

inikulin: I don't think this exact var will ever appear in the URL. So, it's completely OK. Moreover, we will need to modify testcafe vars due to the incosistency. `%`-notation for modules is also better because it will make it distinct from our internal constants, which is good.

churkin: We already use @hammerhead@ in the attribute values:

```html
<iframe src="javascript:'<html><head>...</head><body>
<script type="text/javascript"; class="script-hammerhead-shadow-ui">
```

```javascript
if (window['@hammerhead@'])`
   `window['@hammerhead@'].sandbox.node.raiseBodyCreatedEvent();`
var script = document.currentScript || document.scripts[document.scripts.length - 1];
script.parentNode.removeChild(script);
```
```html
</script></body></html>'" src-hammerhead-stored-value="javascript:'<html><body></body></html>'">
</iframe>
```

churkin:
FPR

churkin:
FPR

inikulin:
lgtm

AlexanderMoskovkin:
lgtm

PR #211, opened 2015-10-27
Web platform tests integration. Part 1 - Base.

churkin:
\cc @inikulin @LavrovArtem

### DIFF
--- a/test/web-platform/views/index.ejs
+++ b/test/web-platform/views/index.ejs
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/2.1.4/semantic.css" rel="stylesheet"
+          type="text/css"/>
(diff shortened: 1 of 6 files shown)


test/web-platform/views/index.ejs
```diff
@@ -0,0 +1,45 @@
...
+<body>
+<div class="ui middle aligned center aligned grid">
+    <div class="column">
+        <div class="ui segments">
+            <div class="ui basic center aligned segment">
+                <h3 class="ui header">Web platform tests</h3>
+            </div>
+            <div class="ui basic left aligned segment">
+                <ul class="ui list">
+                    <% for (var i = 0; i < tests.length; i++) { %>
```
inikulin: Can we use Mustache instead of EJS? We've switched to Mustache everywhere. For now we only have EJS templates in playground (just a couple of lines) and here. Can we switch here and in playground and remove extra dependency?

inikulin:
Overall, lgtm, but let's switch templating engine.
\r-

Issue #179, opened 2015-10-08
Need to refactror FocusBlurSandbox


miherlosev:
Problem places:
1)Nested calls for __raiseEvent
2)Large methods
3)'Callback' parameter of focus, blur methods uses only for tests

inikulin:
The most important: `_raiseEvent` is inconsistent: for IE it should be sync, while for other browsers it should be async. We should split it into two version: the sync one and the async one.

miherlosev:
I'm working on this

miherlosev:
Done

PR #219, opened 2015-10-29
Change Promise lib, expose Promise in client API(fixes #189)

### LINKED ISSUE #189: Let's rethink our Promise lib (once again =))
What I don't like currently:
- es6-promise lib, because we need to do [this](https://github.com/superroma/testcafe-hammerhead/blob/master/Gulpfile.js#L135) and it's quite big. I found nice, small, browser-compatible alternative which doesn't pollute global scope: https://github.com/floatdrop/pinkie Ideal candidate for client-side usage.
- testcafe also uses es6-promises which is required by each it's client module. So we have a lot if copies of the same module on client side. Hammerhead should expose it's Promise in API, so it can be used by testcafe.
- on server side we could use same library OR just abandon node 10 support and use native promises (I starting to think more and more in regard of this option)

inikulin:
\cc @churkin @miherlosev


miherlosev:
lgtm

src/client/sandbox/event/focus-blur.js
```diff
@@ -115,8 +115,8 @@ export default class FocusBlurSandbox extends SandboxBase {
 
 
     _raiseEvent (el, type, callback, withoutHandlers, isAsync, forMouseEvent, preventScrolling) {
-        // NOTE: We cannot use Promise from the es6-promise library because the 'resolve' method is called
-        // from setTimeout(1) in IE9 and IE10.
+        // NOTE: We cannot use Promise because 'resolve' will be called async, but we need to resolve
+        // immediately in IE9 and IE10.
```
churkin: As far as I understand, in this case, the 'pinkie' will execute the code in sync.

churkin: https://github.com/floatdrop/pinkie/blob/master/index.js#L33

inikulin: We've decided to not refactor this code right now, because we shouldn't rely on how promise lib executes the code. We should refactor the whole code so we will explicitly use Promise based function there we expect async behavior, and don't use it there we expect sync (see #179)

churkin:
lgtm

PR #224, opened 2015-10-30
A service message will not be sent from iframe if it is removed afterwards (fix #64)

### LINKED ISSUE #64: A service message will not be sent from iframe if it is removed afterwards
A service message will not be sent from iframe if it is removed afterwards
Our on-farm functional tests in the default branch fail in Chrome due to this error.

1) in the latest version of Chrome, the timeout is reset (even the window.top timeout that was set after the iframe was unloaded)
sendServiceMsg method:
```js
 NativeMethods.setTimeout.call(topWindow, function () {
            targetWindow[RECEIVE_MSG_FN]({
                data:   JSON.parse(JSON.stringify(message)), 
                source: window
            });
        }, 10);
```
2)  in IE, a parse error is raised after the iframe is removed
```js
JSON.parse(JSON.stringify(message))
```
Regression qunit test:
```js
asyncTest('service message from removed iframe', function () {
    var $iFrame         = $('<iframe>');
    var messageReceived = false;

    $iFrame[0].src = window.QUnitGlobals.getResourceUrl('../../data/same-domain/service-message-from-removed-iframe.html');
    $iFrame.appendTo('body');

    window.setTimeout(function(){
        ok(messageReceived, 'message is not received');
        start();
    }, 100);

    MessageSandbox.on(MessageSandbox.SERVICE_MSG_RECEIVED, function () {
        messageReceived = true;
    });
});
```
tested iframe markup:
```js
<!DOCTYPE html>
<html>
<head>
    <title></title>
    <meta charset="utf-8">
    <script src="/hammerhead.js" class="script-TC2b9a6d"></script>
</head>
<body>
<script type="text/javascript">
    var hammerhead = window.Hammerhead;
    var urlUtil = hammerhead.get('./utils/url');

    hammerhead.get('./settings').set({
        CROSS_DOMAIN_PROXY_PORT: 2000
    });

    urlUtil.OriginLocation.get = function () {
        return 'https://example.com';
    };

    var MessageSandbox = Hammerhead.get('./sandboxes/message');
    var scriptProcessor = Hammerhead.get('../processing/script');

    Hammerhead.init();

    var iframe = window.top.document.getElementsByTagName('iframe')[0];

    debugger;
    MessageSandbox.sendServiceMsg('', window.top);

    iframe.parentNode.removeChild(iframe);
</script>
<div></div>
</body>
</html>
```

LavrovArtem:
/cc @churkin @miherlosev

### DIFF
--- a/src/client/sandbox/event/message.js
+++ b/src/client/sandbox/event/message.js
@@ -4,7 +4,7 @@ import nativeMethods from '../native-methods';
 import * as originLocation from '../../utils/origin-location';
 import { formatUrl, getCrossDomainProxyUrl, isSupportedProtocol } from '../../utils/url';
 import { parse as parseJSON, stringify as stringifyJSON } from '../../json';
-import { isIE9, isSafari } from '../../utils/browser';
+import { isIE9 } from '../../utils/browser';
 import { isCrossDomainWindows } from '../../utils/dom';
 import { isObjectEventListener } from '../../utils/event';
 import { Promise } from 'es6-promise';
@@ -17,7 +17,7 @@ const MESSAGE_TYPE = {
 };
 
 export default class MessageSandbox extends SandboxBase {
-    constructor (listeners) {
+    constructor (listeners, unloadSandbox) {
         super();
 
         this.PING_DELAY                 = 200;
@@ -33,9 +33,12 @@ export default class MessageSandbox extends SandboxBase {
         this.topWindow = null;
         this.window    = null;
 
-        this.listeners = listeners;
+        this.listeners     = listeners;
+        this.unloadSandbox = unloadSandbox;
 
         this.storedOnMessageHandler = null;
+
+        this.iframeInternalMsgQueue = [];
     }
 
     _onMessage (e) {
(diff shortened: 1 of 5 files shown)


src/client/sandbox/event/message.js
```diff
@@ -33,9 +33,12 @@ export default class MessageSandbox extends SandboxBase {
         this.topWindow = null;
         this.window    = null;
 
-        this.listeners = listeners;
+        this.listeners     = listeners;
+        this.unloadSandbox = unloadSandbox;
 
         this.storedOnMessageHandler = null;
+
+        this.serviceMsgsQueue = [];
```
inikulin: Is this client-server service messages or cross-domain communication service messages?
If first, we should move this code to `transport`, if second we should rename them somehow to distinct them from client-server XHR methods.

inikulin:
\r-

src/client/sandbox/event/message.js
LavrovArtem: Is this service messages between iframe without src and same-domain iframe

inikulin: Then we need to rename it. Discuss naming with @VasilyStrelyaev

LavrovArtem: maybe ```serviceMsgsBetweenSameDomainIframe```

VasilyStrelyaev: Won't `interIframeMsgQueue` be enough?

LavrovArtem: this name does not show that applies only to same-domain without cross-domain

VasilyStrelyaev: should it? is there something that can be confused with it?

VasilyStrelyaev: We shouldn't use `interIframeSameDomainMsgQueue` unless it's super-necessary.

LavrovArtem:
FPR

src/client/sandbox/event/message.js
```diff
@@ -33,9 +33,12 @@ export default class MessageSandbox extends SandboxBase {
         this.topWindow = null;
         this.window    = null;
 
-        this.listeners = listeners;
+        this.listeners     = listeners;
+        this.unloadSandbox = unloadSandbox;
 
         this.storedOnMessageHandler = null;
+
+        this.interIframeMsgQueue = [];
```
inikulin: It's better be named `iframeInternalMsgQueue

src/client/sandbox/event/message.js
```diff
@@ -105,11 +108,32 @@ export default class MessageSandbox extends SandboxBase {
         return this.window.top === this.window.self && this.window !== this.topWindow;
     }
 
+    _removeServiceMsgFromQueue (sendFunc) {
```
inikulin: `_removeInternalMsgFromQueue`

inikulin:
\r-

LavrovArtem:
FPR

inikulin:
lgtm

miherlosev:
lgtm

churkin:
lgtm

PR #225, opened 2015-10-30
Few simple fixes (close #194, close #192)

### LINKED ISSUE #192: Rename 'uploadStoragePath' to 'uploadsRoot'
in [Session constructor](https://github.com/superroma/testcafe-hammerhead/blob/master/src/session/index.js#L15) and [UploadStorage constructor](https://github.com/superroma/testcafe-hammerhead/blob/master/src/upload/storage.js#L12)

/cc @inikulin @churkin

### LINKED ISSUE #194: Multiple element processing inside DomProcessor.processPage
This is perfomance issue.
`CONST.ELEMENT_PROCESSED_FLAG` always equal `false` because `Const` does not contain `ELEMENT_PROCESSED_FLAG`
Code fragment:
```javascript
var $all     = $('*');

        for (var i = 0; i < this.elementProcessorPatterns.length; i++) {
            var pattern = this.elementProcessorPatterns[i];

            /*eslint-disable no-loop-func*/
            $all.filter(function () {
                return pattern.selector(this);
            }).each(function () {
                if (!this[CONST.ELEMENT_PROCESSED_FLAG]) {
                    for (var j = 0; j < pattern.elementProcessors.length; j++)
                        pattern.elementProcessors[j].call(domProc, this, replacer, pattern);
                }
            });
            /*eslint-enable no-loop-func*/
        }
```
Need to provide perfomance check results

churkin:
\cc @inikulin @LavrovArtem

### DIFF
--- a/src/processing/dom/index.js
+++ b/src/processing/dom/index.js
@@ -34,6 +34,8 @@ const TARGET_ATTR_TAGS = {
     base: true
 };
 
+const ELEMENT_PROCESSED = 'hammerhead|element-processed';
+
 export default class DomProcessor {
     constructor (adapter) {
         this.adapter = adapter;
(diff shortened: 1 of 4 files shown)


LavrovArtem:
lgtm

inikulin:
lgtm

PR #236, opened 2015-11-02
Remove  'integrity' attributes from resources (close #235)

### LINKED ISSUE #235: Remove `integrity` attributes from resources
See: http://www.w3.org/TR/SRI/

This attribute is used to avoid MITM attacks. Attribute contains SHA of the desired resource. Which doesn't work for us. This is widely used currently. And it brokes e.g. GitHub proxying. I suggest to remove this attribute entirely. It applies only to `<script>` and `<link>` elements.

churkin:
\cc @inikulin @LavrovArtem


churkin:
FPR

src/processing/dom/index.js
```diff
@@ -209,6 +216,13 @@ export default class DomProcessor {
         this.adapter.setAttr(el, 'autocomplete', 'off');
     }
 
+    // NOTE: We just remove the 'integrity' attribute, because its value is not relevant after the script content
```
inikulin: Please, check comment with @VasilyStrelyaev

inikulin:
lgtm, just need to check comment

src/processing/dom/index.js
```diff
@@ -209,6 +216,13 @@ export default class DomProcessor {
...
+    // is changed (http://www.w3.org/TR/SRI/). If this causes problems in the future, we will need to generate
+    // the correct SHA for the changed script.
```
VasilyStrelyaev: already checked, but there's nothing that can't be improved
```
    // NOTE: We simply remove the 'integrity' attribute because its value will not be relevant after the script content
    // changes (http://www.w3.org/TR/SRI/). If this causes problems in the future, we will need to generate
    // the correct SHA for the changed script.
```

VasilyStrelyaev:
lgtm

churkin:
FPR

inikulin:
lgtm

LavrovArtem:
lgtm

PR #241, opened 2015-11-05
Scripts passed to the WebWorker via Blob or data-uri doesn't have processing header (close #231)

### LINKED ISSUE #231: Scripts passed to the WebWorker via Blob or data-uri doesn't have processing header
We should always have processing header in WebWorkers since we need to provide instruction fallbacks. But if script was provided via Blob object or data-uri we don't explicitly add processing header which results in error. 

**Refs:**
About WebWorker Blob :  http://stackoverflow.com/questions/10343913/how-to-create-a-web-worker-from-a-string

Failing page: http://www.homedepot.com/c/HomePageRegional3?s_tnt=56312:1:0 
(in HHM: https://testcafe-hhhm.devexpress.com/history?timestamp=1446404591534&url=http://www.homedepot.com/c/HomePageRegional3?s_tnt=56312:1:0)

churkin:
\cc @miherlosev @inikulin

### DIFF
--- a/src/client/sandbox/node/window.js
+++ b/src/client/sandbox/node/window.js
@@ -117,23 +117,23 @@ export default class WindowSandbox extends SandboxBase {
 
         if (window.Blob) {
             window.Blob = function (parts, opts) {
+                if (arguments.length === 0)
+                    return new nativeMethods.Blob();
+
+                var type = opts && opts.type && opts.type.toString().toLowerCase();
+
+                // NOTE: If we cannot identify the content type of data, we're trying to process it as a script.
+                // Unfortunately, we do not have the ability to exactly identify a script. That's why we make such
+                // an assumption. We cannot solve this problem at the Worker level either, because the operation of
+                // creating a new Blob instance is asynchronous. (GH-231)
+                if (!type || type === 'text/javascript' || type === 'application/javascript' ||
+                    type === 'application/x-javascript')
+                    parts = [processScript(parts.join(''), true, false)];
+
                 // NOTE: IE11 throws an error when the second parameter of the Blob function is undefined (GH-44)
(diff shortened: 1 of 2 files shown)


src/client/sandbox/node/window.js
```diff
@@ -117,23 +117,23 @@ export default class WindowSandbox extends SandboxBase {
...
+                if (arguments.length === 0)
+                    return new nativeMethods.Blob();
+
+                var type = opts && opts.type && opts.type.toString().toLowerCase();
+
+                // NOTE: If we cannot identify the content type of data, we're trying to process it as a script.
+                // Unfortunately, we do not have the ability to exactly identify a script. That's why we make such
+                // an assumption. We cannot solve this problem at the Worker level either, because the operation of
+                // creating a new Blob instance is asynchronous. (GH-231)
+                if (!opts || type === 'text/javascript' || type === 'application/javascript' ||
```
inikulin: I assume we should check for the `!type` here instead of `!opts`

test/client/fixtures/sandbox/node/window-test.js
```diff
@@ -13,6 +13,15 @@ test('window.onerror setter/getter', function () {
 
 module('regression');
 
+asyncTest('Blob should try to process data as a script even if the content type is not passed (GH-231)', function () {
```
inikulin: We need make test conditional, since IE9 doesn't support workers and Blob

inikulin:
\r-

churkin:
FPR

miherlosev:
lgtm

churkin:
FPR

inikulin:
lgtm

miherlosev:
lgtm

PR #246, opened 2015-11-09
remove <meta httpequiv="Content-Security-Policy"> from markup (close #243)

### LINKED ISSUE #243: Add support sites with 'Content-Security-Policy'
'Content-Security-Policy' header allows to restrict resouce loading (scripts, styles).
After proxing a same sites we can have a 'Refused to load the script ' errors.

Example - https://www.dropbox.com/

Details:
http://www.w3.org/TR/CSP1/
https://en.wikipedia.org/wiki/Content_Security_Policy
http://content-security-policy.com/
http://www.html5rocks.com/en/tutorials/security/content-security-policy/

miherlosev:
/cc @inikulin @churkin 
@VasilyStrelyaev check test name and comments

### DIFF
--- a/src/processing/dom/index.js
+++ b/src/processing/dom/index.js
@@ -279,12 +279,19 @@ export default class DomProcessor {
     }
 
     _processMetaElement (el, urlReplacer, pattern) {
-        if (this.adapter.getAttr(el, 'http-equiv').toLowerCase() === 'refresh') {
+        var httpEquivAttrValue = this.adapter.getAttr(el, 'http-equiv').toLowerCase();
+
+        if (httpEquivAttrValue === 'refresh') {
             var attr = this.adapter.getAttr(el, pattern.urlAttr);
 
             attr = attr.replace(/(url=)(.*)$/i, (match, prefix, url) => prefix + urlReplacer(url));
 
             this.adapter.setAttr(el, pattern.urlAttr, attr);
+        }
+        // TODO: remove after https://github.com/DevExpress/testcafe-hammerhead/issues/244 implementation
+        else if (httpEquivAttrValue === 'content-security-policy') {
+            this.adapter.removeAttr(el, 'http-equiv');
+            this.adapter.removeAttr(el, 'content');
         }
     }
 
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -128,6 +128,9 @@ export default class ElementSandbox extends SandboxBase {
             setAttrMeth.apply(el, ns ? [ns, storedSandboxAttr, value] : [storedSandboxAttr, value]);
             value += ' allow-scripts';
         }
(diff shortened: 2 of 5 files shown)


src/processing/dom/index.js
```diff
@@ -216,6 +216,11 @@ export default class DomProcessor {
         this.adapter.setAttr(el, 'autocomplete', 'off');
     }
 
+    _processContentSecurityPolicy (el) {
```
inikulin: I believe it should be done in context of `_processMetaElement`

src/processing/dom/index.js
```diff
@@ -216,6 +216,11 @@ export default class DomProcessor {
...
+        if (this.adapter.getAttr(el, 'http-equiv').toLowerCase() === 'content-security-policy')
+            this.adapter.removeElement(el);
```
inikulin: Is it ok to remove element during traversing? Can it potentially brake anything? \cc @churkin

inikulin:
\r-

src/processing/dom/index.js
churkin: @inikulin right, you remove the array item in the foreach loop

inikulin: Maybe we should remove meta attributes instead, so it will be just skiped by the browser?

VasilyStrelyaev:
lgtm

src/processing/dom/index.js
miherlosev: > @inikulin right, you remove the array item in the foreach loop

we use `for` loop for selector patterns - https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/processing/dom/index.js#L171
We don't remove the array item in the foreach loop

miherlosev: ping @inikulin @churkin

inikulin:
@miherlosev What about https://github.com/DevExpress/testcafe-hammerhead/pull/246#discussion_r44264169 ?

miherlosev:
https://github.com/DevExpress/testcafe-hammerhead/pull/246#discussion_r44267905

src/processing/dom/index.js
inikulin: We need to use reverse for-loop then.

miherlosev:
ping @inikulin @churkin

inikulin:
@miherlosev https://github.com/DevExpress/testcafe-hammerhead/pull/246#discussion_r44280991

miherlosev:
After discussing with @inikulin  we decide to remove `http-equiv` and `content` attributes from the `meta` tag instead of removing a whole element.

miherlosev:
FPR

miherlosev:
ping @inikulin @churkin

miherlosev:
FPR

inikulin:
lgtm

inikulin:
ping @churkin

churkin:
lgtm

PR #254, opened 2015-11-11
The onDocumentCleaned event is raised wrongly after document.write is called for a document that is still being loaded (fix #253)

### LINKED ISSUE #253: The onDocumentCleaned event is raised wrongly after document.write is called for a document that is still being loaded
Multiple calls to  `document.write`  leads to the page hanging. Because each time`onDocumentCleaned` is raised Hammerhead gets initialized.

LavrovArtem:
/cc @inikulin @churkin

### DIFF
--- a/src/client/sandbox/node/document.js
+++ b/src/client/sandbox/node/document.js
@@ -3,7 +3,7 @@ import IframeSandbox from '../iframe';
 import INTERNAL_LITERAL from '../../../processing/script/internal-literal';
 import nativeMethods from '../native-methods';
 import * as htmlUtils from '../../utils/html';
-import { isFirefox, isIE } from '../../utils/browser';
+import { isFirefox, isIE, isIE9, isIE10 } from '../../utils/browser';
 import { isIframeWithoutSrc } from '../../utils/dom';
 
 export default class DocumentSandbox extends SandboxBase {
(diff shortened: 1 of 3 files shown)


src/client/sandbox/node/document.js
```diff
@@ -72,11 +76,11 @@ export default class DocumentSandbox extends SandboxBase {
             return null;
         }
 
-        var isUninitializedIframe = this._isUninitializedIframeWithoutSrc(this.window);
+        var needRiseDocumentCleaned = !this._isUninitializedIframeWithoutSrc() && this._isDocumentBeCleanedAfterWrite();
```
inikulin: `needRiseDocumentCleaned` -> `shouldEmitDocumentCleanedEvent`

inikulin: `_isDocumentBeCleanedAfterWrite` -> `_isDocumentCleanedAfterWrite`

inikulin: Actually you don't need method here, using variable will be just fine:
```js
var documentBeCleanedAfterWrite = this.document.readyState !== 'loading';
```

src/client/sandbox/node/document.js
```diff
@@ -16,16 +16,20 @@ export default class DocumentSandbox extends SandboxBase {
...
+            return this.window !== this.window.top && isIframeWithoutSrc(this.window.frameElement) &&
+                   !IframeSandbox.isIframeInitialized(this.window.frameElement);
         }
         catch (e) {
             return false;
         }
     }
 
+    _isDocumentBeCleanedAfterWrite () {
+        return this.document.readyState !== 'loading';
```
inikulin: Why this state indicates that document was cleaned?

inikulin:
\r-

src/client/sandbox/node/document.js
LavrovArtem: this state indicates that document will be cleaned after ```document.write```, because page alredy loaded

inikulin: Well, then it's declared as functions outside `_beforeDocumentCleaned` it doesn't makes much sense. Yet another evidence that it should be used as variable as I suggested in the comment below.

LavrovArtem:
FPR

inikulin:
lgtm

LavrovArtem:
@churkin for this condition
```javascript
(this.readyStateForIE || this.document.readyState) !== 'loading'
```
ie9/ie10 incorrect using only ```loading``` status, other status may be taken from ```this.document.readyState```

LavrovArtem:
FPR

inikulin:
lgtm

churkin:
lgtm

PR #266, opened 2015-11-17
Not all types of 'eval' function calls wrapped (close #260)

### LINKED ISSUE #260: Not all types of 'eval' function calls wrapped
We have to wrap:

```javascript
var temp = eval;

temp('some code');
```

to 

```javascript
var temp = eval === window.eval ? script => eval(__proc$Script(script)) : eval;

temp('some code');
```

churkin:
\cc @inikulin @LavrovArtem


test/server/script-processor-test.js
```diff
@@ -356,115 +356,177 @@ describe('Script processor', function () {
...
-            { src: 'eval("script")', expected: 'eval(__proc$Script("script"))' },
-            { src: 'window.eval(script)', expected: 'window.eval(__proc$Script(script))' },
-            { src: 'window["eval"](script)', expected: 'window["eval"](__proc$Script(script))' },
-
-            { src: 'eval.call(window, script)', expected: 'eval.call(window, __proc$Script(script))' },
-            { src: 'eval.call(window, "script")', expected: 'eval.call(window, __proc$Script("script"))' },
-            { src: 'window.eval.call(window, script)', expected: 'window.eval.call(window, __proc$Script(script))' },
+            {
+                src:      'eval(script)',
+                expected: '(eval === window.eval ? function(script, arg) { return window.eval(__proc$Script(script), arg); } : eval)(script)'
```
inikulin: I can't find `arg` in [spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.1.2.1). Is this some browser-specific implementation?

churkin: for the setTimeout and setInterval

inikulin: `setTimeout` and `setInterval` can accept variable number of arguments (https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/setInterval) and they don't have scope issues like `eval`, so seems like we should keep old transform for timers, and introduce new transform only for eval.

inikulin:
\r-

test/server/script-processor-test.js
```diff
@@ -356,115 +356,177 @@ describe('Script processor', function () {
...
+                expected: '(window.eval === window.eval ? function(script, arg) { return window.eval(__proc$Script(script), arg); } : window.eval).apply(window, [script])'
             },
             {
                 src:      'window["eval"].apply(window, [script])',
-                expected: 'window["eval"].apply(window, __proc$Script([script], true))'
+                expected: '(window["eval"] === window.eval ? function(script, arg) { return window.eval(__proc$Script(script), arg); } : window["eval"]).apply(window, [script])'
+            },
+            {
+                src:      'var tmp = eval;',
+                expected: 'var tmp = eval === window.eval ? function(script, arg) { return window.eval(__proc$Script(script), arg); } : eval;'
```
inikulin: Let's use `INTERNAL_LITERAL.tempVar` instead of `script`. Since `script` is too generic and we may shadow `script` var from outer scope for the `eval`. BTW, we should change literal value from `__set$temp` to `__temp$var` for consistency.

LavrovArtem:
Is the eval context lost in this case as if the window.eval is overridden?

inikulin:
@LavrovArtem nope, we use in-place stub, which preserves original lexical context.

LavrovArtem:
@inikulin In this case preserves lexical context of assing place. Is not it so?
```javascript
(function () {
	var x = function (text) { return eval(text); };
	var a = 10;
	
	alert(x('a === 10')); // success
	
	(function () {
		var b = 5;
		
		alert(x('a === 10')); // success
		alert(x('b === 5')); // fail
	})();
})();
```

test/server/script-processor-test.js
churkin: why do you think that they don't have scope issues like eval?

inikulin:
@LavrovArtem what?
@churkin Just learned that assigned `eval` always executed in the global scope, without capturing local context:
> If there is no calling context or if the eval code is not being evaluated by a direct call (15.1.2.1.1) to the eval function then,
Initialise the execution context as if it was a global execution context using the eval code as C as described in 10.4.1.1.

See: http://www.ecma-international.org/ecma-262/5.1/#sec-10.4.2

So, we can keep whole old transform and introduce eval-get transform:
```js
var foo = eval;
```
->
```js
var foo = __get$Eval(eval);
```

test/server/script-processor-test.js
inikulin: @churkin According to https://html.spec.whatwg.org/multipage/webappapis.html#timer-initialisation-steps timer script context should be set to `environment settings object`, which is, if I get right, is a global scope. Can you check, please, on your side in all browsers?

churkin:
@inikulin Then we have to contain 4 transformations instead of one:
- Get global eval `var foo = eval;`
- Get global setTimeout, setInterval `var foo = setTimeout;`
- Eval (setTimeout/setInterval ) invoking `window.eval()`
- Eval (setTimeout/ setInterval ) apply and call - `window.eval.apply(window, '')`

inikulin:
@churkin Anyway, if https://github.com/DevExpress/testcafe-hammerhead/pull/266#issuecomment-157641982 is correct, the following transform will be incorrect:
```js
var foo = (eval === window.eval ? function (scr) ...
```
Because we will have closure on local scope, but in case of assignment we should execute code in global scope. 
So, let's use two transforms:
1) For direct calls/apply()/call():
```js
(eval === window.eval ? function(script, arg) { return window.eval(__proc$Script(script), arg); } : eval)(script);
```

2)For assignments, to avoid lexical closures (need to check https://github.com/DevExpress/testcafe-hammerhead/pull/266#issuecomment-157641982):
```js
var foo = __get$EvalOrTimer(eval)
```

inikulin:
It would be nice to run transformed code against eval/timers webplatform tests in all browsers BTW, to be sure we done everything right.

inikulin:
However, previous having separate transforms for direct call/ apply() / call() is much better from the resulting code perspective. So I vote for keeping old transforms for direct calls and introducing new transform for assignments.

churkin:
@inikulin in the Chrome following code 
`(function() { var ev = eval, t = 'localScope'; return ev('t'); })()`
returns 'localScope'

churkin:
If we will use __get$Eval, we will lost local variables

test/server/script-processor-test.js
churkin: in the Chrome 
`(function() { var t = 'localScope'; window.setTimeout('alert(t)', 10) })()`
returns 'localScope'

inikulin:
In personal discussion we figure out that spec is correct. Assigned eval executes in the global scope (checked in Chrome, requires check in other browsers). So we need 2 transforms:
1)Keep old transform for direct call/call()/apply()
2)`__get$Eval()` transform, which will return `eval()` wrapper:
```js
function evalWrapper(script) {
   if (typeof script === 'string')
       processScript(script);
   
   // Assign eval to var to simulate original behavior
   var ev = eval;
   return eval(script);
}
```

churkin:
FPR

inikulin:
@churkin What about `var t = window.setTimeout` cases? Have you tried to run webplatform tests?

churkin:
@inikulin 
> Have you tried to run webplatform tests?

It seems they do not have tests for the eval.

> What about var t = window.setTimeout cases?

I thought we are not going to do it now. I can add them to the eval transform.

inikulin:
> I thought we are not going to do it now. I can add them to the eval transform.

It would be great.

> It seems they do not have tests for the eval.

Yep, `eval` is a part of ECMAScript spec and has nothing to do with the DOM

churkin:
@inikulin If `setTimeout` already uses global scope, maybe we can just use the 'override' mechanism:

`window.setTimeout = function(script ....) { nativeMethods.setTimeout(processScript(script )) }`

?

inikulin:
> If setTimeout always uses global scope, maybe we can just use the 'override' mechanism:

We need to check it first. BTW,I thought we already override timers

churkin:
@inikulin I have created separate issue for this #270

churkin:
@inikulin @LavrovArtem review pls

src/client/sandbox/code-instrumentation/index.js
```diff
@@ -31,6 +31,22 @@ export default class CodeInstrumentation extends SandboxBase {
         this.locationAccessorsInstrumentation.attach(window);
         this.elementPropertyAccessors = this.propertyAccessorsInstrumentation.attach(window);
 
+        window[INSTRUCTION.getEval] = ev => {
```
inikulin: Let's rename `ev` arg to `evalFn`

src/client/sandbox/code-instrumentation/index.js
```diff
@@ -31,6 +31,22 @@ export default class CodeInstrumentation extends SandboxBase {
...
+            if (ev !== window.eval)
+                return ev;
+
+            return script => {
+                if (typeof script === 'string')
+                    script = processScript(script);
+
+                // NOTE: Assign the eval method to a variable to simulate the original behavior -
+                // eval switches to the global context. (GH-260)
+                var evl = eval;
```
inikulin: We can avoid doing it if we'll use `evalFn`. As I understand `eval` will switch to global context mode if you pass it as parameter (requires check)

test/client/fixtures/sandbox/code-instrumentation/process-script-test.js
```diff
@@ -78,3 +78,14 @@ test('should not process non-string args', function () {
     deepEqual(execScript('eval.apply(this, [[1, 2 ,3]])'), [1, 2, 3]);
 });
 
+module('var foo = eval');
```
inikulin: Let's call it `eval assignment`

test/client/fixtures/sandbox/code-instrumentation/process-script-test.js
```diff
@@ -78,3 +78,14 @@ test('should not process non-string args', function () {
...
+
+test('The eval method should switch its own context to global', function () {
+    throws(function () {
```
inikulin: We need more test cases:
1)return eval from the scope - it's shouldn't have closure
2)pass eval as the func argument - it's shouldn't have closure on the upper scope

test/client/fixtures/sandbox/code-instrumentation/process-script-test.js
```diff
@@ -78,3 +78,14 @@ test('should not process non-string args', function () {
...
+        execScript('(function(){var t = "localScope", ev = eval; return ev("t"); })()');
+    });
+});
+
+test('The script processor should process eval\'s global', function () {
+    deepEqual(execScript('var ev = eval; ev("1+1")'), 2);
```
inikulin: This test doesn't test what script processing works for the assigned eval

test/server/script-processor-test.js
```diff
@@ -379,6 +379,26 @@ describe('Script processor', function () {
...
+            },
+            {
+                src:      '{a: eval, b:window.eval, c: window["eval"]}',
+                expected: '{a: __get$Eval(eval), b:__get$Eval(window.eval), c: __get$Eval(window["eval"])}'
+            },
+            {
+                src:      '(function() {return eval;}, function(){return window.eval;}, function(){return window["eval"]})',
+                expected: '(function() {return __get$Eval(eval);}, function(){return __get$Eval(window.eval);}, function(){return __get$Eval(window["eval"]);})'
+            },
+            {
```
inikulin: Add test cases then it's passed as an argument

inikulin:
\r-

churkin:
FPR

churkin:
Tests are falling because of saucelabs problems

test/client/fixtures/sandbox/code-instrumentation/process-script-test.js
```diff
@@ -78,3 +80,29 @@ test('should not process non-string args', function () {
     deepEqual(execScript('eval.apply(this, [[1, 2 ,3]])'), [1, 2, 3]);
 });
 
+module('eval assignment');
+
+test('the eval method should switch its own context to global', function () {
+    throws(function () {
+        execScript('(function(){ var t1 = "localScope", ev = eval; return ev("t1"); })()');
```
inikulin: Just thought what it would be better to use the following tactic in these tests:
Instead of expecting the throw, which may occur because of many different reasons, let's define var with the same name in the global scope, but with different value. So we will have the following test:
```js
 var actual = execScript(' window.t1 = 'globalScope'; var res = (function(){ var t1 = "localScope", ev = eval; return ev("t1"); })(); delete window.t1; res');

strictEqual(actual, 'globalScope');
```

inikulin: \cc @churkin

churkin:
@inikulin ping

test/client/fixtures/sandbox/code-instrumentation/process-script-test.js
churkin: agree

churkin:
FPR

inikulin:
lgtm

LavrovArtem:
lgtm

PR #274, opened 2015-11-19
Update to v.2.0.1

churkin:
\cc @inikulin @AlexanderMoskovkin @helen-dikareva

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "testcafe-hammerhead",
   "description": "A powerful web-proxy used as a core for the TestCafe testing framework (https://github.com/DevExpress/testcafe).",
-  "version": "2.0.0",
+  "version": "2.0.1",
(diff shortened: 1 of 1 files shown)


inikulin:
lgtm, don't forget about git tag

inikulin:
Fail in active window tracker looks not accidental. \cc @helen-dikareva

helen-dikareva:
Ok, I'll see

helen-dikareva:
I could not reproduce test failure locally on MacBook's iPhone simulator and with manually created virtual machine on SauceLabs. So tests deserve another chance.
sauce_tunnel
sauce_tunnel2

inikulin:
@helen-dikareva However, there are still fails that involves active window tracker. Seems like some timing issue, which reproduces only on slow machines.

AlexanderMoskovkin:
lgtm

PR #291, opened 2015-11-25
Hammerhead client proxy handle document.location set incorrectly (close #290)

### LINKED ISSUE #290: document.location for the top window is proxied incorrectly
If we set document.location for the top window from iframe: 

```javascript
document.location = 'http://some.domain.com/';
```

it is proxied as for the iframe:

`http://hammerhead:1400/sessionId!`**iframe**`/http://some.domain.com/`

churkin:
Unfortunately, we can not write a client-test for this bug, since the bug is associated with a changing location property of the top window (If we set `document.location` for the top window, it always proxied with 'iframe' resource type).

\cc @inikulin @AlexanderMoskovkin

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -288,7 +288,8 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
                         if (window.self !== window.top)
                             location = destLocation.resolveUrl(location, window.top.document);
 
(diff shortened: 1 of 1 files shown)


churkin:
@testcafe-build-bot \retest

churkin:
FPR

inikulin:
@churkin Can you explain in the original issue what this bug is about, please? I can't review it then I don't understand the reasoning.

inikulin:
lgtm

AlexanderMoskovkin:
lgtm

PR #307, opened 2015-12-02
Process pages with status 204 and return status 200 instead (close #306)

### LINKED ISSUE #306: After we process a page with status 204, it is loaded for an infinite time on the client
Status 204 means that the resource has no content. After we process a page, it gets a non-empty body, which we send to the browser. If the page is sent with status 200, everything works as expected.

LavrovArtem:
/cc @inikulin @churkin

### DIFF
--- a/src/request-pipeline/context.js
+++ b/src/request-pipeline/context.js
@@ -145,12 +145,11 @@ export default class RequestPipelineContext {
 
         var isCSS      = contentTypeUtils.isCSSResource(contentType, accept);
         var isManifest = contentTypeUtils.isManifest(contentType);
-        var isJSON     = contentTypeUtils.isJSON(contentType);
         var isScript   = this.dest.resourceType === urlUtils.SCRIPT ||
                          contentTypeUtils.isScriptResource(contentType, accept);
 
-        var requireProcessing = !this.isXhr &&
-                                (this.isPage || this.isIframe || isCSS || isScript || isManifest || isJSON);
+        var requireAssetsProcessing = (isCSS || isScript || isManifest) && this.destRes.statusCode !== 204;
+        var requireProcessing       = !this.isXhr && (this.isPage || this.isIframe || requireAssetsProcessing);
 
         var isIframeWithImageSrc = this.isIframe && !this.isPage && /^\s*image\//.test(contentType);
 
@@ -175,7 +174,6 @@ export default class RequestPipelineContext {
             isCSS,
             isScript,
             isManifest,
-            isJSON,
             encoding,
             contentTypeUrlToken
         };
--- a/src/request-pipeline/index.js
+++ b/src/request-pipeline/index.js
@@ -40,6 +40,9 @@ var stages = {
     3: function decideOnProcessingStrategy (ctx, next) {
         ctx.buildContentInfo();
 
+        if (ctx.contentInfo.requireProcessing && ctx.destRes.statusCode === 204)
+            ctx.destRes.statusCode = 200;
+
         // NOTE: Just pipe the content body to the browser if we don't need to process it.
         if (!ctx.contentInfo.requireProcessing) {
             sendResponseHeaders(ctx);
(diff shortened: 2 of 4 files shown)


src/request-pipeline/index.js
```diff
@@ -113,7 +113,11 @@ function createReqOpts (ctx) {
 function sendResponseHeaders (ctx) {
     var headers = headerTransforms.forResponse(ctx);
 
-    ctx.res.writeHead(ctx.destRes.statusCode, headers);
+    if (ctx.contentInfo.requireProcessing && ctx.destRes.statusCode === 204 && headers['content-length'] > 0)
```
inikulin: Wouldn't it be better to not process content with 204 status at all? BTW, how browser behave if the receive 204 for the page? Do they perform page transfer? According to spec, browser should just updated meta information from headers. If it's shows blank page, then we can signal it as fatal error for the page, but otherwise we should keep things as is.

inikulin:
\r-

src/request-pipeline/index.js
LavrovArtem: If we not process page with 204 status, then hammerhead will be not initialized. [Description](https://github.com/DevExpress/testcafe-hammerhead/issues/306)

LavrovArtem:
FPR

src/request-pipeline/context.js
```diff
@@ -145,12 +145,11 @@ export default class RequestPipelineContext {
...
         var isCSS      = contentTypeUtils.isCSSResource(contentType, accept);
         var isManifest = contentTypeUtils.isManifest(contentType);
-        var isJSON     = contentTypeUtils.isJSON(contentType);
         var isScript   = this.dest.resourceType === urlUtils.SCRIPT ||
                          contentTypeUtils.isScriptResource(contentType, accept);
 
-        var requireProcessing = !this.isXhr &&
-                                (this.isPage || this.isIframe || isCSS || isScript || isManifest || isJSON);
+        var requireProcessing = !this.isXhr && (this.isPage || this.isIframe ||
+                                                (isCSS || isScript || isManifest) && this.destRes.statusCode !== 204);
```
inikulin: Can we simplify this condition somehow?

test/server/proxy-test.js
```diff
@@ -666,6 +675,25 @@ describe('Proxy', function () {
             });
         });
 
+        it('Should process pages with status 204 and return status 200 instead (GH-306)', function (done) {
```
inikulin: We need test this for non-page scenario as well.

inikulin:
\r-

src/request-pipeline/context.js
```diff
@@ -145,12 +145,11 @@ export default class RequestPipelineContext {
 
         var isCSS      = contentTypeUtils.isCSSResource(contentType, accept);
         var isManifest = contentTypeUtils.isManifest(contentType);
-        var isJSON     = contentTypeUtils.isJSON(contentType);
         var isScript   = this.dest.resourceType === urlUtils.SCRIPT ||
                          contentTypeUtils.isScriptResource(contentType, accept);
 
-        var requireProcessing = !this.isXhr &&
-                                (this.isPage || this.isIframe || isCSS || isScript || isManifest || isJSON);
+        var requireNonPageProcessing = (isCSS || isScript || isManifest) && this.destRes.statusCode !== 204;
```
inikulin: Let' rename it to `requireAssetsProcessing`

inikulin:
\r-

test/server/proxy-test.js
```diff
@@ -666,6 +675,56 @@ describe('Proxy', function () {
...
+
+                compareCode(body, expected);
+                expect(res.statusCode).eql(200);
+                expect(res.headers['content-length']).eql(body.length.toString());
+
+                done();
+            });
+        });
+
+        it('Should not process non-page resources with status 204 (GH-306)', function (done) {
```
VasilyStrelyaev:
```
        it('Should not process assets with status 204 (GH-306)', function (done) {
```

LavrovArtem:
FPR

inikulin:
lgtm

VasilyStrelyaev:
lgtm

churkin:
lgtm

PR #326, opened 2015-12-09
FormData.prototype.append must be overridden (close #324)

### LINKED ISSUE #324: FormData.prototype.append must be overridden
If we append our file wrapper to ```FormData```, we will lose the file name. And we should not send our hidden input's value along with the file info, because our input may have incorrect value if the input with the file has been removed from DOM.

LavrovArtem:
/cc @churkin @miherlosev


LavrovArtem:
FPR

LavrovArtem:
FPR

churkin:
lgtm

miherlosev:
lgtm

PR #343, opened 2015-12-29
fix urlUtils.getProxyUrl always add the trailing slash (close #342)

### LINKED ISSUE #342: urlUtils.getProxyUrl always add the trailing slash
For example:
urlUtils.getProxyUrl('http://google.com') -> "http://<host>:<port>/<sessionId>/http://google.com/"

Reproduced in IE, Chrome, FF

miherlosev:
/cc @churkin @LavrovArtem

### DIFF
--- a/src/client/utils/destination-location.js
+++ b/src/client/utils/destination-location.js
@@ -1,6 +1,7 @@
 import * as sharedUrlUtils from '../../utils/url';
 import * as domUtils from './dom';
 import { get as getSettings } from '../settings';
+import { endsWith } from '../../utils/string';
 
 const DOCUMENT_URL_RESOLVER = 'hammerhead|document-url-resolver';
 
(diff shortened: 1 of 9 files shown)


src/client/utils/destination-location.js
```diff
@@ -66,7 +67,13 @@ export function resolveUrl (url, doc) {
             return resolveUrl(url, window.parent.document);
     }
 
-    return urlResolver.href;
+    var result           = urlResolver.href;
+    var hasTrailingSlash = endsWith(url, '/');
```
LavrovArtem: may be use
```javascript
url.lastIndexOf('/') === url.length - 1
```

miherlosev:
`endsWith` and `trim` group into `utils/string`.
I use the @LavrovArtem variant for `endsWith` implementation.
It is breaking change.
FPR

churkin:
lgtm

miherlosev:
for `endsWith` use imprlementation from http://stackoverflow.com/questions/280634/endswith-in-javascript
FPR

LavrovArtem:
lgtm

PR #348, opened 2015-12-31
override querySelector, querySelectorAll for document, element, documentFragment nodes (close #340)

### LINKED ISSUE #340: Method 'querySelectorAll' should be overrided to use JQuery selectors that returs elements by value of attribute contains URL
When test contains selector like this '[href="https://www.google.ru/"]' (or other selector based on attr) jQuery try to get element (by Sizzle) and just call ```document.querySelectorAll(selector)```
Check please may be we have other proxed attr.

miherlosev:
/cc @churkin @LavrovArtem 
@VasilyStrelyaev

### DIFF
--- a/src/client/sandbox/node/index.js
+++ b/src/client/sandbox/node/index.js
@@ -3,7 +3,11 @@ import SandboxBase from '../base';
 import WindowSandbox from './window';
 import DocumentSandbox from './document';
 import ElementSandbox from './element';
-import { parseDocumentCharset } from '../../utils/dom';
+import nativeMethods from '../native-methods';
+import domProcessor from '../../dom-processor';
+import { parseDocumentCharset, isDocumentFragment } from '../../utils/dom';
+
+const ATTRIBUTE_SELECTOR_REG_EX = /\[([\w-]+)(\^?=.+?)]/g;
 
 export default class NodeSandbox extends SandboxBase {
     constructor (nodeMutation, iframeSandbox, eventSandbox, uploadSandbox, shadowUI) {
@@ -52,7 +56,9 @@ export default class NodeSandbox extends SandboxBase {
             // OPTIMIZATION: Use querySelectorAll to iterate through descendant nodes.
             this._overrideElement(el);
 
-            var children = el.querySelectorAll('*');
+            var nativeQuerySelectorAll = isDocumentFragment(el) ? nativeMethods.documentFragmentQuerySelectorAll
+                                                                : nativeMethods.elementQuerySelectorAll;
+            var children = nativeQuerySelectorAll.call(el, '*');
 
             for (var i = 0; i < children.length; i++)
                 this._overrideElement(children[i]);
@@ -100,5 +106,22 @@ export default class NodeSandbox extends SandboxBase {
         this.doc.attach(window, document);
         this.win.attach(window);
         this.element.attach(window);
+    }
+
+    static processSelector (selector) {
+        selector = String(selector);
+
+        if (!ATTRIBUTE_SELECTOR_REG_EX.test(selector))
+            return selector;
+
+        return selector.replace(ATTRIBUTE_SELECTOR_REG_EX, (str, name, operatorWithValue) => {
+            if (domProcessor.URL_ATTRS.indexOf(name) !== -1) {
+                name  = domProcessor.getStoredAttrName(name);
+
+                return '[' + name + operatorWithValue + ']';
+            }
+
+            return str;
+        });
     }
 }
--- a/test/client/fixtures/sandbox/node/document-test.js
+++ b/test/client/fixtures/sandbox/node/document-test.js
@@ -123,6 +123,141 @@ if (!browserUtils.isFirefox) {
     });
 }
 
+module('querySelector, querySelectorAll (GH-340)');
+
+test('quote types in attribute selectors', function () {
+    var link = document.createElement('a');
+
+    link.setAttribute('href', 'http://some.domain.com');
+    document.body.appendChild(link);
+
+    ok(document.querySelector('[href="http://some.domain.com"]'));
+    ok(document.querySelector("[href='http://some.domain.com']"));
+
+    link.parentNode.removeChild(link);
+});
+
+test('non-processed attributes', function () {
+    var link = document.createElement('a');
+
+    link.setAttribute('data-info', 'external link');
+    link.setAttribute('hreflang', 'ru-RU');
+    document.body.appendChild(link);
+
+    ok(document.querySelector('[data-info~=external]'));
+    ok(document.querySelector('[hreflang|=ru]'));
+
+    link.parentNode.removeChild(link);
+});
+
+//http://www.w3.org/TR/css3-selectors/#attribute-selectors
+test('attrubute types', function () {
+    var link = document.createElement('a');
+    var div  = document.createElement('div');
+
+    link.setAttribute('href', 'http://some.domain.com');
+    div.className = 'container';
+    div.appendChild(link);
+
+    document.body.appendChild(div);
+
+    // [attribute]
+    ok(div.querySelector('[href]'));
+
+    // [attribute=value]
+    ok(document.querySelector('[href="http://some.domain.com"]'));
+
+    // [attribute~=value] - whitespace-separated values
+    // Proxied attributes don't contain whitespace-separated values
+
+    // [attribute|=value] - equal or starts with for value that ends with '-'
+    // This is primarily intended to allow language subcode matches
+
+    // [attribute^=value] - starts with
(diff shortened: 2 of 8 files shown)


src/client/sandbox/node/index.js
```diff
@@ -101,4 +104,25 @@ export default class NodeSandbox extends SandboxBase {
         this.win.attach(window);
         this.element.attach(window);
     }
+
+    static _processSelector (selector) {
+        var attributeSelectorRegEx = /\[([\w-^]+)=(.+)]/;
```
LavrovArtem: What about selector '[action=url] [href=url]'? You must use global flag  /\[([\w-^]+)=(.+)]/g

miherlosev:
For me
Search web-platform tests for querySelector, querySelectorAll and run them.

miherlosev:
Checked tests:
http://w3c-test.org/selectors/attribute-selectors/attribute-case/cssom.html

inikulin:
@miherlosev You need to run http://w3c-test.org/dom/nodes/ParentNode-querySelector-All.html

miherlosev:
Waiting for https://github.com/DevExpress/testcafe-hammerhead/issues/351

miherlosev:
FPR

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,36 @@ export default class NodeSandbox extends SandboxBase {
...
+            if (value === null)
+                return 'null';
+
+            else if (value === void 0)
+                return 'undefined';
+
+            return selector;
+        };
+
+        selector = stringifySelector(selector);
```
inikulin: `stringifySelector(selector)` === `String(selector)`

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,36 @@ export default class NodeSandbox extends SandboxBase {
...
+
+        return selector.replace(attributeSelectorRegEx, (str, name, value) => {
```
inikulin: In case then we don't have attribute selector is `replace` equivalent in speed  to `test`? Can you measure using https://jsperf.com/? If not, we should add basic filter with `test`

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,36 @@ export default class NodeSandbox extends SandboxBase {
         this.win.attach(window);
         this.element.attach(window);
     }
+
+    static processSelector (selector) {
+        var attributeSelectorRegEx = /\[([\w-^]+)=(.+?)]/g;
```
inikulin: Should be a constant. RegExp's has `RE` or `Re` suffix in our code style.

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,36 @@ export default class NodeSandbox extends SandboxBase {
...
+            return selector;
+        };
+
+        selector = stringifySelector(selector);
+
+        return selector.replace(attributeSelectorRegEx, (str, name, value) => {
+            var isAttributeValueBeginWithSelector = endsWith(name, '^');
+
+            name = name.replace(/\^$/g, '');
+            if (domProcessor.URL_ATTRS.indexOf(name) !== -1) {
```
inikulin: Need new line before `if`

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,36 @@ export default class NodeSandbox extends SandboxBase {
...
+                value = value.replace(/^['"]|['"]$/g, '');
+                value = urlUtils.getProxyUrl(value);
```
inikulin: This may throw. We should check for the valid, proxiable URL as well. It could be `src=javascript:void(0)`

inikulin:
\r-

src/client/sandbox/node/index.js
miherlosev: `string.replace` is slower than `regex.test`  - http://jsperf.com/string-replace-versus-regex-test
(Chrome - 78%, FF - 20%, IE11 - 50%.)

inikulin: Let's perform `test` first then.

src/client/sandbox/node/index.js
miherlosev: Inside `urlUtils.getProxyUrl` check for proxiable URL already exists - https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/url.js#L11.
Also I wrote the test for this case.

miherlosev:
FPR

test/client/fixtures/sandbox/node/document-test.js
```diff
@@ -123,6 +123,142 @@ if (!browserUtils.isFirefox) {
...
+    document.body.appendChild(div);
+
+    // [attribute]
+    ok(div.querySelector('[href]'));
+
+    // [attribute=value]
+    ok(document.querySelector('[href=http://some.domain.com]'));
+
+    // [attribute~=value] - writespace separeted values
+    // Proxed attributes don't contain writespace separated values
```
VasilyStrelyaev:
```
    // Proxied attributes don't contain whitespace-separated values
```

miherlosev:
FPR

test/client/fixtures/sandbox/node/document-test.js
```diff
@@ -123,6 +123,142 @@ if (!browserUtils.isFirefox) {
...
+
+    document.body.appendChild(div);
+
+    // [attribute]
+    ok(div.querySelector('[href]'));
+
+    // [attribute=value]
+    ok(document.querySelector('[href=http://some.domain.com]'));
+
+    // [attribute~=value] - writespace separeted values
```
VasilyStrelyaev: sorry, one more
```
    // [attribute~=value] - whitespace-separated values
```

VasilyStrelyaev:
lgtm

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,31 @@ export default class NodeSandbox extends SandboxBase {
         this.win.attach(window);
         this.element.attach(window);
     }
+
+    static processSelector (selector) {
+        const ATTRIBUTE_SELECTOR_REG_EX = /\[([\w-^]+)=(.+?)]/g;
```
inikulin: It should be module-level re I guess, no need to declare it everytime we call the funciton

src/client/sandbox/node/index.js
```diff
@@ -101,4 +107,31 @@ export default class NodeSandbox extends SandboxBase {
...
+
+        selector = String(selector);
+
+        if (!ATTRIBUTE_SELECTOR_REG_EX.test(selector))
+            return selector;
+
+        return selector.replace(ATTRIBUTE_SELECTOR_REG_EX, (str, name, value) => {
+            var isAttributeValueBeginWithSelector = endsWith(name, '^');
```
inikulin: `isAttributeValueBeginWithSelector ` this is completely confusing. First of all, more likely it's `isAttrNameEndsWithSelector`. The best way will be to modify the re: 
```js
/\[([\w-^]+)(^?=)(.+?)]/g
```
You will have additional group in the middle - the `operator`. You can use it then to restore the selector.

src/client/sandbox/node/index.js
inikulin: Have you seen implementation of `isSupportedProtocol`? It doesn't ensure that we have a URL. We should check for `http://`, `https://` or `//` at the begining of the string. Seems like we should make this modification in `isSupportedProtocol`, but I', not sure. \cc @churkin

inikulin:
See also: https://github.com/DevExpress/testcafe-hammerhead/pull/348#discussion_r49736383

\r-

src/client/sandbox/node/index.js
miherlosev: I don't know scenario than this may throw.
Url value can begins not only with `http://`, `https://`, `//`.
For example:
```javascript
link.setAttribute('href', 'index.html');
document.querySelector('[href=index.html]');
```

churkin:
lgtm

src/client/sandbox/node/index.js
inikulin: We can't do it this way then. If we will convert any value to proxy URL then we have a huge amount of things that can brake

inikulin:
\r-

inikulin:
See also: https://github.com/DevExpress/testcafe-hammerhead/pull/348#discussion_r49835800

miherlosev:
FPR

inikulin:
@miherlosev Have you tried WebPlatform tests with this iteration?

miherlosev:
Yes. It's ok.

inikulin:
lgtm then

LavrovArtem:
lgtm

PR #352, opened 2016-01-12
fix 'Url processing on server side always adds traling slash for proxy url' (close #351)

### LINKED ISSUE #351: Url processing on server side always adds traling slash for proxy url
Problem located inside standard module `url`.
Example:
`require('url').resolve('', 'http://www.example.org/') -> 'http://www.example.org/'`
`require('url').resolve('', 'http://www.example.org') -> 'http://www.example.org/'`

It causes problem for `element.querySelector([attribute$=value])`.
Selector `[attribute$=value]` means that attribute value ends with specified value.

miherlosev:
/cc @churkin @LavrovArtem

### DIFF
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -3,7 +3,7 @@
 // Do not use any browser or node-specific API!
 // -------------------------------------------------------------
 
-import { trim } from './string';
+import * as stringUtils from './string';
 
 //Const
 const PROTOCOL_RE        = /(^(\w+?\:))/;
@@ -146,7 +146,7 @@ export function parseUrl (url) {
     if (!url)
         return parsed;
 
-    url = trim(url);
+    url = stringUtils.trim(url);
 
     // Protocol
     var hasImplicitProtocol = false;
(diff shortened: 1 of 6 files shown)


miherlosev:
FPR

src/utils/url.js
```diff
@@ -241,3 +241,12 @@ export function prepareUrl (url) {
     // For example, the "//////google.com" url is equal to "//google.com".
     return url.replace(/^\/+(\/\/.*$)/, '$1');
 }
+
+export function fixUrlAfterResolving (urlBefore, urlAfter) {
```
inikulin: It should be `ensureTrailingSlash` or something like that. I also don't like arg names. Maybe `srcUrl` and `processedUrl` or smthg like this. 
\cc @VasilyStrelyaev 
Overall this feels extremely hacky, mostly because currently URL-related stuff is a completely fucked up. But  for now, this is just fine

inikulin:
\r-

src/utils/url.js
VasilyStrelyaev: ok

miherlosev:
FPR

inikulin:
lgtm

LavrovArtem:
lgtm

churkin:
lgtm

PR #367, opened 2016-01-21
Quotes in the cookies are not escaped when a task script for an iframe is built on the client (close #366)

### LINKED ISSUE #366: Quotes in the cookies are not escaped when a task script for an iframe is built on the client
http://66.55.147.18:13500/history?timestamp=1453358894078&siteUrl=http://www.flipkart.com/

churkin:
\сс @LavrovArtem @miherlosev

### DIFF
--- a/test/client/fixtures/sandbox/iframe-test.js
+++ b/test/client/fixtures/sandbox/iframe-test.js
@@ -1,4 +1,5 @@
-var urlUtils       = hammerhead.get('./utils/url');
+var urlUtils = hammerhead.get('./utils/url');
+var settings = hammerhead.get('./settings');
 
 var iframeSandbox = hammerhead.sandbox.iframe;
 var nativeMethods = hammerhead.nativeMethods;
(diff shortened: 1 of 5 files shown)


inikulin:
Regression tests?

churkin:
FPR

test/client/fixtures/sandbox/iframe-test.js
```diff
@@ -201,6 +202,23 @@ asyncTest('native methods are properly initialized in an iframe without src (GH-
     document.body.appendChild(iframe);
 });
 
+asyncTest('quotes in the cookies are not escaped when a task script for an iframe is built on the client (GH-366)', function () {
+    var iframe = document.createElement('iframe');
+    var cookie = '""\'\'';
+
+    iframe.id             = 'test_cookie';
+    settings.get().cookie = cookie;
+
```
miherlosev: Before `start` calling we need to restore origin `settings.get().cookie` value.

miherlosev:
\r-

test/client/fixtures/sandbox/iframe-test.js
churkin: We do not have default cookie value, so I think it is not necessary

miherlosev: So, ok

LavrovArtem:
lgtm

PR #380, opened 2016-01-27
Get rid of the horrible code

churkin:
\cc @inikulin


inikulin:
@churkin Don't blame yourself, it's not that horrible and can happen with everyone =) The really frustrating thing is that nobody had mentioned it on review =(.

inikulin:
lgtm, waiting for tests

PR #381, opened 2016-01-27
Ignore IE11's and Edge's service handlers (close #379)

### LINKED ISSUE #379: Error occurs in _getEventListenerWrapper method in IE11
Added listener doesn't have 'call' method after removing related element

churkin:
\cc @inikulin @LavrovArtem

### DIFF
--- a/src/client/sandbox/event/listeners.js
+++ b/src/client/sandbox/event/listeners.js
@@ -54,6 +54,12 @@ export default class Listeners extends EventEmitter {
 
     static _getEventListenerWrapper (eventCtx, listener) {
         return function (e) {
+            var isIEServiceHandler = listener.toString() === '[object FunctionWrapper]';
+
+            // NOTE: Ignore IE11's and Edge's service handlers (GH-379)
+            if (isIEServiceHandler)
+                return null;
+
             if (eventCtx.cancelOuterHandlers)
                 return null;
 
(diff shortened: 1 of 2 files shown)


src/client/sandbox/event/listeners.js
```diff
@@ -54,16 +54,20 @@ export default class Listeners extends EventEmitter {
...
+                        return listener.handleEvent.call(listener, e);
 
-            if (isObjectEventListener(listener))
-                return listener.handleEvent.call(listener, e);
+                    return listener.call(this, e);
+                }
+            }
 
-            return listener.call(this, e);
+            return null;
```
inikulin: I don't get it. It will not work with regular handlers?

inikulin: Got it now.

src/client/sandbox/event/listeners.js
```diff
@@ -54,16 +54,20 @@ export default class Listeners extends EventEmitter {
 
     static _getEventListenerWrapper (eventCtx, listener) {
         return function (e) {
-            if (eventCtx.cancelOuterHandlers)
-                return null;
+            // NOTE: Ignore IE11's and Edge's service handlers (GH-379)
+            if (listener.toString() !== '[object FunctionWrapper]') {
```
inikulin: I'd rather use variable here for clarity:
```js
var isIEServiceHandler = listener.toString() === '[object FunctionWrapper]';

if (!isIEServiceHandler) {
 ...
```

or revert logic: 
```js
if (listener.toString() === '[object FunctionWrapper]')
    return;

...
```

inikulin:
This requires regression test. \r-

churkin:
FPR

LavrovArtem:
lgtm

inikulin:
lgtm

Issue #20, opened 2015-08-05
Refactor client url resolver


churkin:
Exclude url resolver from the url utils to the separate url-resolver.js file.

miherlosev:
done in #383 pull request

PR #383, opened 2016-01-28
fix 'Wrong url resolving for page with <base> tag' (close #371)

### LINKED ISSUE #371: Wrong url resolving for page with <base> tag
http://66.55.147.18:13500/history?timestamp=1453575754575&siteUrl=http://www.alibaba.com/

Test for reproduction:

```javascript
test('Maximum call stack size exceeded', function () {
    var baseEl = document.createElement('base');

    baseEl.href = 'http://localhost:1337/sessionId/http://test.example.com';
    document.head.appendChild(baseEl);

    strictEqual(urlUtils.getProxyUrl('{{inquiry.imgSrc}}-->'), // <-- Maximum call stack size exceeded
        'http://localhost:1337/sessionId/http://test.example.com/%7B%7Binquiry.imgSrc%7D%7D--%3E');

    document.head.removeChild(baseEl);
});
```

miherlosev:
/cc @inikulin @churkin

### DIFF
--- a/src/client/sandbox/index.js
+++ b/src/client/sandbox/index.js
@@ -18,6 +18,7 @@ import XhrSandbox from './xhr';
 import StorageSandbox from './storages';
 import { isIE, isWebKit } from '../utils/browser';
 import { create as createSandboxBackup, get as getSandboxBackup } from './backup';
+import urlResolver from '../utils/url-resolver';
 
 export default class Sandbox extends SandboxBase {
     constructor () {
@@ -115,6 +116,8 @@ export default class Sandbox extends SandboxBase {
         if (isIE || isWebKit)
             this._refreshNativeMethods(window, document);
 
+        urlResolver.init(document);
+
         this.event.initDocumentListening();
 
         if (isWebKit)
@@ -128,6 +131,8 @@ export default class Sandbox extends SandboxBase {
 
     attach (window) {
         super.attach(window);
+
+        urlResolver.init(document);
 
         // NOTE: Eval Hammerhead code script.
         this.iframe.on(this.iframe.EVAL_HAMMERHEAD_SCRIPT, e => initHammerheadClient(e.iframe.contentWindow, true));
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -7,7 +7,8 @@ import { process as processStyle } from '../../../processing/style';
 import * as urlUtils from '../../utils/url';
 import * as domUtils from '../../utils/dom';
 import * as hiddenInfo from '../upload/hidden-info';
-import { sameOriginCheck } from '../../utils/destination-location';
+import * as urlResolver from '../../utils/url-resolver';
+import { sameOriginCheck, get as getDestLocation } from '../../utils/destination-location';
 import { stopPropagation } from '../../utils/event';
 import { isPageHtml, processHtml } from '../../utils/html';
 import { waitCookieMsg } from '../../transport';
@@ -105,6 +106,9 @@ export default class ElementSandbox extends SandboxBase {
                     else if (isIframe || domProcessor.isOpenLinkInIframe(el))
                         resourceType = urlUtils.IFRAME;
 
+                    if (ElementSandbox._isHrefAttrForBaseElement(el, attr))
+                        urlResolver.updateBase(value, this.document);
+
                     value = isIframe && isCrossDomainUrl ? urlUtils.getCrossDomainIframeProxyUrl(value) :
                             urlUtils.getProxyUrl(value, null, null, null, resourceType, elCharset);
                 }
@@ -148,6 +152,9 @@ export default class ElementSandbox extends SandboxBase {
             else
                 removeAttrFunc.apply(el, ns ? [arg[0], storedAttr] : [storedAttr]);
         }
+
+        if (ElementSandbox._isHrefAttrForBaseElement(el, attr))
+            urlResolver.updateBase(getDestLocation(), this.document);
 
         if (attr !== 'autocomplete')
             return removeAttrFunc.apply(el, arg);
@@ -326,6 +333,10 @@ export default class ElementSandbox extends SandboxBase {
         return domProcessor.URL_ATTR_TAGS[attr] && domProcessor.URL_ATTR_TAGS[attr].indexOf(tagName) !== -1;
     }
 
+    static _isHrefAttrForBaseElement (el, attr) {
+        return domUtils.isBaseElement(el) && attr === 'href';
+    }
+
     static _removeFileInputInfo (el) {
         hiddenInfo.removeInputInfo(el);
     }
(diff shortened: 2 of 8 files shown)


miherlosev:
@testcafe-build-bot \retest

src/client/sandbox/index.js
```diff
@@ -18,6 +18,8 @@ import XhrSandbox from './xhr';
 import StorageSandbox from './storages';
 import { isIE, isWebKit } from '../utils/browser';
 import { create as createSandboxBackup, get as getSandboxBackup } from './backup';
+import { get } from '../utils/destination-location';
```
inikulin: You shouldn't export single method, since it's lacks context completely. Export `destinationLocation` as a whole

src/client/utils/url-resolver.js
```diff
@@ -0,0 +1,59 @@
+import nativeMethods from '../sandbox/native-methods';
+import { ensureTrailingSlash } from '../../utils/url';
+import { get } from '../utils/destination-location';
+
+const DOCUMENT_URL_RESOLVER = 'hammerhead|document-url-resolver';
+
+function createResolver (doc) {
```
inikulin: This module should be an object and not the set of the utility methods. Since now it's single entity - `urlResolver`

src/client/utils/url-resolver.js
```diff
@@ -0,0 +1,59 @@
...
+
+    return htmlDocument;
+}
+
+function getResolver (doc) {
+    // NOTE: IE cleans the document up after document.open is called.
+    if (!doc[DOCUMENT_URL_RESOLVER])
+        doc[DOCUMENT_URL_RESOLVER] = createResolver(doc);
+
+    return doc[DOCUMENT_URL_RESOLVER];
```
inikulin: I forgot, why we need to always get it from the document? Why we can't store it in the var?

inikulin:
\r-

src/client/utils/url-resolver.js
miherlosev: After `document.open` is called the `nativeMethods` will refresh.
If we will try to call `urlResolve.updateBase` after `document.open` 
we will use native method from actual document.
But document that contains `a` element for url resolving created using previous `nativeMethods` version.

inikulin: We need a comment that clarifies this behavior in details then. I bet I've already asked about this piece of code and already forgot why we have it.

LavrovArtem:
this pull request fix #20

miherlosev:
wating for #384

miherlosev:
FPR

src/client/sandbox/index.js
```diff
@@ -84,6 +86,10 @@ export default class Sandbox extends SandboxBase {
             this.nativeMethods.refreshWindowMeths(window);
     }
 
+    _initUrlResolver (document) {
```
churkin: static?

churkin: Actually, I think this method should be in the urlResolver

src/client/utils/url.js
```diff
@@ -86,7 +87,7 @@ export function changeDestUrlPart (proxyUrl, prop, value, resourceType) {
     var parsed = sharedUrlUtils.parseProxyUrl(proxyUrl);
 
     if (parsed) {
-        var resolver  = destLocation.getResolver(document);
+        var resolver  = urlResolver.getResolverElement(document);
```
churkin: I would have moved all associated with the resolver element code in the urlResolver

src/client/utils/url-resolver.js
```diff
@@ -0,0 +1,62 @@
+import nativeMethods from '../sandbox/native-methods';
+import { ensureTrailingSlash } from '../../utils/url';
+
+const DOCUMENT_URL_RESOLVER = 'hammerhead|document-url-resolver';
+
+var urlResolver = {
```
churkin: why not use classes?

miherlosev: see https://github.com/DevExpress/testcafe-hammerhead/pull/383#discussion-diff-51120556

src/client/sandbox/node/element.js
```diff
@@ -349,12 +350,23 @@ export default class ElementSandbox extends SandboxBase {
 
             if (domUtils.isFileInput(el))
                 this.addFileInputInfo(el);
+
+            else if (domUtils.isBaseElement(el)) {
+                var storedHrefAttrName  = domProcessor.getStoredAttrName('href');
+                var storedHrefAttrValue = el.getAttribute(storedHrefAttrName);
+
+                urlResolver.updateBase(storedHrefAttrValue, this.document);
```
churkin: What if I change `href` value through a `setAttribute` method?

churkin: Test needed

src/client/sandbox/node/element.js
```diff
@@ -349,12 +350,23 @@ export default class ElementSandbox extends SandboxBase {
...
+            }
+
         }
     }
 
     onElementRemoved (el) {
         if (domUtils.isBodyElement(el))
             this.shadowUI.onBodyElementMutation();
+
+        else if (domUtils.isBaseElement(el))
```
churkin: I think we should make `DomMutation` event

src/client/utils/url-resolver.js
churkin:
```javascript
class UrlResolver {
...
}
export default new UrlResolver();
```
?

churkin:
-\r

src/client/sandbox/node/element.js
miherlosev: already present - https://github.com/DevExpress/testcafe-hammerhead/pull/383/files#diff-d69a730a5366fcee8825a74b81d833f4R309

src/client/sandbox/index.js
miherlosev: Ok

src/client/utils/url.js
miherlosev: Ok

src/client/utils/url-resolver.js
inikulin: No. This was a bad practice. Since it's a singletone it can be just plain object. No need for class here.

inikulin: But it's done wrong here anyway. Just `export default {}`, no need for explicit `var` here. Use `this` to refer object fields

src/client/sandbox/node/element.js
miherlosev: create the separate issue - https://github.com/DevExpress/testcafe-hammerhead/issues/385

miherlosev:
FPR

inikulin:
lgtm

miherlosev:
I disagree with @churkin .
We should not move this code to `_overrideSetAttributeCore`.
This method perform only 2 actions: set origin attribute with proxing value and set stored attribute with origin value.
In this sundbox all actions with external submodules are performed like I wrote: 
```javascript
 overridenMethod () {
    sandbox.before();
    var result = nativeMethodCall();
    sanbox.after();    
    return result;
 }
```
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/element.js#L247
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/element.js#L252
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/element.js#L291

For `_overridedRemoveAttributeCore` the same.

churkin:
The `_overridedSetAttributeCore` already contains all attribute checks. What sense make one of them to a higher level?

miherlosev:
So, after discussing with @churkin we decide to move `urlResolver.updateBase` call inside _overridedSetAttributeCore.
Also I added the 5 point in the issue https://github.com/DevExpress/testcafe-hammerhead/issues/385

miherlosev:
FPR

churkin:
lgtm

PR #391, opened 2016-02-02
JavaScript errors when task.js contains special sequences (close #389)

### LINKED ISSUE #389: JavaScript errors when task.js contains special sequences
There are sequences starting with `$` sign, that  have a special meaning and can be used in second argument of [String.prototype.replace](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter), and will transform output of replace. So according to code in [/src/client/sandbox/iframe.js](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/iframe.js#L125) if ```taskScriptTemplate``` has a such sequence, it will be corrupted.

Tip: A function can be used as a second argument to avoid this behavior:
```js
.replace('{{{iframeTaskScriptTemplate}}}', () => JSON.stringify(taskScriptTemplate));
```

churkin:
\cc @AndreyBelym @AlexanderMoskovkin 

Hard to write a good test for it.

### DIFF
--- a/src/client/sandbox/iframe.js
+++ b/src/client/sandbox/iframe.js
@@ -120,9 +120,9 @@ export default class IframeSandbox extends SandboxBase {
         // If it is needed elsewhere in a certain place, we should consider using Mustache.
         var taskScriptTemplate = settings.get().iframeTaskScriptTemplate;
         var taskScript         = taskScriptTemplate
-            .replace('{{{cookie}}}', JSON.stringify(this.cookieSandbox.getCookie()))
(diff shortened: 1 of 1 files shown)


AlexanderMoskovkin:
lgtm

AndreyBelym:
lgtm. Also publish it to npm, please.

inikulin:
That's missing regression tests.

Issue #277, opened 2015-11-19
navigator.serviceWorker may throw an exception due to the sandbox settings of the iframe


LavrovArtem:
```javascript
test('', function() {
    var iframe = document.createElement('iframe');

    iframe.setAttribute('sandbox', '');
    iframe.src = window.getCrossDomainPageUrl('../../../data/cross-domain/service-worker-not-available.html');

    document.body.appendChild(iframe);
});
```
```Uncaught SecurityError: Failed to read the 'serviceWorker' property from 'Navigator': Service worker is disabled because the context is sandboxed and lacks the 'allow-same-origin' flag.```
[Source code](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/native-methods.js#L96) where the error occurs.

miherlosev:
Behaviour changed in FF 44.
I am working on this.

miherlosev:
fixed in PR https://github.com/DevExpress/testcafe-hammerhead/pull/406

PR #403, opened 2016-02-05
fix 'Node.replaceChild doesn't handle file input elements' (close #399)

### LINKED ISSUE #399: Node.replaceChild doesn't handle file input elements
You can request upload for a file input, and the replace  it with another input with help of [Node.replaceChild](https://developer.mozilla.org/ru/docs/Web/API/Node.replaceChild) before submitting. In that case a value of the second input must be submitted, but hammerhead submits file selected in the first input. So we need to override replaceChild for file inputs just the way we are overriding Node.appendChild and Node.removeChild.

miherlosev:
/cc @churkin @AndreyBelym

### DIFF
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -237,12 +237,7 @@ export default class ElementSandbox extends SandboxBase {
             },
 
             removeChild (child) {
-                if (domUtils.isDomElement(child)) {
-                    domUtils.find(child, 'input[type=file]', ElementSandbox._removeFileInputInfo);
-
-                    if (domUtils.isFileInput(child))
-                        ElementSandbox._removeFileInputInfo(child);
-                }
+                sandbox._onRemoveFileInputInfo(child);
 
                 var result = nativeMethods.removeChild.call(this, child);
 
@@ -255,7 +250,13 @@ export default class ElementSandbox extends SandboxBase {
                 if (newChild.nodeType === 3)
                     ElementSandbox._processTextNodeContent(newChild, this);
 
-                return nativeMethods.replaceChild.call(this, newChild, oldChild);
+                sandbox._onRemoveFileInputInfo(oldChild);
+
+                var result = nativeMethods.replaceChild.call(this, newChild, oldChild);
+
+                sandbox._onAddFileInputInfo(newChild);
+
+                return result;
             },
 
             cloneNode () {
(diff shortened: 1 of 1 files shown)


churkin:
test?

miherlosev:
I don't know how to create a test because we cannot setup value for `input` with `type=file` on client side.

miherlosev:
@testcafe-build-bot \retest

miherlosev:
Waiting for #277

miherlosev:
FPR

miherlosev:
ping @churkin @AndreyBelym

miherlosev:
According the browser security restrictions, we cannot setup value for `input` with `type=file` on client side.
So, to create the client test we need to mock the following methods:
`domUtils.isDomElement`
`domUtils.isFileInput`
`UploadSandbox._getCurrentInfoManager`
`UploadSandbox.getFiles`
`getSandboxBackup`

It have made test very large and difficult.
What do you think?
@churkin @AlexanderMoskovkin

churkin:
@AlexanderMoskovkin Can be replaced by a testcafe functional test, i think.

miherlosev:
We decide to create separate tests for it. see https://github.com/DevExpress/testcafe/issues/293#issuecomment-181312367
So, FPR

AndreyBelym:
lgtm. I think there is a problem with a test because the PR adds functionality in `node` sandbox, but it's also heavily depends on `upload` sandbox. It means that we need to test both sandboxes in the same time. So can we create a test for upload sandbox, like test [here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/test/client/fixtures/sandbox/upload-test.js), but also connect the `node` sandbox?

miherlosev:
>> I think there is a problem with a test because the PR adds functionality in node sandbox, but it's also heavily depends on upload sandbox. It means that we need to test both sandboxes in the same time.

I don't see problem in this place. All `upload` subsistems are tested in [here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/test/client/fixtures/sandbox/upload-test.js).
We cannot write the integration qunit test (see [here] (https://github.com/DevExpress/testcafe-hammerhead/pull/403#issuecomment-181299735)).

This is why we will create intergration functional tests

miherlosev:
ping @churkin

churkin:
lgtm

PR #406, opened 2016-02-05
fix tests for FF 44

miherlosev:
/cc @churkin @LavrovArtem

PR #407, opened 2016-02-05
fix according https://github.com/DevExpress/testcafe-hammerhead/pull/406#diff-for-comment-52006861

miherlosev:
/cc @inikulin


inikulin:
Lgtm

PR #419, opened 2016-02-10
ie11 memory leaks

churkin:
\cc @inikulin @MargaritaLoseva

### DIFF
--- a/src/client/sandbox/xhr/index.js
+++ b/src/client/sandbox/xhr/index.js
@@ -117,8 +117,6 @@ export default class XhrSandbox extends SandboxBase {
             if (!this.corsSupported)
                 xhr.withCredentials = false;
 
-            XMLHttpRequestWrapper.prototype = xhr;
-
             return new XMLHttpRequestWrapper(xhr);
         };
     }


src/client/sandbox/xhr/index.js
```diff
@@ -117,9 +117,13 @@ export default class XhrSandbox extends SandboxBase {
...
-            XMLHttpRequestWrapper.prototype = xhr;
+            XhrSandbox.XMLHttpRequestWrapper.prototype = xhr;
 
-            return new XMLHttpRequestWrapper(xhr);
+            return new XhrSandbox.XMLHttpRequestWrapper(xhr);
         };
     }
 }
+
+// HACK: Do not even try to understand why it's done. Believe me,
```
MargaritaLoseva: Changed: Do not even try to understand why _this has been done_.

churkin:
@inikulin I have found a simple fix. We do not need to rewrite wrapper prototype as we add all the methods and properties manually. And it solves the problem.

inikulin:
That's cool. But we did it actually?

churkin:
I did not understand. It looks like a very old code.

churkin:
https://gist.github.com/churkin/e3a4fdd29010cb532572

churkin:
FPR

inikulin:
I guess it was added to keep properties that are not covered because of this filter: https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/xhr/xml-http-request-wrapper.js#L68 

I'd like to investigate this one a little better. At least try to find this proto assignment was introduced.

MargaritaLoseva:
lgtm

churkin:
@inikulin in this case 
```javascript
if (!Object.prototype.hasOwnProperty(prop)) {
```
check is unnecessary. It does not apply to prototype, always will return true to the XMLHttpRequest.

inikulin:
Ok, let's try it as is and rollback in case of issues in HHM

Issue #424, opened 2016-02-16
Perfomance optimization


miherlosev:
Lately health monitor is indicate perfomance problems.
We need to:
1)Profile client script and find bottlenecks
2)For some functions is need to get rid of iterative algorithm.
For example  `isElementInDocument` from /src/client/utils/dom.js.

miherlosev:
I'm working on this

miherlosev:
Some perfomance optimizations were done. 
Next perfomance optimizations were accomplished as separate issues.

PR #427, opened 2016-02-17
perfomance optimization for domUtils.isElementInDocument (part of #424)

miherlosev:
see http://jsperf.com/document-contains-versus-iterative-algorith/2

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -334,17 +334,9 @@ export function isDomElement (el) {
 }
 
 export function isElementInDocument (el, currentDocument) {
-    var doc        = currentDocument || document;
-    var curElement = el.parentNode;
+    var doc = currentDocument || document;
 
-    while (curElement) {
-        if (curElement === doc)
-            return true;
-
-        curElement = curElement.parentNode;
-    }
-
-    return false;
+    return doc.documentElement.contains(el);
 }
 
 export function isElementInIframe (el, currentDocument) {


src/client/utils/dom.js
```diff
@@ -334,17 +347,12 @@ export function isDomElement (el) {
...
 
 export function isElementInDocument (el, currentDocument) {
-    var doc        = currentDocument || document;
-    var curElement = el.parentNode;
+    var doc = currentDocument || document;
 
-    while (curElement) {
-        if (curElement === doc)
-            return true;
+    if (doc.contains)
```
AlexanderMoskovkin: Not all of browsers support `document.contains` (for example, IE11), but as I've found most of browsers support `document.body.contains`. Maybe we can use it here?

AlexanderMoskovkin: or `document.documentElement.contains`

miherlosev: We cannot use `document.body.contains`.
Inside functions `appendChild`, `insertBefore` we can use this function for `head` element.
see https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/element.js#L201

miherlosev: Ok. I will research for using `document.documentElement`.

miherlosev:
There is a potential problem with `document.documentElement.contains`.
In some browsers (IE, FF) `document.documentElement` for iframe don't exists right after iframe adding.
In hammerhead we don't use `isElementInDocument` for `document` without `documentElement`.
If we don't use this scenario in `testcafe` we can just use `document.documentElement.contains`.

AlexanderMoskovkin:
As i see there is no problem with this.

AlexanderMoskovkin:
lgtm

LavrovArtem:
lgtm

PR #429, opened 2016-02-17
remove isDomElement calling from common DOM functions (isImgElement, isInputElement and so on)

miherlosev:
part of https://github.com/DevExpress/testcafe-hammerhead/issues/424

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -90,7 +90,7 @@ export function getActiveElement (currentDocument) {
     var activeElement = doc.activeElement &&
                         doc.activeElement.tagName ? doc.activeElement : doc.body;
 
-    if (activeElement.tagName.toLowerCase() === 'iframe') {
+    if (isIframeElement(activeElement)) {
         try {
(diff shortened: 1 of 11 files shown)


src/client/utils/dom.js
```diff
@@ -346,8 +350,8 @@ export function isHammerheadAttr (attr) {
            attr.indexOf(INTERNAL_ATTRS.storedAttrPostfix) !== -1;
 }
 
-export function isIframe (el) {
```
AlexanderMoskovkin: This is a breaking change. But `testcafe` doesn't use changed functions. Will we increase the `hammerhead` major version because of this?
/cc @inikulin

miherlosev: ping @inikulin

inikulin: > This is a breaking change. But testcafe doesn't use changed functions

We can go with the minor then

AlexanderMoskovkin:
lgtm

LavrovArtem:
lgtm

PR #443, opened 2016-02-24
'isInputWithoutSelectionPropertiesInFirefox' method removed

helen-dikareva:
/cc @churkin 
we no longer need 'isInputWithoutSelectionPropertiesInFirefox' method, because bug with 'selectionStart/selectionEnd/selectionDirection' for input type = 'number' was fixed in Firefox


churkin:
lgtm

PR #457, opened 2016-03-04
Add service message disconnection handler

inikulin:
\cc @AlexanderMoskovkin


AlexanderMoskovkin:
It seems it not necessary now

PR #460, opened 2016-03-09
Replace apply by call 

miherlosev:
1)Replace apply by call  because `call` faster then `apply` (http://jsperf.com/call-apply-optimization)
2)fix Google Chrome profiler hint: 'Not optimized: Bad value context for arguments value'.
profiler_hint


churkin:
@miherlosev we can use arrows now

miherlosev:
We cannot declare function without braces because it has `return` statement.

churkin:
lgtm

AlexanderMoskovkin:
lgtm

PR #466, opened 2016-03-10
rewrite urlUtils.isSupportedProtocol(close #463)

miherlosev:
/cc @churkin @LavrovArtem 

Also I return some tests that we lost during move code to the github.

PR #475, opened 2016-03-14
fix wrong processing unsupported urls on server side (fix #474)

### LINKED ISSUE #474: A lot of 404 errors in HHM
Seems like the cause is #466

miherlosev:
/cc @churkin @inikulin

### DIFF
--- a/src/processing/resources/index.js
+++ b/src/processing/resources/index.js
@@ -9,6 +9,9 @@ import { ensureTrailingSlash } from '../../utils/url';
 
 function getResourceUrlReplacer (ctx) {
     return function (resourceUrl, resourceType, charsetAttrValue, baseUrl) {
+        if (!urlUtil.isSupportedProtocol(resourceUrl))
+            return resourceUrl;
+
         // NOTE: Resolves base URLs without a protocol ('//google.com/path' for example).
         baseUrl     = baseUrl ? url.resolve(ctx.dest.url, baseUrl) : '';
         resourceUrl = urlUtil.prepareUrl(resourceUrl);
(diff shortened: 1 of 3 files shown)


miherlosev:
@testcafe-build-bot \retest

miherlosev:
FPR

churkin:
gltm

src/processing/resources/index.js
```diff
@@ -9,6 +9,9 @@ import { ensureTrailingSlash } from '../../utils/url';
 
 function getResourceUrlReplacer (ctx) {
     return function (resourceUrl, resourceType, charsetAttrValue, baseUrl) {
+        if (!urlUtil.isSupportedProtocol(resourceUrl))
```
inikulin: So what was the cause of the problems and how what fixes it? Also missing regression tests

miherlosev: ### Algorithm
If we support the protocol then we should proxy url. If not then we shoul kept as is.
see client implementation - https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/url.js#L10

On server side exception was raised if protocol is not supported.
So in https://github.com/DevExpress/testcafe-hammerhead/pull/466 we remove error generation for non-supported protocol.
After then we became on server side calculate proxy url for all protocols.

### Regression test
I extend the existing test `Should process stylesheets` (see https://github.com/DevExpress/testcafe-hammerhead/blob/master/test/server/proxy-test.js#L491).
It use two files that was changed:
test/server/data/stylesheet/expected.css
test/server/data/stylesheet/src.css

inikulin: But why we don't throw like we did before if protocol is not supported?

inikulin:
lgtm

PR #495, opened 2016-03-23
Overriding FontFace only for the necessary browsers

churkin:
\cc @miherlosev


miherlosev:
lgtm.
waiting for green tests

PR #497, opened 2016-03-24
Update to v8.0.0, add 'createNativeXHR' API method

churkin:
After the https://github.com/DevExpress/testcafe-hammerhead/commit/ab60a4f59057887b2363e3d154bd4c855b3af74e commit testcafe can not use the `nativeMethods.XMLHttpRequest` anymore, because we are now override a prototype methods, and it is not the really native. For this purpose, we add a new `createNativeXHR` API method.

\cc @AlexanderMoskovkin


AlexanderMoskovkin:
lgtm

PR #503, opened 2016-03-28
Override svg elements with xlink:href attribute (close  #434)

### LINKED ISSUE #434: SVG: xlink:href attribute value is not proxed
see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href.
If xlink attribute value contains external url we should to proxy its value.

Example:
```javascript
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>
    <h1>Main document</h1>
    <svg xmlns="http://www.w3.org/2000/svg" style="display: none">
        <rect id="svg-rect" x="10" y="10" height="100" width="100" style="stroke:#ff0000; fill: #0000ff"/>
    </svg>
    <svg>
        <use style="width: 100px; height: 100px;" xlink:href="http://localhost:3000/#svg-rect"></use>
    </svg>
</body>
</html>
```

On example page there is the blue rectangle.
If you open this page using playouground then blue rectangle don't display.

miherlosev:


miherlosev:
FPR @inikulin @mo

AlexanderMoskovkin:
lgtm

src/client/utils/dom.js
```diff
@@ -520,8 +520,12 @@ export function isLocation (instance) {
     return instance && isObject(instance) && !isUndefined(instance.href) && !isUndefined(instance.assign);
 }
 
-export function isSVGElement (obj) {
-    return window.SVGElement && obj instanceof window.SVGElement;
+export function isSVGElement (instance) {
+    if (instance instanceof nativeMethods.svgElementClass)
+        return true;
+
+    return instance && isObject(instance) && !isUndefined(instance.ownerSVGElement) && instance.toString &&
```
inikulin: Why we need this `isUndefined` check and this method (and all other methods in `type` utils)?  Why just don't test `instance.ownerSVGElement !== void 0` in place?

src/utils/parse5.js
```diff
@@ -49,3 +73,42 @@ export function createTextNode (content, parent) {
...
+    for (var i = 0; i < el.attrs.length; i++) {
+        if (getAttrName(el.attrs[i]) === attr) {
+            el.attrs.splice(i, 1);
+
+            return;
+        }
+    }
+}
+
+export function getAttr (el, attr) {
```
inikulin: `attr` arg should be renamed to `name`. And we have code duplication that searches by attr name. Maybe we will have internal `findAttr` method that will return `attr` object and use it `getAttr` and `setAttr` methods?

src/client/utils/dom.js
miherlosev: We have any shared type methods - https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/utils/types.js.
Is there something wrong?

inikulin:
\r-

src/client/utils/dom.js
inikulin: Yep, why we need them?

src/utils/parse5.js
miherlosev: Ok. I will do.

src/client/utils/dom.js
inikulin: `inaccessibleTypeToStr` - ok
`isNullOrUndefine` - kinda ok (it has unnecessary `!obj` check)

all other - not ok

miherlosev: Separate issue - https://github.com/DevExpress/testcafe-hammerhead/issues/507

miherlosev:
FPR

miherlosev:
ping @inikulin

src/processing/dom/index.js
```diff
@@ -57,55 +65,61 @@ export default class DomProcessor {
...
 
-            IS_SANDBOXED_IFRAME: el => adapter.getTagName(el).toLowerCase() === 'iframe' &&
-                                       adapter.hasAttr(el, 'sandbox')
+            IS_SANDBOXED_IFRAME: el => adapter.getTagName(el) === 'iframe' &&
+                                       adapter.hasAttr(el, 'sandbox'),
+
+            IS_SVG_ELEMENT_WITH_XLINK_HREF_ATTR: el => {
+                return adapter.hasAttr(el, 'xlink:href') &&
+                       SVG_XLINK_HREF_TAGS.indexOf(adapter.getTagName(el)) !== -1 &&
+                       adapter.isSVGElement(el);
```
inikulin: I'd rather move `adapter.isSVGElement(el)` condition to the first place, since it's the most assertive and low-cost check.

inikulin:
lgtm

miherlosev:
ping @AlexanderMoskovkin

AlexanderMoskovkin:
lgtm

PR #511, opened 2016-03-31
implement about:blank and about:error service pages (close #339)

### LINKED ISSUE #339: Implement about:blank page
We need to implement page for the `http://proxy-host/session-id/about:blank`. It should be just blank page with our stuff injected. We will need for the new API, there tests could start on blank page. BTW, we can fallback to this blank page URL then we can't create proxy URL instead of throwing an error in `getProxyUrl`

miherlosev:
@inikulin @AlexanderMoskovkin

### DIFF
--- a/src/proxy/index.js
+++ b/src/proxy/index.js
@@ -2,7 +2,7 @@ import Router from './router';
 import http from 'http';
 import * as urlUtils from '../utils/url';
 import { readSync as read } from 'read-file-relative';
-import { respond500, respondWithJSON, fetchBody } from '../utils/http';
+import { respond500, respondWithJSON, fetchBody, preventCaching } from '../utils/http';
 import { ie9FileReaderShim } from '../upload';
 import { run as runRequestPipeline } from '../request-pipeline';
 import prepareShadowUIStylesheet from '../shadow-ui/create-shadow-stylesheet';
@@ -105,8 +105,7 @@ export default class Proxy extends Router {
 
         if (session) {
             res.setHeader('content-type', 'application/x-javascript');
-            res.setHeader('cache-control', 'no-cache, no-store, must-revalidate');
-            res.setHeader('pragma', 'no-cache');
+            preventCaching(res);
             res.end(session.getTaskScript(referer, refererDest.destUrl, serverInfo, isIframe, true));
         }
         else
--- a/src/request-pipeline/context.js
+++ b/src/request-pipeline/context.js
@@ -44,10 +44,11 @@ export default class RequestPipelineContext {
         this.destResBody   = null;
         this.hasDestReqErr = false;
 
-        this.isXhr       = false;
-        this.isPage      = false;
-        this.isIframe    = false;
-        this.contentInfo = null;
+        this.isXhr         = false;
+        this.isPage        = false;
+        this.isIframe      = false;
+        this.isSpecialPage = false;
+        this.contentInfo   = null;
 
         var acceptHeader = req.headers['accept'];
 
@@ -56,8 +57,8 @@ export default class RequestPipelineContext {
     }
 
     _getDestFromReferer (parsedReferer) {
-        // NOTE: Browsers may send the default port in the ‘referer’ header. But since we compose the destination
-        // URL from it, we need to skip the port number if it's the protocol’s default port. Some servers have
+        // NOTE: Browsers may send the default port in the 'referer' header. But since we compose the destination
+        // URL from it, we need to skip the port number if it's the protocol's default port. Some servers have
         // host conditions that do not include a port number.
         var rDest         = parsedReferer.dest;
         var isDefaultPort = rDest.protocol === 'https:' && rDest.port === '443' ||
@@ -109,7 +110,7 @@ export default class RequestPipelineContext {
         parsedReqUrl  = flattenParsedProxyUrl(parsedReqUrl);
         parsedReferer = flattenParsedProxyUrl(parsedReferer);
 
-        // NOTE: Try to extract the destination from the ‘referer’ header.
+        // NOTE: Try to extract the destination from the 'referer' header.
         if (!parsedReqUrl && parsedReferer)
             parsedReqUrl = this._getDestFromReferer(parsedReferer);
 
--- a/src/request-pipeline/index.js
+++ b/src/request-pipeline/index.js
@@ -7,6 +7,7 @@ import connectionResetGuard from './connection-reset-guard';
 import { check as checkSameOriginPolicy } from './xhr/same-origin-policy';
 import { fetchBody, respond404 } from '../utils/http';
 import { inject as injectUpload } from '../upload';
+import { isSpecialPage } from '../utils/url';
 
 // Stages
 var stages = {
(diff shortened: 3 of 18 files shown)


src/proxy/index.js
```diff
@@ -77,6 +79,8 @@ export default class Proxy extends Router {
         this.POST('/messaging', (req, res, serverInfo) => this._onServiceMessage(req, res, serverInfo));
         this.GET('/task.js', (req, res, serverInfo) => this._onTaskScriptRequest(req, res, serverInfo, false));
         this.GET('/iframe-task.js', (req, res, serverInfo) => this._onTaskScriptRequest(req, res, serverInfo, true));
+        this.GET('/about:blank', (req, res) => this._onServicePageRequest(req, res, SERVICE_PAGE_TYPE.blank));
+        this.GET('/error', (req, res) => this._onServicePageRequest(req, res, SERVICE_PAGE_TYPE.error));
```
inikulin: It should be `about:error`

src/proxy/index.js
```diff
@@ -77,6 +79,8 @@ export default class Proxy extends Router {
         this.POST('/messaging', (req, res, serverInfo) => this._onServiceMessage(req, res, serverInfo));
         this.GET('/task.js', (req, res, serverInfo) => this._onTaskScriptRequest(req, res, serverInfo, false));
         this.GET('/iframe-task.js', (req, res, serverInfo) => this._onTaskScriptRequest(req, res, serverInfo, true));
+        this.GET('/about:blank', (req, res) => this._onServicePageRequest(req, res, SERVICE_PAGE_TYPE.blank));
```
inikulin: This is not correct. This pages should be dispatched via proxy. So I will be able to get proxy URL fro them. But shouldn't perform any requests to server, instead it should return empty page (with no content), but we should be able to extract session id, etc.

inikulin:
\r-

src/request-pipeline/context.js
```diff
@@ -101,7 +102,7 @@ export default class RequestPipelineContext {
 
     // API
     dispatch (openSessions) {
-        var parsedReqUrl  = urlUtils.parseProxyUrl(this.req.url);
+        var parsedReqUrl  = urlUtils.parseProxyUrl(this.req.url, true);
```
inikulin: Why we need additional parameter?

miherlosev: see https://github.com/DevExpress/testcafe-hammerhead/pull/511/files#diff-d611cf61b8eaf92ca0d07018fdd1bec6

src/request-pipeline/index.js
```diff
@@ -136,9 +136,17 @@ function isDestResBodyMalformed (ctx) {
...
-    if (ctx.dispatch(openSessions)) {
+    var ctx         = new RequestPipelineContext(req, res, serverInfo);
+    var isDispatch  = ctx.dispatch(openSessions);
+    var servicePage = ctx.getServicePage();
+
+    if (isDispatch && servicePage) {
+        res.setHeader('content-type', 'text/html');
+        res.statusCode = 200;
+        res.write('<html><body><h1>' + servicePage.name + '</h1></body></html>');
+        res.end();
```
inikulin: Wrong. We should check parsed url for `about:blank' or 'about:error'. Set content to empty string and directly jump to the step `5` of pipeline.

src/request-pipeline/index.js
```diff
@@ -136,9 +136,17 @@ function isDestResBodyMalformed (ctx) {
...
-
-    if (ctx.dispatch(openSessions)) {
+    var ctx         = new RequestPipelineContext(req, res, serverInfo);
+    var isDispatch  = ctx.dispatch(openSessions);
+    var servicePage = ctx.getServicePage();
+
+    if (isDispatch && servicePage) {
+        res.setHeader('content-type', 'text/html');
+        res.statusCode = 200;
+        res.write('<html><body><h1>' + servicePage.name + '</h1></body></html>');
```
inikulin: We don't need any body

src/request-pipeline/context.js
```diff
@@ -215,4 +216,17 @@ export default class RequestPipelineContext {
 
         return urlUtils.getProxyUrl(url, this.serverInfo.hostname, port, this.session.id, resourceType, charsetAttrValue);
     }
+
+    getServicePage () {
+        var parsedReqUrl = urlUtils.parseProxyUrl(this.req.url, true);
```
inikulin: We don't need this method

src/service-pages/type.js
```diff
@@ -0,0 +1,11 @@
+export default {
```
inikulin: We don't need this file

inikulin:
\r-

I should be able to do `getProxyUrl('about:blank')` and get proxied page, but without content. Obviously, it can't be dispatched to any destination server, therefore we should return empty page with our stuff injected.

miherlosev:
FPR

src/request-pipeline/context.js
inikulin: Still don't get it

inikulin:
What about getProxyUrl and isSupportedProtocol check?

src/request-pipeline/special-pages.js
```diff
@@ -0,0 +1,9 @@
+// -------------------------------------------------------------
+// WARNING: this file is used by both the client and the server.
+// Do not use any browser or node-specific API!
+// -------------------------------------------------------------
+
+export default {
```
inikulin: They are always used in pair, so it can be just an array. Also this const can be moved to url utils I guess

src/request-pipeline/index.js
```diff
@@ -17,6 +18,17 @@ var stages = {
 
     1: function sendDestinationRequest (ctx, next) {
         var opts = createReqOpts(ctx);
+
+        if (opts.url === SPECIAL_PAGES.blank)
+            ctx.isBlankPage = true;
```
inikulin: Why we need two different flags? Processing is identical

src/request-pipeline/index.js
```diff
@@ -54,7 +66,10 @@ var stages = {
     },
 
     4: async function fetchContent (ctx, next) {
-        ctx.destResBody = await fetchBody(ctx.destRes);
+        if (ctx.isBlankPage || ctx.isErrorPage)
```
inikulin: special page flag should be defined in ctx ctor

src/request-pipeline/index.js
```diff
@@ -133,6 +148,16 @@ function isDestResBodyMalformed (ctx) {
     return !ctx.destResBody || ctx.destResBody.length !== ctx.destRes.headers['content-length'];
 }
 
+function sendSpecialPageHeaders (ctx) {
+    ctx.destRes = {
```
inikulin: This method doesn't send anything, it just assigns property

inikulin:
getProxyUrl and parseProxyUrl should be tested as well. Client location also  requires testing

inikulin:
\r-

inikulin:
So the task is: make proxied `about:blank` page in playground behave like real `about:blank` in browser: has same location, produce same relative hyperlinks, etc. Also, it should be possible to navigate to `about:blank` page via hyperlink from other page

miherlosev:
Also need to do:
- [x] create proxy url on `link.setAttribute('about:blank') and don't create proxy url for another tags

miherlosev:
FPR

inikulin:
lgtm

churkin:
lgtm

PR #513, opened 2016-04-01
fix wrong override for wrong override for window.history.pushState and window.history.replaceState

miherlosev:


AlexanderMoskovkin:
lgtm

Issue #477, opened 2016-03-15
SVG: xml: base attribute value is not proxed


miherlosev:
Reproduced on http://livejournal.ru only in FF.

see video - http://screencast.com/t/H1GuAQs3q

LavrovArtem:
I'm working on this

PR #516, opened 2016-04-04
fix script error for SVGElement.setAttributeNS (close #514)

### LINKED ISSUE #514: Regression: script error inside SVGElement.setAttributeNS
see https://testcafe-hhhm.devexpress.com/history?timestamp=1459537468694&siteUrl=http://badoo.com/
Related to https://github.com/DevExpress/testcafe-hammerhead/pull/503

miherlosev:
@churkin @LavrovArtem

### DIFF
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -137,7 +137,7 @@ export default class ElementSandbox extends SandboxBase {
                  domUtils.isSVGElement(el)) {
             var storedXLinkHrefAttr = domProcessor.getStoredAttrName(attr);
 
-            setAttrMeth.call(el, storedXLinkHrefAttr, value);
+            setAttrMeth.apply(el, ns ? [ns, storedXLinkHrefAttr, value] : [storedXLinkHrefAttr, value]);
 
             if (!HASH_RE.test(value))
                 value = urlUtils.getProxyUrl(value);
--- a/test/client/fixtures/sandbox/node/dom-processor-test.js
+++ b/test/client/fixtures/sandbox/node/dom-processor-test.js
@@ -475,10 +475,12 @@ asyncTest('script error when a new element is added to a "body" element that is 
     document.body.appendChild(iframe);
 });
 
-test('xlink:href attribute of svg elements should be overriden (GH-434)', function () {
-    var use    = document.createElementNS('http://www.w3.org/2000/svg', 'use');
-    var svgUrl = 'http://domain.com/test.svg#rect';
-    var div    = document.createElement('div');
+test('xlink:href attribute of svg elements should be overriden (GH-434)(GH-514)', function () {
+    var svgNameSpaceUrl   = 'http://www.w3.org/2000/svg';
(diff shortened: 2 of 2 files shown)


src/processing/dom/html-namespaces.js
```diff
@@ -0,0 +1,13 @@
...
+// -------------------------------------------------------------
+
+export default {
+    HTML:   'http://www.w3.org/1999/xhtml',
+    MATHML: 'http://www.w3.org/1998/Math/MathML',
+    SVG:    'http://www.w3.org/2000/svg',
+    XLINK:  'http://www.w3.org/1999/xlink',
+    XML:    'http://www.w3.org/XML/1998/namespace',
+    XMLNS:  'http://www.w3.org/2000/xmlns/'
+};
```
churkin: I propose to use lowercase

miherlosev: It named as in https://www.w3.org/TR/2011/WD-html5-20110405/namespaces.html.
So, in parse5 it named simular (https://github.com/inikulin/parse5/blob/24cb16c4a46f84fba01c6f41fc96817a013a6495/lib/common/html.js).

churkin: We use lowercase format for a constants. @inikulin

miherlosev: ping @inikulin

inikulin: @churkin is right. For constant enumerations we use lowercase. It should be always used as 
`HTML_NAMESPACE.html` then. But, I have no idea why we need ALL namespace constants - only two of them used in code. I don't see why we need dedicated file for them. It's just fine to have inline constants in each file in that case.

miherlosev: > But, I have no idea why we need ALL namespace constants - only two of them used in code.

For future use.
For example, for https://github.com/DevExpress/testcafe-hammerhead/issues/477  issue we will need the XML namespace.

> I don't see why we need dedicated file for them. It's just fine to have inline constants in each file in that case.

The same logical costants (namespaces) grouped in the separate file.

inikulin: #477 is related to XML

> For future use.

Never add dead code based on assumption. Let's add namespace when it they will be required.

miherlosev: Ok

miherlosev: What about it?

> I don't see why we need dedicated file for them. It's just fine to have inline constants in each file in that case.

The same logical costants (namespace constants) grouped in the separate file.

inikulin: Until they are not used in different files I don't see necessity to move them to the dedicated file

miherlosev: Ok

miherlosev:
FPR

churkin:
lgtm

miherlosev:
ping @inikulin

inikulin:
lgtm

LavrovArtem:
lgtm

PR #536, opened 2016-04-13
Range.createContextualFragment is not overriden (close #535)

### LINKED ISSUE #535: Range.createContextualFragment is not overriden

churkin:
\cc @AlexanderMoskovkin @miherlosev

### DIFF
--- a/src/client/sandbox/node/window.js
+++ b/src/client/sandbox/node/window.js
@@ -6,6 +6,7 @@ import nativeMethods from '../native-methods';
 import { processScript } from '../../../processing/script';
 import styleProcessor from '../../../processing/style';
 import * as destLocation from '../../utils/destination-location';
+import { processHtml } from '../../utils/html';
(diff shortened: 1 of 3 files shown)


src/client/sandbox/node/window.js
```diff
@@ -183,6 +184,17 @@ export default class WindowSandbox extends SandboxBase {
             };
         }
 
+        window.Range.prototype.createContextualFragment = function () {
```
miherlosev: It is wrong. In 9 `window.Range.prototype` does not contains method `createContextualFragment`.
Need to change on:
```javascript
if (window.Range.prototype.createContextualFragment) {
    window.Range.prototype.createContextualFragment = function (){...}
}
```

miherlosev:
\r-

churkin:
FPR

AlexanderMoskovkin:
lgtm

PR #540, opened 2016-04-14
outerHTML property is not overridden (fix #538)

### LINKED ISSUE #538: outerHTML property is not overridden
Some ASPxEditors regression tests use it to perform markup checks.

LavrovArtem:
\cc @churkin @AndreyBelym

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -261,6 +261,40 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
                 }
             },
 
+            outerHTML: {
+                condition: el => domUtils.isElementNode(el) && typeof el.outerHTML === 'string',
+                get:       el => cleanUpHtml(el.outerHTML, el.parentNode && el.parentNode.tagName),
+
+                set: (el, value) => {
+                    var parentEl = el.parentNode;
+
+                    if (parentEl) {
+                        var parentDocument = domUtils.findDocument(parentEl);
+                        var parentWindow   = parentDocument ? parentDocument.defaultView : null;
+
+                        el.outerHTML = processHtml('' + value, parentEl.tagName);
+
+                        // NOTE: For the iframe with an empty src.
+                        if (parentWindow && parentWindow !== window &&
+                            parentWindow[INTERNAL_PROPS.processDomMethodName])
+                            parentWindow[INTERNAL_PROPS.processDomMethodName](parentEl, parentDocument);
+                        else if (window[INTERNAL_PROPS.processDomMethodName])
+                            window[INTERNAL_PROPS.processDomMethodName](parentEl);
+
+                        // NOTE: This check is required for an element in an unavailable window.
+                        // NOTE: Use timeout, so that changes take effect.
+                        if (window.self && domUtils.isBodyElement(el))
+                            nativeMethods.setTimeout.call(window, () => this.shadowUI.onBodyElementMutation(), 0);
+
+                        return value;
+                    }
+
+                    el.outerHTML = value;
+
+                    return value;
+                }
+            },
+
             onerror: {
                 condition: domUtils.isWindow,
                 get:       owner => owner[ORIGINAL_WINDOW_ON_ERROR_HANDLER_KEY] || null,
--- a/test/client/fixtures/sandbox/code-instrumentation/getters-test.js
+++ b/test/client/fixtures/sandbox/code-instrumentation/getters-test.js
@@ -1,9 +1,11 @@
 var INTERNAL_PROPS         = hammerhead.get('../processing/dom/internal-properties');
+var SHADOW_UI_CLASSNAME    = hammerhead.get('../shadow-ui/class-name');
 var urlUtils               = hammerhead.get('./utils/url');
 var processScript          = hammerhead.get('../processing/script').processScript;
 var removeProcessingHeader = hammerhead.get('../processing/script/header').remove;
 var destLocation           = hammerhead.get('./utils/destination-location');
 var attributesProperty     = hammerhead.get('../client/sandbox/code-instrumentation/properties/attributes');
+var processHtml            = hammerhead.get('../client/utils/html').processHtml;
 
 var browserUtils  = hammerhead.utils.browser;
 var nativeMethods = hammerhead.nativeMethods;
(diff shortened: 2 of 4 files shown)


LavrovArtem:
FPR

src/client/sandbox/code-instrumentation/properties/index.js
```diff
@@ -229,6 +229,40 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
...
+                        // NOTE: For the iframe with an empty src.
+                        if (parentWindow && parentWindow !== window &&
+                            parentWindow[INTERNAL_PROPS.processDomMethodName])
+                            parentWindow[INTERNAL_PROPS.processDomMethodName](parentEl, parentDocument);
+                        else if (window[INTERNAL_PROPS.processDomMethodName])
+                            window[INTERNAL_PROPS.processDomMethodName](parentEl);
+
+                        // NOTE: This check is required for an element in an unavailable window.
+                        // NOTE: Use timeout, so that changes take effect.
+                        if (window.self && domUtils.isBodyElement(el))
```
AndreyBelym: You can't use `el` after `el.outerHTML` was executed (in Chrome, at least), because:
```js
var body = document.body;

body.outerHTML;
//<body>BODY1</body>

body.outerHTML = '<body>BODY2</body>'
//"<body>BODY2</body>"

body.outerHTML
//"<body>BODY1</body>"

body.parentNode
//null

document.body
//<body>BODY2</body>
```
At least for `<body>`, rewiting `el.outerHTML` leads to behavior like `el.parentNode.replaceChild(newElFromOuterHTML, el)`, and `el` points to original element which HTML is not changed and that is removed from DOM.

AndreyBelym:
\r-

src/client/sandbox/code-instrumentation/properties/index.js
AndreyBelym: On second thought, I realized that purpose of this code is to emit an if 'body' has been changed, so it's probably fine.

AndreyBelym: Wow, that is a really twisted thing actually:
```js
document.head
//<head></head>

document.body
//<body></body>

document.head.outerHTML = '<head id="myhead"></head><body id="mybody"></body>';
//'<head id="myhead"></head><body id="mybody"></body>'

document.head
//<head id="myhead"></head>

document.body
//<body id="mybody"></body>

document.documentElement
//<html>
//    <head id="myhead"></head>
//    <body id="mybody"></body>
//    <body></body>
//</html>
```
Don't know, will be there a practical point of messing with that case?
I just need getter that cleans HTML to fix ASP tests.

LavrovArtem: Hmmm, @churkin what you think about this?

AndreyBelym: Also, if `outerHTML` is `<body>...</body>` or `<head>...</head>`, note that they need special handling in processHtmlInternal [here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/html.js#L160). `container.innerHTML = html` will not work if `container` is '<div>'.

LavrovArtem: https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/html.js#L134

```preparePageHtml``` wrap ```html```, ```body```, ```head``` and ```doctype``` tags

Comment (AndreyBelym) [inline src/client/sandbox/code-instrumentation/properties/index.js]: Oh, I see it now. 
BTW tests for that I need this fix are fine, but I want to check it on complete test suite, so give me some time.

File: src/client/sandbox/code-instrumentation/properties/index.js [hunk 2]
```diff
@@ -229,6 +229,40 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
                 }
             },
 
+            outerHTML: {
+                condition: domUtils.isElementNode,
+                get:       el => cleanUpHtml(el.outerHTML, el.parentNode && el.parentNode.tagName),
```
AndreyBelym: Need to check if `el.outerHTML` is not `undefined`, otherwise it will break `wrapTextNodes` inside `cleanUpHtml`.

LavrovArtem: Before call of `get` function, call `condition` which checks that `el` is node element

AndreyBelym: It doesn't work. In ASPxUpload control uses `XMLDocument` instance with content like 
```xml
<status progress="100" totalSize="4" totalUploadedSize="4" contentType="" fileProgress="100" fileUploadedSize="4" fileSize="4" fileName="test.txt"></status>
```

It checks for  `xml.documentElement.outerHTML` being `undefined` to detect, it is valid XML document or not. In case of invalid XML, `documentElement` doesn't have `outerHTML` property, but it's `nodeType` is `1`, so it passes our `domUtils.isElementNode` check.

LavrovArtem:
FPR

test/client/fixtures/sandbox/code-instrumentation/setters-test.js
```diff
@@ -304,6 +291,26 @@ if (!browserUtils.isIE) {
...
+    parentDiv.appendChild(childDiv);
+
+    strictEqual(parentDiv.children.length, 1);
+    strictEqual(parentDiv.firstChild, childDiv);
+
+    eval(processScript('childDiv.outerHTML = htmlText', true, false));
+
+    strictEqual(parentDiv.children.length, 2);
+    strictEqual(parentDiv.firstChild.href, urlUtils.getProxyUrl('http://domain.com/'));
+    strictEqual(parentDiv.lastChild.src, urlUtils.getProxyUrl('http://domain.com/script', null, null, null, 's'));
```
AndreyBelym: I'd add a test for IE where `outerHTML` is undefined:
```js
if (browserUtils.isIE) {
   var doc = document.implementation.createDocument(null, 'status', null);
   strictEqual(doc.documentElement.outerHTML, void 0);
}
```

AndreyBelym:
\r-, please, add a test where `outerHTML` is undefined.

test/client/fixtures/sandbox/code-instrumentation/getters-test.js
```diff
@@ -205,6 +207,30 @@ test('get style body', function () {
...
+    var div      = document.createElement('div');
+
+    eval(processScript('div.innerHTML = htmlText', true, false));
+
+    var a = div.firstChild;
+
+    strictEqual(a.outerHTML, processHtml(htmlText));
+    strictEqual(eval(processScript('a.outerHTML', true, false)), htmlText);
+});
+
```
AndreyBelym: I'd add a test for IE where `outerHTML` is undefined:
```js
if (browserUtils.isIE) {
   var doc = document.implementation.createDocument(null, 'status', null);
   strictEqual(eval(processScript('doc.documentElement.outerHTML', true, false)), void 0);
}
```

LavrovArtem:
FPR

AndreyBelym:
lgtm

churkin:
lgtm

PR #542, opened 2016-04-15
Hammerhead initialization fails in IE after document.write (close #525)

### LINKED ISSUE #525: Hammerhead initialization fails in IE after document.write
A `cannot execute code from freed script` exception is raised when evaluating task script in iframe, after `document.write` was called. Also bad thing is that the error is silently swallowed, because it is thrown on [RUN_TASK_SCRIPT event](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/iframe.js#L94), and event emitter just ignores `freed script` errors (https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/event-emitter.js#L23). 

When `document.write` is called, `hammerhead` reattaches existing sandbox, including ShadowUI. When ShadowUI attaches to DOM, it replace some methods in HTMLBodyElement and HeadHTMLElement prototypes. After that iframe is initialized through `_raiseReadyToInitEvent`, and evaluate task script. When it does that, it calls UrlResolver.updateBase, that uses `document.head.getElementsByTagName` method(https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/url-resolver.js#L61), previously overriden by ShadowUI at `document.write`. So it raises a `freed script` error here.

churkin:
\сс @AndreyBelym @LavrovArtem

### DIFF
--- a/src/client/sandbox/index.js
+++ b/src/client/sandbox/index.js
@@ -46,6 +46,19 @@ export default class Sandbox extends SandboxBase {
         this.node                = new NodeSandbox(nodeMutation, this.iframe, this.event, this.upload, this.shadowUI);
     }
 
+    // NOTE: In some cases, IE raises the "Can't execute code from a freed script" exception,
+    // so that we cannot use a sandbox created earlier and we have to create a new one.
+    static _canUseSandbox (sandbox) {
+        try {
+            sandbox.off();
+        }
+        catch (e) {
+            return false;
+        }
+
+        return true;
+    }
+
     _refreshNativeMethods (window, document) {
         var tryToExecuteCode = func => {
             try {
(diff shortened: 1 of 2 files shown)


LavrovArtem:
What about tests?

AndreyBelym:
It's hard to write a test for it. This bug reproduces very frequently on a lot of ASP tests on farm in automatic mode, but very rarely on my machine or on farm in manual mode. Likely it depends on internal IE's conditions by which it will make script freed or not.

AndreyBelym:
lgtm, tested on farm.

churkin:
@AndreyBelym excellent!

miherlosev:
lgtm

PR #543, opened 2016-04-18
PageProcessor parse binaries as a html page (close #521)

### LINKED ISSUE #521: PageProcessor parse binaries as a html page
Reproduces in the playground when try to download a file.

For example reproduced on the http://testcafe.devexpress.com page
Click `Download TestCafe` button.
We have the following `Content-Type` in response: `application/octet-stream`

But `RequestPipelineContext.isPage`  is `true` because it depends on request headers ([see here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/request-pipeline/context.js#L55))
As a result `PageProcessor.shouldProcessResource(ctx)` returns `true` and parse response via `parse5` module ([see here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/processing/resources/page.js#L129))

churkin:
\cc @AlexanderMoskovkin @miherlosev

### DIFF
--- a/src/processing/resources/page.js
+++ b/src/processing/resources/page.js
@@ -126,7 +126,8 @@ class PageProcessor extends ResourceProcessorBase {
     shouldProcessResource (ctx) {
         // NOTE: In some cases, Firefox sends the default accept header for the script.
         // We should not try to process it as a page in this case.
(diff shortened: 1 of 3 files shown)


AlexanderMoskovkin:
lgtm

miherlosev:
lgtm

PR #549, opened 2016-04-20
change comment (close #489)

### LINKED ISSUE #489: window.XMLHttpRequest.prototype.send refactoring
Overriden method has the code fragment:
```javascript
            var orscHandler = () => {
                if (xhr.readyState === 4)
                    xhrSandbox.emit(xhrSandbox.XHR_COMPLETED_EVENT, { xhr });
            };

            // NOTE: If we're using the sync mode or the response is in cache and the object has been retrieved
            // directly (IE6 & IE7), we need to raise the callback manually.
            if (xhr.readyState === 4)
                orscHandler();
            else {
            ...
```
IE6, IE7 is not supported browsers.
Investigate this case. If it is need then update the comment.

miherlosev:
I removed from comment unclear part.
Updated comment was checked by @MargaritaLoseva .
@churkin

### DIFF
--- a/src/client/sandbox/xhr.js
+++ b/src/client/sandbox/xhr.js
@@ -71,8 +71,8 @@ export default class XhrSandbox extends SandboxBase {
                     xhrSandbox.emit(xhrSandbox.XHR_COMPLETED_EVENT, { xhr });
             };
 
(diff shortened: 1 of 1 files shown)


churkin:
lgtm

Issue #384, opened 2016-01-29
'Can't execute code from a freed script' error for document.write in iframe with same-domain src


miherlosev:
Test for reproduce:
```javascript
   var iframe = document.createElement('iframe');
    var src    = window.QUnitGlobals.getResourceUrl('index.html');

    iframe.id = 'test_ojfnnhsg43';
    iframe.setAttribute('src', src);
    window.QUnitGlobals.waitForIframe(iframe)
        .then(function () {
            var iframeDocument = iframe.contentDocument;
            
            iframeDocument.write('<h1>test</h1>'); //this line throws the error
            
            iframe.parentNode.removeChild(iframe);
            start();
        });
    document.body.appendChild(iframe);
```

```javascript
index.html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title></title>
    <script src="/hammerhead.js" class="script-hammerhead-shadow-ui"></script>
</head>
<body>
    <script type="text/javascript">
        var hammerhead = window['%hammerhead%'];

        hammerhead.get('./utils/destination-location').forceLocation('http://localhost/sessionId/http://subdomain.example.com');
        hammerhead.start({
            sessionId: 'sessionId',
            CROSS_DOMAIN_PROXY_PORT: 2000
        });
    </script>
    <a href="/index.html">/index.html</a>
</body>
</html>
```

Only in IE and Edge.

miherlosev:
I'm working on this

churkin:
Maybe it's a duplicate of https://github.com/DevExpress/testcafe-hammerhead/issues/525?

miherlosev:
Probably. I will check it after fix #525

miherlosev:
Origin bug is dublicate of #525.
But in IE (I check in IE11) 'Permission denied' script error occurs.
It is raised inside DocumentWriter.
@LavrovArtem  please, investiagate it.

LavrovArtem:
I'm working on this

LavrovArtem:
In IE after [write](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/document/index.js#L48) we can not call any function. If we remove [this code](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/document/index.js#L51)
```javascript
            this.nodeSandbox.mutation.onDocumentCleaned({
                window:             this.window,
                document:           this.document,
[...]

PR #556, opened 2016-04-25
[WIP] check all places where we perform writing/reading to window or document (close #533)

### LINKED ISSUE #533: Check all places where we perform writing/reading to window or document
Element can be created in top level window and then added to an iframe.

``` javascript
var div = document.createElement('div');

iframe.addEventListener('load', function () {
     iframe.contentDocument.body.appendChild(div);
}):

```

If we directly read / write to `window` or `document` in our code and it used for elements - it is a bug.
(for example see https://github.com/DevExpress/testcafe-hammerhead/issues/529).
So we need to check all our code and fix wrong behaviour.

miherlosev:


miherlosev:
Waiting for fix #384

miherlosev:
@testcafe-build-bot \retest

miherlosev:
It is not actual now.

PR #557, opened 2016-04-25
node.js v5.11 support

churkin:
\cc @miherlosev @AlexanderMoskovkin @VasilyStrelyaev

PR #560, opened 2016-04-27
fix health monitor regression after #557

churkin:
\cc @miherlosev @kirovboris

### DIFF
--- a/src/client/sandbox/xhr.js
+++ b/src/client/sandbox/xhr.js
@@ -4,6 +4,8 @@ import { getProxyUrl } from '../utils/url';
 import XHR_HEADERS from '../../request-pipeline/xhr/headers';
 import { getOrigin } from '../utils/destination-location';
 
+const IS_OPENED_XHR = 'hammerhead|xhr|is-opened-xhr';
+
 // NOTE: We should avoid using native object prototype methods,
 // since they can be overriden by the client code. (GH-245)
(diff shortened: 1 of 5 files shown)


kirovboris:
lgtm

src/client/sandbox/xhr.js
```diff
@@ -21,6 +23,10 @@ export default class XhrSandbox extends SandboxBase {
         this.corsSupported = typeof xhr.withCredentials !== 'undefined';
     }
 
+    static isOpenedXhr (obj) {
+        return obj[IS_OPENED_XHR];
+    }
+
```
miherlosev: We can detect that xhr request is opened using `readyState` property.
After calling `xhr.open(...)` the `readyState` become equal 1.
So you can rewrite method isOpenedXhr as:
```javascript
static isOpenedXhr (obj) {
       return isXhr(obj) && obj.readyState >= 1 ;
 }
```

miherlosev:
\r-

src/client/sandbox/xhr.js
churkin: What is `isXhr` function?

miherlosev:
After discussing with @churkin:
There is a problem with `isXhr` checking.
So a new flag `hammerhead|xhr|is-opened-xhr` is a good solution for this.

lgtm.

PR #562, opened 2016-04-29
Replacement-based JS instrumentation (close #135)

### LINKED ISSUE #135: Replacement-based JS instrumentation
What if we'll modify our code instrumentation, so it will not regenerate whole AST, but just chunks what we need to replace:

1. Parse JS with the node location info
2. Find nodes which we need to modify
3. Modify node and instantly generate JS code for it, when using location info replace substrings in the source code.

Pros:
- Performance. Code generation is a bottle neck in the whole mechanism, reducing it to just individual nodes should give us significant performance gain
- Quality. We will preserve original formatting and keep comments in place (which is sometimes quite important, e.g.: bugs with `func.toString()`)

\cc @churkin

churkin:
\cc @inikulin @miherlosev

### DIFF
--- a/src/processing/script/index.js
+++ b/src/processing/script/index.js
@@ -74,26 +74,21 @@ function getAst (src, isObject) {
     }
 }
 
-function getCode (ast, src, isObject, beautify) {
+function getCode (ast, src) {
     var code = generate(ast, {
         format: {
             quotes:     'double',
             escapeless: true,
-            compact:    !beautify
+            compact:    true
         }
     });
 
-    if (isObject)
-        code = code.replace(OBJECT_WRAPPER_RE, '$1');
-
-    return removeTrailingSemicolonIfNecessary(code, src);
+    return src ? removeTrailingSemicolonIfNecessary(code, src) : code;
 }
 
 
 // Analyze code
 function analyze (code) {
-    code = removeHtmlComments(code);
-
     var isObject = OBJECT_RE.test(code);
     var ast      = getAst(code, isObject);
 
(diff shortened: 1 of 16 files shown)


churkin:
@testcafe-build-bot \retest

src/processing/script/transformers/empty-function.js
```diff
@@ -0,0 +1,23 @@
...
+
+// Transform:
+// function() {} --> function() {__emptyBody$Marker;});
+
+export default {
+    nodeReplacementRequireTransform: false,
+
+    nodeTypes: [Syntax.BlockStatement],
+
+    condition: (node, parent) => {
```
inikulin: BTW, why we need this transform?

src/processing/script/index.js
```diff
@@ -117,19 +112,45 @@ export function isScriptProcessed (code) {
...
+
+    chunks.push(script.substring(index));
+
+    return chunks.join('');
+}
+
+export function processScript (src, withHeader) {
     var { bom, preprocessed } = preprocess(src);
-    var { ast, isObject }     = analyze(preprocessed);
+    var withoutHtmlComments = removeHtmlComments(preprocessed);
```
inikulin: Formatting

src/processing/script/transformers/node-replacer.js
```diff
@@ -0,0 +1,28 @@
+// -------------------------------------------------------------
+// WARNING: this file is used by both the client and the server.
+// Do not use any browser or node-specific API!
+// -------------------------------------------------------------
+
+export default function replace (node, newNode, parent, key) {
```
inikulin: It should be called `replaceNode` and file should be called `replace-node.js`

inikulin:
\r-

src/processing/script/index.js
churkin: It seems that the WebStorm is not able to properly format similar constructions :(

churkin:
FPR

inikulin:
lgtm

miherlosev:
lgtm.
Waiting for Health Monitor results...

PR #576, opened 2016-05-12
Wrong processing of the script (close #573)

### LINKED ISSUE #573: Wrong processing of the script
Let's consider the following script:
```javascript
script.src += params.join("&");
```
It is processed correctly:
```javascript
__set$(script, "src", __get$(script, "src") + params.join("&"));
```

But if we add a condition to this script:
```javascript
if (1) { } script.src += params.join("&");
```
What we get is:
```javascript
if (1) { } script.src = __get$(script, "src") + params.join("&");
```

churkin:
\cc @LavrovArtem @miherlosev

### DIFF
--- a/src/processing/script/index.js
+++ b/src/processing/script/index.js
@@ -121,12 +121,15 @@ export function applyChanges (script, changes, isObject) {
         return script;
 
     for (var i = 0; i < changes.length; i++) {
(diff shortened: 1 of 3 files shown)


LavrovArtem:
lgtm

miherlosev:
lgtm

PR #582, opened 2016-05-13
Fix wrong 'switch' construction processing

churkin:
\cc @miherlosev @LavrovArtem

### DIFF
--- a/src/processing/script/index.js
+++ b/src/processing/script/index.js
@@ -120,6 +120,8 @@ export function applyChanges (script, changes, isObject) {
     if (!changes.length)
         return script;
 
+    changes.sort((a, b) => a.start - b.start);
+
     for (var i = 0; i < changes.length; i++) {
         var change      = changes[i];
         var changeStart = change.start + indexOffset;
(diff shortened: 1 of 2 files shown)


churkin:
FPR

miherlosev:
lgtm

LavrovArtem:
lgtm

src/processing/script/index.js
```diff
@@ -120,6 +120,8 @@ export function applyChanges (script, changes, isObject) {
     if (!changes.length)
         return script;
 
+    changes.sort((a, b) => a.start - b.start);
```
inikulin: Maybe we should add comment with example which will describe why we need this?

churkin: In combination with the test, I think it clear

inikulin: Well, it's not. I can't figure it out still, just your verbal description helped somehow. Also, when I see code I don't see the test. And BTW, have you tried string concat instead of arr.join for chunks? In V8 concatenation is optimized and should be more robust.

PR #586, opened 2016-05-17
fix eslint error (close #585)

### LINKED ISSUE #585: Fix eslint error 'Parsing error: 'import' and 'export' may appear only with 'sourceType: module''

miherlosev:
The problem that described in #585 was fixed by refreshing node modules.
But the new linting problem is appeared.
@LavrovArtem @AlexanderMoskovkin

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -180,8 +180,14 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
                 },
 
                 get: el => LocationAccessorsInstrumentation.isLocationWrapper(el) ? el.href : PropertyAccessorsInstrumentation._getUrlAttr(el, 'href'),
-                set: (el, value) => LocationAccessorsInstrumentation.isLocationWrapper(el) ?
-                                    el.href = destLocation.resolveUrl(value, document) : el.setAttribute('href', value)
+                set: (el, value) => {
+                    if (LocationAccessorsInstrumentation.isLocationWrapper(el))
+                        el.href = destLocation.resolveUrl(value, document);
+                    else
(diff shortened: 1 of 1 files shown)


src/client/sandbox/code-instrumentation/properties/index.js
```diff
@@ -180,8 +180,14 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
...
                 get: el => LocationAccessorsInstrumentation.isLocationWrapper(el) ? el.href : PropertyAccessorsInstrumentation._getUrlAttr(el, 'href'),
-                set: (el, value) => LocationAccessorsInstrumentation.isLocationWrapper(el) ?
-                                    el.href = destLocation.resolveUrl(value, document) : el.setAttribute('href', value)
+                set: (el, value) => {
+                    if (LocationAccessorsInstrumentation.isLocationWrapper(el))
+                        el.href = destLocation.resolveUrl(value, document);
+                    else
+                        el.setAttribute('href', value);
+
+                    return void 0;
```
inikulin: It should return `href`

miherlosev: more precisely:
if `el.href=<value>` is called then we should return  <value>.
else retrun `void 0`.

tmp

inikulin: BTW FYI @churkin @miherlosev @LavrovArtem WHATWG just standartized location object: https://html.spec.whatwg.org/multipage/browsers.html#the-location-interface Might be useful

miherlosev:
FPR

LavrovArtem:
lgtm

inikulin:
lgtm

AlexanderMoskovkin:
lgtm

PR #612, opened 2016-06-01
fix Object.define in fetchSundbox. Bump version

miherlosev:
@churkin @LavrovArtem 
It needs for https://github.com/DevExpress/testcafe/pull/548

Description of the problem:
There are two chains for fetch promise. The first one - in the customer code, second - in the testcafe's request barrier. So, the overriding of the `fetchPromise.then`([see](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/fetch.js#L70)) handler arguments is called two times. And then we perform `Object.defineProperty` the second time we getting a client error `TypeError: Cannot redefine property 'type'`.
Since `Object.defineProperty` the `configurable` property has `false` value by default ([see] (https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty))

### DIFF
--- a/src/client/sandbox/fetch.js
+++ b/src/client/sandbox/fetch.js
@@ -72,22 +72,18 @@ export default class FetchSandbox extends SandboxBase {
                     throw new TypeError();
 
                 Object.defineProperty(response, 'type', {
-                    get: function () {
-                        return FetchSandbox._getResponseType(response);
-                    },
-                    set: () => void 0
+                    get:          () => FetchSandbox._getResponseType(response),
+                    set:          () => void 0,
+                    configurable: true
                 });
 
                 var responseStatus = response.status === SAME_ORIGIN_CHECK_FAILED_STATUS_CODE ? 0 : response.status;
 
                 Object.defineProperty(response, 'status', {
-                    get: function () {
-                        return responseStatus;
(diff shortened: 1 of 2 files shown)


miherlosev:
@testcafe-build-bot \retest

src/client/sandbox/fetch.js
```diff
@@ -75,7 +75,8 @@ export default class FetchSandbox extends SandboxBase {
                     get: function () {
                         return FetchSandbox._getResponseType(response);
                     },
```
churkin: get: () => FetchSandbox._getResponseType(response);

src/client/sandbox/fetch.js
```diff
@@ -84,7 +85,8 @@ export default class FetchSandbox extends SandboxBase {
                     get: function () {
                         return responseStatus;
                     },
```
churkin:
```javascript
get: () => responseStatus,
```

miherlosev:
FPR

miherlosev:
@testcafe-build-bot \retest

churkin:
lgtm

LavrovArtem:
lgtm

PR #622, opened 2016-06-06
fix 'HH opens new window if the form target attribute points to a nonexistent window name' (close #247)

### LINKED ISSUE #247: HH opens new window if the form target attribute points to a nonexistent window name

miherlosev:
After discussion with @churkin we decide to implement the partial fix:
If `element.target` contains non-existing window name then we set to target `_self` value.
Now, we don't update existing element's url after changing set of the avalible window names.
Because it may cause perfomance problem - we need to update all `form.action` and `a.href` attribute values for all windows in the page.

### DIFF
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -14,6 +14,9 @@ import { isPageHtml, processHtml } from '../../utils/html';
 import transport from '../../transport';
 import getNativeQuerySelectorAll from '../../utils/get-native-query-selector-all';
 import { HASH_RE } from '../../../utils/url';
+import * as windowsStorage from '../windows-storage';
+
+const KEYWORD_TARGETS = ['_blank', '_self', '_parent', '_top'];
 
 export default class ElementSandbox extends SandboxBase {
     constructor (nodeSandbox, uploadSandbox, iframeSandbox, shadowUI) {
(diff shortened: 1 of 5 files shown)


miherlosev:
ping @churkin @LavrovArtem

src/client/sandbox/node/element.js
```diff
@@ -29,6 +30,10 @@ export default class ElementSandbox extends SandboxBase {
         this.BEFORE_FORM_SUBMIT = 'hammerhead|event|before-form-submit';
     }
 
+    static _isNamedTarget (value) {
+        return value[0] !== '_';
+    }
```
churkin: Is '_' can not be in the middle of target?

churkin:
-\r

src/client/sandbox/node/element.js
miherlosev: `_isNamedTarget` is a unclear function name.
It checks that target is not in reserved words - `_self`, `_parent`, `_top`.
Maybe rename it to `_isBuildInTarget` or `_isReservedTarget`.
@churkin What do you think?

miherlosev:
FPR

src/client/sandbox/node/element.js
```diff
@@ -27,6 +28,12 @@ export default class ElementSandbox extends SandboxBase {
         this.overridedMethods = null;
 
         this.BEFORE_FORM_SUBMIT = 'hammerhead|event|before-form-submit';
+
+        this.keywordTargets = ['_blank', '_self', '_parent', '_top'];
+    }
+
+    _isKeywordTarget (value) {
+        return this.keywordTargets.indexOf(value) !== -1;
```
churkin: Is keyword target case sensitivity?

src/client/sandbox/node/element.js
```diff
@@ -27,6 +28,12 @@ export default class ElementSandbox extends SandboxBase {
         this.overridedMethods = null;
 
         this.BEFORE_FORM_SUBMIT = 'hammerhead|event|before-form-submit';
+
+        this.keywordTargets = ['_blank', '_self', '_parent', '_top'];
```
churkin: Make KEYWORD_TARGETS const

```javascript
const KEYWORD_TARGETS = ['_blank', '_self', '_parent', '_top'];
```

src/client/sandbox/node/element.js
miherlosev: No. Keyword target is case insensitive.

miherlosev:
FPR

churkin:
lgtm

LavrovArtem:
lgtm

PR #628, opened 2016-06-08
The col tag is incorrectly wrapped in the processHtml function (close #625)

### LINKED ISSUE #625: The col tag is incorrectly wrapped in the processHtml function
Before the HTML is processed, the self-closing `col` tag is wrapped as follows:
`<col/> -> <fake_tag_name_col/>`
But `fake_tag_name_col` is not self-closing, so that the markup gets broken in IE.

LavrovArtem:
/cc @churkin @georgiy-abbasov


churkin:
lgtm

georgiy-abbasov:
lgtm

PR #641, opened 2016-06-27
fix "Location assignment doesn't work" (close #640)

### LINKED ISSUE #640: Location assignment doesn't work
Run in playground:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>
<input type="button" id="btn">

<script>
    document.querySelector('#btn').addEventListener('click', function () {
        location = 'https://google.com';
    });
</script>
</body>
</html>
```

Clicking `btn` doesn't change location. I'm surprised we didn't encountered it yet. Also, it's worth checking other properties that instrumented the way location does. BTW, replacing `location` with `document.location` or `window.location` works just fine.

miherlosev:
@churkin @inikulin 

Also I planned to add functional test in `testcafe` for case with top level window.

### DIFF
--- a/src/client/sandbox/code-instrumentation/location/index.js
+++ b/src/client/sandbox/code-instrumentation/location/index.js
@@ -2,6 +2,9 @@ import LocationWrapper from './wrapper';
 import SandboxBase from '../../base';
 import { isLocation } from '../../../utils/dom';
 import INSTRUCTION from '../../../../processing/script/instruction';
+import * as destLocation from '../../../utils/destination-location';
+import { getResourceTypeString } from '../../../../utils/url';
+import * as urlUtils from '../../../utils/url';
 
 const LOCATION_WRAPPER = 'hammerhead|location-wrapper';
 
@@ -44,11 +47,17 @@ export default class LocationAccessorsInstrumentation extends SandboxBase {
         Object.defineProperty(window, INSTRUCTION.setLocation, {
             value: (location, value) => {
                 if (isLocation(location)) {
-                    location = value;
+                    var resourceType = null;
 
-                    return location;
+                    if (window !== window.top) {
+                        value        = destLocation.resolveUrl(value, window.top.document);
+                        resourceType = getResourceTypeString({ isIframe: true });
+                    }
+
+                    window.location = urlUtils.getProxyUrl(value, null, null, null, resourceType);
+
+                    return window.location;
                 }
-
                 return null;
             },
             configurable: true
--- a/src/client/sandbox/code-instrumentation/location/wrapper.js
+++ b/src/client/sandbox/code-instrumentation/location/wrapper.js
@@ -1,7 +1,7 @@
 import createPropertyDesc from '../../../utils/create-property-desc';
 import { get as getDestLocation, getParsed as getParsedDestLocation } from '../../../utils/destination-location';
 import { getProxyUrl, changeDestUrlPart, parseProxyUrl, parseResourceType } from '../../../utils/url';
-import { getDomain, stringifyResourceType } from '../../../../utils/url';
+import { getDomain, getResourceTypeString } from '../../../../utils/url';
 
 export default class LocationWrapper {
     constructor (window) {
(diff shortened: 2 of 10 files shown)


inikulin:
@miherlosev Do we have a test for this case: https://github.com/DevExpress/testcafe-hammerhead/issues/640#issuecomment-228679853

inikulin:
Also, it's odd that WPT tests doesn't show this problem, do they run correctly?

miherlosev:
> Also, it's odd that WPT tests doesn't show this problem, do they run correctly?
We run tests that don't contain location change.
I cloned web platform test repository and search using keywords 'location =', ' location = ', ' location='.
See fixture list https://github.com/DevExpress/hammerhead-health-monitor/blob/master/web-platform/tests_big.json.

miherlosev:
> @miherlosev Do we have a test for this case: #640 (comment)

Create the separate issue https://github.com/DevExpress/testcafe-hammerhead/issues/642

miherlosev:
FPR

miherlosev:
ping @churkin @inikulin

src/client/sandbox/code-instrumentation/location/index.js
```diff
@@ -34,11 +37,17 @@ export default class LocationAccessorsInstrumentation extends SandboxBase {
...
         window[INSTRUCTION.setLocation] = (location, value) => {
             if (isLocation(location)) {
-                location = value;
+                var resourceType = null;
 
-                return location;
-            }
+                if (window.self !== window.top) {
+                    value        = destLocation.resolveUrl(value, window.top.document);
+                    resourceType = stringifyResourceType(true);
```
inikulin: Can we have full form here: `stringifyResourceType(true, false, false)`? Also, it took me a while to figure out what it does. First of all `stringify` suggest that you pass something to be stringified. It more likely `getResourceTypeString`. The second issue it's not clear what arguments mean here. In such cases it's always better to have options object:
`getResourceTypeString({ isIframe: true})`

inikulin:
\r-

src/client/sandbox/code-instrumentation/location/index.js
miherlosev: So we need to do the following:
1)rename `stringifyResourceType` to the `getResourceTypeString`
2)move 3 function arguments to the single parameters object

miherlosev:
FPR

miherlosev:
@testcafe-build-bot \retest

miherlosev:
Error for node 0.10.
```javascript
/home/travis/build/testcafe-build-bot/testcafe-hammerhead/node_modules/eslint-plugin-import/lib/index.js:6

const rules = exports.rules = {

^^^^^
```
Possible reason - https://github.com/DevExpress/testcafe-hammerhead/commit/62eed525fdd3b08ba9def5924dcf6e06fa0020ad

SyntaxError: Failed to load plugin import: Use of const in strict mode.

miherlosev:
FPR

src/client/sandbox/code-instrumentation/location/index.js
```diff
@@ -34,11 +37,17 @@ export default class LocationAccessorsInstrumentation extends SandboxBase {
 
         window[INSTRUCTION.setLocation] = (location, value) => {
             if (isLocation(location)) {
-                location = value;
+                var resourceType = null;
 
-                return location;
-            }
+                if (window.self !== window.top) {
```
churkin: @LavrovArtem @miherlosev  Guys, we no longer use the `window.self` property, it can be overriden `window.self = {};`, just use `window === window.top` check.

churkin:
-\r

miherlosev:
FPR
ping @inikulin @churkin

src/client/sandbox/code-instrumentation/location/index.js
inikulin: BTW, it would be nice to have eslint rule for this. @LavrovArtem can you take care of this, please?

inikulin:
lgtm

churkin:
lgtm

churkin:
rebase pls.

miherlosev:
Done.
FPR

miherlosev:
ping @churkin

PR #648, opened 2016-06-30
fix "`navigateTo` doesn't work if current location is `about:blank`" (close #645)

### LINKED ISSUE #645: `navigateTo` doesn't work if current location is `about:blank`
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/index.js#L156

miherlosev:
@inikulin @churkin

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -15,6 +15,7 @@ import * as typeUtils from './utils/types';
 import * as positionUtils from './utils/position';
 import * as styleUtils from './utils/style';
 import trim from '../utils/string-trim';
+import { isRelativeUrl, parseProxyUrl } from '../utils/url';
 import { getProxyUrl } from './utils/url';
 import { processScript } from '../processing/script';
 import { SCRIPT_PROCESSING_START_COMMENT, SCRIPT_PROCESSING_END_HEADER_COMMENT, SCRIPT_PROCESSING_END_COMMENT } from '../processing/script/header';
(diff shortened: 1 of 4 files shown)


miherlosev:
@testcafe-build-bot \retest

AlexanderMoskovkin:
we have some troubles with firefox on saucelabs after it was updated there. @georgiy-abbasov is working on this

inikulin:
lgtm

georgiy-abbasov:
@miherlosev, yesterday was released a new version of selenium which supports Firefox 47, so Sauce Labs provides this version by default for now. Unfortunately, it has an error - we cannot get properties of global objects (for example `window.global_test_results`), while publishing test results via qunit-harness. At the current moment we have only one way to keep test process stable - we should specify firefox version - 46. Make next corrections in this PR, please.
```js
{
        platform:    'Windows 10',
        browserName: 'firefox',
        version: '46'
}
```
and
```js
{
        browserName: 'firefox',
        platform:    'OS X 10.11',
        version:      '46'
}
```

miherlosev:
Ok. I will do this in the separate pull request.

inikulin:
@georgiy-abbasov 
> @miherlosev, yesterday was released a new version of selenium which supports Firefox 47, so Sauce Labs provides this version by default for now. Unfortunately, it has an error - we cannot get properties of global objects (for example window.global_test_results

What does this mean, can you elaborate? We use such properties both in testcafe and hammerhead client. Does it mean they don't work in Firefox 47?

georgiy-abbasov:
@inikulin, yes, it is.

miherlosev:
FPR

miherlosev:
Need to do:
1)set proxy url for `iframes` and `images` for special pages(`about:blank`, `about:error`).
2)in `hammerhead.navigateTo` use `parseProxyUrl.destUrl`.

miherlosev:
FPR

inikulin:
lgtm

churkin:
lgtm

PR #704, opened 2016-07-28
fix 'domUtils.isIframeWithoutSrc returns wrong result after documentopen in same-domain iframe' (close #703)

### LINKED ISSUE #703: domUtils.isIframeWithoutSrc returns wrong result after document.open in same-domain iframe
see test for reproduce:
```javascript
var iframe  = document.createElement('iframe');

iframe.src = <same-domain-url>
iframe.id = 'test_unique_id';
    window.QUnitGlobals.waitForIframe(iframe)
        .then(function () {
            var result = domUtils.isIframeWithoutSrc(iframe);
            ok(!result);

            iframe.contentDocument.open();

            result = domUtils.isIframeWithoutSrc(iframe); // result equal 'true'. It is wrong

            iframe.parentNode.removeChild(iframe);
            start();
        });
});
```

miherlosev:
@churkin @LavrovArtem

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -365,8 +365,16 @@ export function isIframeWithoutSrc (iframe) {
     var parsedParentWindowLocation = urlUtils.parseProxyUrl(parentWindowWithSrc.location.toString());
     var parentWindowLocation       = parsedParentWindowLocation ? parsedParentWindowLocation.destUrl : parentWindowWithSrc.location.toString();
 
-    if (iframeDocumentLocationHaveSupportedProtocol)
+    if (iframeDocumentLocationHaveSupportedProtocol) {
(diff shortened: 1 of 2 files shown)


miherlosev:
@testcafe-build-bot \retest

miherlosev:
FPR

LavrovArtem:
lgtm

miherlosev:
I checked the following iframe types:
1. iframe with cross-domain src.
2. iframe with same-domain src.
3. iframe with about:blank
4. iframe with empty src.
5. iframe without src attribute.
6. iframe with not supported protocol (javascript protocol).

I don't find case then iframe without src has document.location as top window with src.
So, FPR

churkin:
lgtm

PR #705, opened 2016-07-29
fix 'Operator 'instanceof' wrong work with Blob class' (close #690)

### LINKED ISSUE #690: Operator 'instanceof' wrong work with Blob  class
Test code from http://html5test.com
```javascript
var b = new Blob();
b instanceof Blob;
> false
```
Also we need to check all overriden constructors (Image, Worker, FontFace, etc)

miherlosev:
@churkin @AlexanderMoskovkin


miherlosev:
FPR

churkin:
lgtm

miherlosev:
@LavrovArtem review please

LavrovArtem:
lgtm

PR #719, opened 2016-08-04
fix 'The base url is incorrectly saved and restored in html utils' (close #718)

### LINKED ISSUE #718: The base url is incorrectly saved and restored in html utils
We are losing the last slash in `http://example.com/path/path/`.

LavrovArtem:
/cc @churkin @miherlosev @VasilyStrelyaev


VasilyStrelyaev:
lgtm

miherlosev:
lgtm

churkin:
lgtm

PR #732, opened 2016-08-12
'Permission denied' error is raised when an iframe with a nested iframe is reloaded (close #727)

### LINKED ISSUE #727: "Permission denied" error is raised when an iframe with a nested iframe is reloaded
The error is appeared in the `hammerhead.js` script. As a result the `window['%hammerhead%'] is undefined` error appeared in the `task.js` script (reproduced in IE11, IE9).

Create the following pages to reproduce the issue:

`page1.html`
```html
...
<iframe src="page2.html"></iframe>
...
```

`page2.html`
```html
...
<iframe src="page3.html"></iframe>

<script>
    window.setTimeout(function () {
        window.location.reload();
    }, 1000);
</script>
...
```

`page3.html`
```html
...
Just an empty page
...
```

Open `page1.html` in the `hammerhead-playground`. After the second the iframe is reloaded and  errors are appeared [here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/backup.js#L16) and [here](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/windows-storage.js#L20).

It brokes test execution in `testcafe`. I suppose it can be one of the reason of some our blinking functional tests in IE (but it just an assumption).

churkin:
\сс @AlexanderMoskovkin @LavrovArtem

### DIFF
--- a/src/client/sandbox/backup.js
+++ b/src/client/sandbox/backup.js
@@ -2,6 +2,21 @@ import { getTopSameDomainWindow, getFrameElement } from '../utils/dom';
 
 const SANDBOX_BACKUP = 'hammerhead|sandbox-backup';
 
+
+function findRecord (storage, iframe) {
+    for (var i = storage.length - 1; i >= 0; i--) {
+        try {
+            if (storage[i].iframe === iframe)
+                return storage[i];
+        }
+        catch (e) {
+            storage.splice(i, 1);
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


AlexanderMoskovkin:
lgtm

LavrovArtem:
lgtm

PR #740, opened 2016-08-18
fix 'Very slow proxing for http://www.answers.com' (close#657)

### LINKED ISSUE #657: Very slow proxing for http://www.answers.com
see https://testcafe-hhhm.devexpress.com/history?timestamp=1467659469614&siteUrl=http://www.answers.com/.
It happens because we cyclically send `set-cookie` service message. (see video - http://screencast.com/t/vBYd5NfW3j).
Need to investigate why we do this.

miherlosev:
see discussion in #657  for more details.
@churkin @inikulin 

Pull request changes:
1)fix #657 
2)remove unnecessary `processDomMeth` calls from tests

### DIFF
--- a/src/client/sandbox/cookie.js
+++ b/src/client/sandbox/cookie.js
@@ -6,8 +6,21 @@ import * as cookieUtils from '../utils/cookie';
 import { isCrossDomainWindows } from '../utils/dom';
 import transport from '../transport';
 import trim from '../../utils/string-trim';
+import throttle from '../utils/throttle';
+
+const MIN_COOKIE_SYNC_INTERVAL = 100;
 
 export default class CookieSandbox extends SandboxBase {
+    constructor () {
+        super();
+
+        this.throttledCookieSync = throttle(cookieMsg => {
+            transport.queuedAsyncServiceMsg(cookieMsg);
+        }, MIN_COOKIE_SYNC_INTERVAL);
+
+        this.MIN_COOKIE_SYNC_INTERVAL = MIN_COOKIE_SYNC_INTERVAL;
+    }
+
     _getSettings () {
         var windowSettings = this.window !== this.window.top && !isCrossDomainWindows(this.window, this.window.top) ?
                              this.window.top['%hammerhead%'].get('./settings') : settings;
@@ -129,12 +142,11 @@ export default class CookieSandbox extends SandboxBase {
 
         var setCookieMsg = {
             cmd:    COMMAND.setCookie,
-            cookie: value,
+            cookie: this.getCookie(),
             url:    document.location.href
         };
 
-        // NOTE: Meanwhile, synchronize cookies with the server cookie jar.
-        transport.queuedAsyncServiceMsg(setCookieMsg);
+        this.throttledCookieSync(setCookieMsg);
 
         return value;
     }
--- a/src/client/utils/throttle.js
+++ b/src/client/utils/throttle.js
@@ -1,0 +1,31 @@
+import nativeMethods from '../sandbox/native-methods';
+
+export default function throttle (func, ms) {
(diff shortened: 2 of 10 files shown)


src/client/utils/throttle.js
```diff
@@ -0,0 +1,31 @@
...
+
+export default function throttle (func, ms) {
+    var isSuspend = false;
+    var savedArgs = null;
+    var savedThis = null;
+
+    function wrapper () {
+        if (isSuspend) {
+            savedArgs = arguments;
+            savedThis = this;
```
inikulin: I'd rather go without saving context, expecting that `func` is already bound

src/client/utils/throttle.js
```diff
@@ -0,0 +1,31 @@
...
+            return;
+        }
+
+        func.apply(this, arguments);
+
+        isSuspend = true;
```
inikulin: `suspended`

inikulin:
\r-

src/client/sandbox/cookie.js
```diff
@@ -129,12 +142,11 @@ export default class CookieSandbox extends SandboxBase {
...
         var setCookieMsg = {
             cmd:    COMMAND.setCookie,
-            cookie: value,
+            cookie: this.getCookie(),
             url:    document.location.href
         };
 
-        // NOTE: Meanwhile, synchronize cookies with the server cookie jar.
-        transport.queuedAsyncServiceMsg(setCookieMsg);
+        this.throttledCookieSync(setCookieMsg);
```
inikulin: it should be named `syncCookies`, `Sync` suffix means that functions has `sync` nature, but we need to synchronize cookies here, also we can go without `throttled`, it's just an implementation details.

src/client/utils/throttle.js
```diff
@@ -0,0 +1,31 @@
+import nativeMethods from '../sandbox/native-methods';
+
+export default function throttle (func, ms) {
+    var isSuspend = false;
```
inikulin: I believe it's a little bit more complicated then that. We set cookie for particular URL, what if we throttle setting cookies for particular URL and page unload happens? I guess we should perform additional synchronization on before unload.

inikulin: We need tests for these scenarios as well

inikulin:
\r-

src/client/utils/throttle.js
miherlosev: After an additional investigation I found that cookie synchronization every <value> ms is a bad choise.
We can get out of cookie sync for the following case:
```javascript
document.cookie = 'key1=value1';
document.cookie = 'key2=value2';
xhrRequest with cookie (or another request with cookie).
```

inikulin:
We can attach cookies with xhr in that case in special hader

src/client/utils/throttle.js
inikulin: We can attach cookies with xhr in that case if synchronization is not done yet

miherlosev: The task is more complicated.
We need an addition cookie sync for the following cases (if it is not done yet):
1)XmlHttpRequest and Fetch requests (store in custom header)
2)Images, script and style loading (pass via additional url parameter)
3)Page unload (store into custom input)

Waiting for https://github.com/DevExpress/testcafe-hammerhead/pull/721

miherlosev: I think we can completely remove throttling and `setCookie` service messages and
perform cookie synchronization with server only for cases from list above.

inikulin: Not sure about it

miherlosev:
We will rethink cookie processing mechanism -https://github.com/DevExpress/testcafe-hammerhead/issues/1137

PR #761, opened 2016-09-01
remove ie8 related code

miherlosev:
@churkin @AlexanderMoskovkin


miherlosev:
Oops.
`MSEventObj` (IE9, IE01) doesn't have `preventDefault` method.

Issue #722, opened 2016-08-05
Submit from a top window to a frame doesn't work


LavrovArtem:
Server for reproducing:

``` javascript
var http = require('http');

http.createServer(function(req, res) {
    res.setHeader('content-type', 'text/html')

    switch (req.url) {
        case '/':
            return res.end(`
                <iframe name="iframe_name"></iframe>
                <form action="form.html"></form>
                <script>
                    var iframe = document.querySelector("iframe");
                    var form = document.querySelector("form");

                    var storedTarget = form.target;
                    var storedAction = form.action;

                    form.target = "iframe_name";
                    form.action = "new.html";

                    form.submit();

                    form.target = storedTarget;
                    form.action = storedAction;
                </script>
            `);
        case '/form.html?':
            return res.end('<h1>Fail!!!</h1>');
        case '/new.html?':
            return res.end('<h1>Success!!!</h1>');
        default:
            res.end();
    }
}).listen(5000);
```

LavrovArtem:
[temporary solution](https://github.com/DevExpress/testcafe-hammerhead/issues/804)

churkin:
We have a few bugs because of this behavior. The error appeared after transitions to promises, due to the fact that the 'pinkie' library implements the asynchronous promises. Using callback is bad, but in this case we have to return them. All this mechanism disappears after https://github.com/DevExpress/testcafe-hammerhead/issues/667 fix.

Farfurix:
I didn't manage to reproduce it with Chrome, Firefox, and Windows 10. So, I'll close this issue.

PR #806, opened 2016-09-20
temporary solution for #722 (close #804, close #799)

### LINKED ISSUE #799: [Testing day 0.9.0-rc1] Error raised on page https://jsfiddle.net/ in hammerhead-playground
For more info see https://github.com/DevExpress/testcafe/issues/826

### LINKED ISSUE #804: We need a temporary solution for #722
If the queue of cookie message is empty, we need call a callback function synchronously.

LavrovArtem:
/cc @miherlosev @churkin @VasilyStrelyaev

### DIFF
--- a/src/client/transport.js
+++ b/src/client/transport.js
@@ -166,21 +166,21 @@ class Transport extends EventEmitter {
         sendMsg();
     }
 
-    waitCookieMsg () {
-        return new Promise(resolve => {
-            var handler = () => {
-                if (!this._cookieMsgInProgress()) {
-                    this.off(this.MSG_RECEIVED_EVENT, handler);
+    // NOTE: We don't use a promise here because if the cookie message queue is empty,
+    // we need to call a callback function synchronously (GH-722)
+    waitCookieMsg (callback) {
+        var handler = () => {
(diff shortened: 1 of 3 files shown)


src/client/transport.js
```diff
@@ -166,21 +166,21 @@ class Transport extends EventEmitter {
...
-        return new Promise(resolve => {
-            var handler = () => {
-                if (!this._cookieMsgInProgress()) {
-                    this.off(this.MSG_RECEIVED_EVENT, handler);
-
-                    resolve();
-                }
-            };
+    // TODO: Don't use the Promise because if the queue of cookie message is empty,
+    // we need call a callback function synchronously (GH-722)
```
VasilyStrelyaev: So, you got rid of the promise. Then it's maybe a `NOTE` not `TODO`?

```
    // NOTE: We don't use a promise here because if the cookie message queue is empty,
    // we need to call a callback function synchronously (GH-722)
```

LavrovArtem: Yes, I confused

LavrovArtem:
FPR

LavrovArtem:
FPR @churkin

PR #813, opened 2016-09-23
additional case for the #803 fix (close #803)

### LINKED ISSUE #803: The `postMessage` mechanism does not work for embedded frames
top window
    -> iframe
         -> iframe without src

If we try to send a message from the embedded iframe to the parent iframe we will get an error.

churkin:
\cc @LavrovArtem @miherlosev


churkin:
The test will be added in the context of #814

Issue #823, opened 2016-09-28
PhantomJS: Script error on Object.defineProperty for DOM element and Event objects


miherlosev:
We setup events and dom elements in source and test code using `Object.defineProperty`.
But PhantomJS doesn't support this - see issue https://github.com/ariya/phantomjs/issues/13895.
Need to investigate way to drop using `Object.defineProperty`.

churkin:
Maybe we should try https://github.com/ianstormtaylor/create-event/pull/9 solution?

miherlosev:
I tried this dark magic and got that it works only for attribute properties.
For event properties it doesn't work.

Step to reproduce:
- download latest PhantomJS from http://phantomjs.org/download.html
- run `phantomjs test.js`

#### Test.js
```js
/*Dom element*/ /*Worked*/
console.log('Dom element');
console.log('document.body.attributes.length before redefine:', document.body.attributes.length);
document.body.attributes.constructor.prototype.__defineGetter__('length', function () {return 1000; }); //work
console.log('document.body.attributes.constructor.prototype.__defineGetter__:', document.body.attributes.length);
document.body.attributes.__defineGetter__('length', function () { return 2000; });// don't work
console.log('document.body.attributes.__defineGetter__:',document.body.attributes.length);



/*Event*/
console.log('Event');
var evt = document.createEvent('StorageEvent');

evt.initStorageEvent('storage',false,false,'key','oldValue','newValue', 'http://fabrikam.com/alleged.html', { test: 1 });
console.log('storageArea before redefine:', evt.storageArea);

evt.constructor.prototype.__defineGetter__('storageArea', function () { return window.sessionStorage; });//don't work
console.log('evt.constructor.prototype.__defineGetter__:', evt.storageArea);

evt.constructor.prototype.__defineGetter__.call(evt, 'storageArea', function () { return window.sessionStorage; });//dont't work
console.log('evt.constructor.prototype.__defineGetter__.call:', evt.storageArea);

evt.__defineGetter__('storageArea', function () { return window.sessionStorage; });//dont't work
console.log('evt.__defineGetter__:', evt.storageArea);

evt.__proto__.__defineGetter__('storageArea', function () { return window.sessionStorage; });//dont't work
console.log('evt.__proto__.__defineGetter__:', evt.storageArea);

window.StorageEvent.__defineGetter__.call(evt, 'storageArea', function () { return window.sessionStorage; });//don't work
[...]

PR #840, opened 2016-10-07
fix `PhantomJS: Add 'Upload' support` (close #829)

### LINKED ISSUE #829: PhantomJS: Add 'Upload' support
PhantomJS hasn't `fileRead.readAsDataURL` method that used in `InfoManager.loadFileListData`.
So we should use another way to support 'Upload' functionality.

LavrovArtem:
/cc @churkin @miherlosev


LavrovArtem:
ping

miherlosev:
I think at first we should fir `Object.defineProperty` bug (see #823).
And we don't need to change test code.

PR #842, opened 2016-10-07
Restart listening only in top windows (closes #839)

### LINKED ISSUE #839: Event listeners added twice in an iframe after document.write
Because https://github.com/DevExpress/testcafe-hammerhead/blob/75922f0a9eb136a8c704f1e2b0824e64a954e691/src/client/sandbox/index.js#L163 is intended to work in top-level windows only. Related to https://isc.devexpress.com/Thread/WorkplaceDetails/T236223 and ASP tests.

AndreyBelym:
\cc @churkin @helen-dikareva

### DIFF
--- a/src/client/sandbox/index.js
+++ b/src/client/sandbox/index.js
@@ -160,7 +160,7 @@ export default class Sandbox extends SandboxBase {
 
         this.event.initDocumentListening();
 
-        if (isWebKit)
+        if (window.top === window.self && isWebKit)
             this.event.listeners.restartElementListening(window);
 
         this.shadowUI.attach(window);
--- a/src/client/sandbox/index.js
+++ b/src/client/sandbox/index.js
@@ -160,7 +160,7 @@ export default class Sandbox extends SandboxBase {
 
         this.event.initDocumentListening();
 
-        if (window.top === window.self && isWebKit)
+        if (isWebKit && window.top === window)
             this.event.listeners.restartElementListening(window);
 
         this.shadowUI.attach(window);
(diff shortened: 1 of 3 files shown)


test/client/data/iframe/window-event-listeners.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title></title>
+    <meta charset="utf-8">
+    <script type="text/javascript">
+        window.eventListenersCount = {};
+
+        var origWindowAddEventListener = window.constructor.prototype.addEventListener;
```
helen-dikareva: may be `originWindowAddEventListener `

src/client/sandbox/index.js
```diff
@@ -160,7 +160,7 @@ export default class Sandbox extends SandboxBase {
 
         this.event.initDocumentListening();
 
-        if (isWebKit)
+        if (window.top === window.self && isWebKit)
```
LavrovArtem: May be ```window.top === window```?

AndreyBelym: Maybe there is a dedicated function in Hammerhead utils to do this check?

AlexanderMoskovkin: We should not use window.self because of https://github.com/DevExpress/testcafe/issues/632

@AndreyBelym it's enough to use `window.top === window` because no one can override `window` in our code
@LavrovArtem maybe we should add this rule to eslint-plugin-hammerhead?

helen-dikareva: https://github.com/DevExpress/testcafe/blob/5cfc571037c21d470e7d5da489974a16cdfd1439/src/client/core/utils/dom.js#L427
according to comment we need to swap `window.top === window.self` and `isWebKit`

AndreyBelym:
FPR

PR #888, opened 2016-10-24
refactor target attr tests

miherlosev:
@churkin @LavrovArtem 
Changes:
- remove tests for inheritance target attribute from base element (because target of the base element already contains corrected value)
- remove dublicated tests
- fix target attribute removing
- add test for empty target attribute value


miherlosev:
FPR

PR #897, opened 2016-10-31
refactor target related code for pageNavigationWatch

miherlosev:
@churkin @AlexanderMoskovkin 

I removed the test `only one of several handlers must be called (document handlers) (T233158)` because it presents in test file twice.

### DIFF
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -546,6 +546,8 @@ export default class ElementSandbox extends SandboxBase {
         window.HTMLTableRowElement.prototype.insertCell    = this.overridedMethods.insertCell;
         window.HTMLFormElement.prototype.submit            = this.overridedMethods.formSubmit;
 
+        this._setValidBrowsingContextOnElementClick(window);
+
         // NOTE: Cookie can be set up for the page by using the request initiated by img.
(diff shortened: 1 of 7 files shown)


src/client/sandbox/node/element.js
```diff
@@ -574,11 +576,18 @@ export default class ElementSandbox extends SandboxBase {
         el.setAttribute('target', storedAttr || el.target);
     }
 
-    _setValidBrowsingContextOnClick (el) {
-        el.addEventListener('click', () => {
+    _setValidBrowsingContextOnElementClick (window) {
+        window.addEventListener('click', e => {
```
AlexanderMoskovkin: I suppose we should call `sandbox.addInternalEventListener` here instead `element.addEventListener` because now [this handler](https://github.com/DevExpress/testcafe-hammerhead/pull/897/files#diff-d02008885b9c21dddc064a025f83c5ceR79) will be called before yours

miherlosev:
FPR

src/client/sandbox/node/element.js
AlexanderMoskovkin: What about this comment?

miherlosev:
FPR

PR #898, opened 2016-10-31
Bump version to v.10.1.0

churkin:
\cc @AlexanderMoskovkin


AlexanderMoskovkin:
I've checked on Saucelabs - all tests are ok, but Saucelabs can't run iOS 10.0 vm because of some internal problems. So, I merge this

PR #909, opened 2016-11-03
TableData is focusable element for now (closes #901)

### LINKED ISSUE #901: Make `<td>`  element focusable for IE via domUtils
https://msdn.microsoft.com/en-us/library/ms534654(VS.85).aspx
>> The following elements can have focus by default but are not tab stops. These elements can be set as tab stops by setting the tabIndex property to a positive integer. applet, div, frameSet, span, table, **td**.

georgiy-abbasov:
\cc @churkin, @miherlosev

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -489,6 +489,9 @@ export function isElementFocusable (el) {
                matches(el, 'a[href]');
     }
 
+    if (isTableDataElement(el) && isIE)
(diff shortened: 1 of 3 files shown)


georgiy-abbasov:
Original ASP test, for which was created [this regression test](http://screencast.com/t/dJLzqQ6sr3X), passes with this Hammerhead fix

PR #929, opened 2016-11-10
fn.constructor should return wrapped Function object (closes #913)

### LINKED ISSUE #913: [todomvc] Js-error appears on the Humble demo page
page: http://todomvc.com/examples/humble
error: `Uncaught Error: interface conversion: interface is map[string]interface {}, not func(...interface {}) *js.Object`

reproduced in playground. Tested in Chrome but should be check in other browsers

AlexanderMoskovkin:
/cc @churkin @LavrovArtem

### DIFF
--- a/test/client/fixtures/sandbox/node/window-test.js
+++ b/test/client/fixtures/sandbox/node/window-test.js
@@ -132,3 +132,15 @@ test('window.onerror must be overriden (B238830)', function () {
     ok(!error);
 });
 
+test('the constructor field of a function should return wrapped Function object (GH-913)', function () {
+    var f = function () {
+    };
+
+    switch (f.constructor) {
+        case Function:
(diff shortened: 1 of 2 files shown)


test/client/fixtures/sandbox/node/window-test.js
```diff
@@ -132,3 +132,15 @@ test('window.onerror must be overriden (B238830)', function () {
...
+test('the constructor field of a function should return a wrapped Function object (GH-913)', function () {
+    var f = function () {
+    };
+
+    switch (f.constructor) {
+        case Function:
+            ok(true);
+            break;
+        default:
+            ok(false);
```
inikulin: Whole `switch` can be replaced with `strictEqual(f.constructor, Function)`

PR #944, opened 2016-11-16
Should not recognize DOM elements only by 'tagName' property (closes #942)

### LINKED ISSUE #942: Uncaught TypeError: input.getAttribute is not a function on 'https://www.twitch.tv/directory'
It's reproducing in playground.

Simplified snippet to reproduce without original page:
1. Open any page via playground, 'http://example.com' for example
2. Open DevTools and type
```js
__get$({tagName:'input'}, 'autocomplete')
```

georgiy-abbasov:
\cc @churkin, @LavrovArtem, @MarinaRukavitsyna

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -121,7 +121,7 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
             },
 
             autocomplete: {
-                condition: domUtils.isInputElement,
+                condition: input => domUtils.isInputElement(input) && domUtils.isDomElement(input),
                 get:       input => input.getAttribute('autocomplete') || '',
                 set:       (input, value) => input.setAttribute('autocomplete', value)
             },
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -121,7 +121,7 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
             },
 
             autocomplete: {
-                condition: input => domUtils.isInputElement(input) && domUtils.isDomElement(input),
+                condition: domUtils.isInputElement,
                 get:       input => input.getAttribute('autocomplete') || '',
                 set:       (input, value) => input.setAttribute('autocomplete', value)
             },
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -347,7 +347,7 @@ export function isHammerheadAttr (attr) {
 }
 
 export function isIframeElement (el) {
-    return getTagName(el) === 'iframe';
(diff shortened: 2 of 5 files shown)


src/client/sandbox/code-instrumentation/properties/index.js
```diff
@@ -121,7 +121,7 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
             },
 
             autocomplete: {
-                condition: domUtils.isInputElement,
+                condition: input => domUtils.isInputElement(input) && domUtils.isDomElement(input),
```
churkin: I think we should add `isDomElement` everywhere - `isInputElement`, `isHtmlElement`, `isBodyElement` etc.

test/client/fixtures/sandbox/code-instrumentation/getters-test.js
```diff
@@ -400,3 +400,10 @@ test('should not create proxy url for invalid url (GH-778)', function () {
         strictEqual(linkVal, nativeLinkVal);
     }
 });
+
+test("Should not use 'autocomplete' getter if an object isn't DOM Element (GH-942)", function () {
```
MarinaRukavitsyna: test("Should not use the 'autocomplete' getter if an object isn't a DOM element  (GH-942)", function () {

// NOTE: This code checks the PropertyAccessorsInstrumentation.attach(window).autocomplete condition

src/client/utils/dom.js
```diff
@@ -347,7 +347,7 @@ export function isHammerheadAttr (attr) {
 }
 
 export function isIframeElement (el) {
-    return getTagName(el) === 'iframe';
+    return getTagName(el) === 'iframe' && isDomElement(el);
```
LavrovArtem: May be we can to move the `isDomElement` condition into the `getTagName` function?

miherlosev: Additional check `domUtils.isDomElement` needs only in code instrumentation.
If we will add it to the `domUtils.isSomeElement` then we get proxy perfomance degradation.
@churkin Are you sure that we should do it?

georgiy-abbasov:
@testcafe-build-bot \retest

georgiy-abbasov:
@testcafe-build-bot \retest

georgiy-abbasov:
FPR

inikulin:
I guess we'll not have significant performance degradation. In most of cases we'll hit the first control path in `isDomElement` which is `instanceof` check. It's quite fast. Also, more likely `isDomElement` will be inlined by VM

test/client/fixtures/sandbox/code-instrumentation/getters-test.js
MarinaRukavitsyna: // NOTE: This code checks PropertyAccessorsInstrumentation.attach(window).autocomplete.condition

PR #966, opened 2016-11-21
Dispatched focus/blur events have a correct type (closes #947)

### LINKED ISSUE #947: Wrong dispatched focus event types
It does not pass check instanseof. Focus and blur events should has type `FocusEvent`, but now they have `Event` type 
It connected with https://github.com/DevExpress/testcafe-hammerhead/issues/941

helen-dikareva:

### DIFF
--- a/src/client/sandbox/event/simulator.js
+++ b/src/client/sandbox/event/simulator.js
@@ -537,6 +537,33 @@ export default class EventSimulator {
         return this._raiseDispatchEvent(el, ev, args);
     }
 
+    _dispatchFocusEvent (el, name) {
+        var browserWithNewEventsStyle = !browserUtils.isIE || browserUtils.version > 11;
+        var event                     = null;
+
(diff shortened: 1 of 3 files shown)


src/client/sandbox/event/simulator.js
```diff
@@ -538,17 +538,30 @@ export default class EventSimulator {
     }
 
     _dispatchEvent (el, name, shouldBubble, flag) {
-        var ev = null;
+        var event                     = null;
```
AlexanderMoskovkin: @helen-dikareva Do you like the idea to split this into two functions `_dispatchEvent` and `_dispatchFocusEvent` ? On my opinion it would be better

helen-dikareva: ok, let's

helen-dikareva:
@testcafe-build-bot \retest

helen-dikareva:
FPR

PR #974, opened 2016-11-29
fix: prevent native credentials window

churkin:
\cc @miherlosev @MarinaRukavitsyna

### DIFF
--- a/src/request-pipeline/header-transforms.js
+++ b/src/request-pipeline/header-transforms.js
@@ -92,6 +92,9 @@ var responseTransforms = {
     'x-content-security-policy-report-only': skip,
     'x-webkit-csp':                          skip,
 
+    // NOTE: Even if we are not able to be authorized, we should prevent showing the native credentials window.
+    'www-authenticate': skip,
+
     // NOTE: We perform CORS checks on our side, so we skip the related headers.
     'access-control-allow-origin': skip,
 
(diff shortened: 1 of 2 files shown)


test/server/auth-test.js
```diff
@@ -126,7 +126,8 @@ describe('Authentication', function () {
             request(proxy.openSession('http://127.0.0.1:1507/', session), function (err, res, body) {
                 expect(body).equal('Access denied');
                 expect(res.statusCode).equal(401);
-
+                // NOTE: prevent native browser credentials window.
```
MarinaRukavitsyna: // prevent showing the native credentials window

src/request-pipeline/header-transforms.js
```diff
@@ -92,6 +92,9 @@ var responseTransforms = {
     'x-content-security-policy-report-only': skip,
     'x-webkit-csp':                          skip,
 
+    // NOTE: Even if we were not able to authorize we should prevent native browser credentials window.
```
MarinaRukavitsyna: // NOTE: Even if we are not able to authorize a user, we should prevent showing the native credentials window.

MarinaRukavitsyna: // NOTE: Even if we are not able to be authorized, we should prevent showing the native credentials window.

jshor:
The comments on this are useless and there is no explanation on the PR as to what this even does. What is a "native credentials window"? Why should it not be shown? Why was this merged?

PR #976, opened 2016-11-30
The most reliable way to inspect html element

LavrovArtem:
/cc @churkin @miherlosev

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -21,9 +21,15 @@ const NATIVE_ELEMENT_PROTOTYPE_STRINGS = [
     instanceToString(Object.getPrototypeOf(nativeMethods.elementClass.prototype))
 ];
 
-const NATIVE_WINDOW_STR   = instanceToString(window);
-const NATIVE_DOCUMENT_STR = instanceToString(document);
-const IS_SVG_ELEMENT_RE   = /^\[object SVG\w+?Element]$/i;
+const NATIVE_MAP_ELEMENT_STRINGS = [
+    '[object HTMLMapElement]',
+    '[object HTMLAreaElement]'
+];
+
+const NATIVE_WINDOW_STR     = instanceToString(window);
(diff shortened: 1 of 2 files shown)


src/client/utils/dom.js
```diff
@@ -404,55 +411,55 @@ export function isIframeWithoutSrc (iframe) {
 }
 
 export function isImgElement (el) {
-    return getTagName(el) === 'img' && isDomElement(el);
+    return instanceToString(el) === '[object HTMLImageElement]';
```
miherlosev: Please add pefomance comparison using jsperf.com

LavrovArtem: https://jsperf.com/props-vs-tostring01

miherlosev: Ok.

LavrovArtem: I don't think that it has any effects to performance.

miherlosev: Ok. I checked it and and it doesn't affect perfomance.

src/client/utils/dom.js
```diff
@@ -647,15 +655,15 @@ export function isDocumentFragmentNode (el) {
...
 
 export function isTableElement (el) {
-    return getTagName(el) === 'table' && isDomElement(el);
+    return instanceToString(el) === '[object HTMLTableElement]';
 }
 
-export function isTableDataElement (el) {
-    return getTagName(el) === 'td' && isDomElement(el);
+export function isTableCellElement (el) {
+    return instanceToString(el) === NATIVE_TABLE_CELL_STR;
```
miherlosev: In standart it named as `TableDataCell`.
https://www.w3.org/TR/html401/struct/tables.html#h-11.2.6

`TableCell` is common name.
There are th - table header cell, td - table data cell.
So we should rename it to the `isTableDataCellElement`.

LavrovArtem: ok

PR #981, opened 2016-12-01
'event.key' and 'event.keyIdentifier' should be undefined if not specified

georgiy-abbasov:
\cc @AlexanderMoskovkin


georgiy-abbasov:
@testcafe-build-bot \retest

PR #1018, opened 2017-01-24
Support file:// protocol (close #908)

### LINKED ISSUE #908: Support file:// protocol
Electron application usually uses file protocol for loading pages.
Also some application use absolute paths for resources, so we can't use simple static server for the proxying pages.
E.g. vscode uses absolute paths for scripts:
https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/electron-browser/bootstrap/index.js#L136
@inikulin @churkin What are you thinking about it?

LavrovArtem:
\cc @churkin @kirovboris

### DIFF
--- a/src/processing/resources/index.js
+++ b/src/processing/resources/index.js
@@ -7,7 +7,7 @@ import * as urlUtil from '../../utils/url';
 import { encodeContent, decodeContent } from '../encoding';
 import { platform } from 'os';
 
-const IS_WIN32 = platform() === 'win32';
+const IS_WIN = platform() === 'win32';
 const DISK_RE  = /^[A-Za-z]:/;
 
 function getResourceUrlReplacer (ctx) {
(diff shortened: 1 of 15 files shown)


kirovboris:
It's necessary provide the specifying of absolute url in tests:
```js
    //Win platforms
    fixture.page('C:/index.html');

    //*nix platforms
    fixture.page('/usr/home/index.html');
```

LavrovArtem:
FPR

src/processing/resources/index.js
```diff
@@ -5,12 +5,19 @@ import scriptProcessor from './script';
 import stylesheetProcessor from './stylesheet';
 import * as urlUtil from '../../utils/url';
 import { encodeContent, decodeContent } from '../encoding';
+import { platform } from 'os';
+
+const IS_WIN32 = platform() === 'win32';
```
kirovboris: I think just `IS_WIN`

src/processing/resources/index.js
```diff
@@ -5,12 +5,19 @@ import scriptProcessor from './script';
...
+const DISK_RE  = /^[A-Za-z]:/;
 
 function getResourceUrlReplacer (ctx) {
     return function (resourceUrl, resourceType, charsetAttrValue, baseUrl) {
         if (!urlUtil.isSupportedProtocol(resourceUrl) && !urlUtil.isSpecialPage(resourceUrl))
             return resourceUrl;
 
+        if (IS_WIN32 && ctx.dest.protocol === 'file:' && DISK_RE.test(resourceUrl))
```
kirovboris: What is about *nix platforms? Can we at least add an unit test with hadling paths like `/usr/home/index.html`?

LavrovArtem: `/usr/home/index.html` is already relative path which can be resolved via the `url.resolve` function

kirovboris: I would add test anyway.

LavrovArtem: Sure)

PR #1020, opened 2017-01-26
fix `Authorization Header is removed even if it should not` (close #1016)

### LINKED ISSUE #1016: Authorization Header is removed even if it should not
After spending some time digging into this topic I think the current implementation of transformation of Authorization headers `header-transforms.js` does not cover all real life uses cases, even if it seems to follow specification (https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials).

What I observed:

So, this hammerhead lib is used within TestCafe to proxy http requests coming from the page/application under test. In our case we use Authorization header not for HTTP Authentication, but to pass an OAuth2 token to our API backend. The token is added from within the JS application and not automatically by the browser after the user signed in (as it happens for HTTP auth). By intend we do not create XHR requests using `withCredentials = true`.

What happens now is (which is actual different from default browser behaviour):

When the API backend request is sent from our application under test (by TestCafe), the request is intercepted by hammerhead and because withCredentials=false the Authorization header is removed, which leads to unauthenticated request to our backend and in turn to failing tests. IMHO, this is not correct, as this specific auth header is not like the ones which are automatically send by the browser when accessing sites that required http authentication.

LavrovArtem:
/cc @churkin @miherlosev

### DIFF
--- a/src/client/sandbox/xhr.js
+++ b/src/client/sandbox/xhr.js
@@ -2,6 +2,7 @@ import SandboxBase from './base';
 import nativeMethods from './native-methods';
 import { getProxyUrl } from '../utils/url';
 import XHR_HEADERS from '../../request-pipeline/xhr/headers';
+import AUTHORIZATION from '../../request-pipeline/xhr/authorization';
 import { getOrigin } from '../utils/destination-location';
 import reEscape from '../../utils/regexp-escape';
 
@@ -150,6 +151,13 @@ export default class XhrSandbox extends SandboxBase {
             var headers = nativeMethods.xhrGetAllResponseHeaders.call(this);
 
             return headers ? headers.replace(REMOVE_SET_COOKIE_HH_HEADER, '') : headers;
+        };
+
+        xmlHttpRequestProto.setRequestHeader = function (header, value) {
+            if (typeof header === 'string' && AUTHORIZATION.headers.indexOf(header.toLowerCase()) !== -1)
+                value = AUTHORIZATION.valuePrefix + value;
+
+            return nativeMethods.xhrSetRequestHeader.call(this, header, value);
         };
     }
 }
--- a/src/request-pipeline/xhr/authorization.js
+++ b/src/request-pipeline/xhr/authorization.js
@@ -1,0 +1,15 @@
+// -------------------------------------------------------------
+// WARNING: this file is used by both the client and the server.
+// Do not use any browser or node-specific API!
+// -------------------------------------------------------------
+/* eslint hammerhead/proto-methods: 2 */
+
+export default {
+    valuePrefix: 'hammerhead|prefix|by-client',
+    headers:     [
+        'authorization',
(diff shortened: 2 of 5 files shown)


src/client/sandbox/xhr.js
```diff
@@ -151,5 +152,12 @@ export default class XhrSandbox extends SandboxBase {
 
             return headers ? headers.replace(REMOVE_SET_COOKIE_HH_HEADER, '') : headers;
         };
+
+        xmlHttpRequestProto.setRequestHeader = function (header, value) {
+            if (typeof header === 'string' && AUTHORIZATION.headers.indexOf(header.toLowerCase()) !== -1)
+                value = AUTHORIZATION.valuePrefix + value;
+
+            nativeMethods.xhrSetRequestHeader.call(this, header, value);
```
churkin: return

src/request-pipeline/xhr/authorization.js
```diff
@@ -0,0 +1,15 @@
+// -------------------------------------------------------------
+// WARNING: this file is used by both the client and the server.
+// Do not use any browser or node-specific API!
+// -------------------------------------------------------------
+/* eslint hammerhead/proto-methods: 2 */
+
+export default {
+    valuePrefix: 'hammerhead-prefix-by-client',
```
churkin: `hammerhead-prefix-by-client` --> `hammerhead|prefix|by-client`

test/client/fixtures/sandbox/xhr-test.js
```diff
@@ -165,3 +166,28 @@ asyncTest('set cookie from a header of the XMLHttpRequest response (GH-905)', fu
...
+    xhr.open('GET', '/echo-request-headers/', true);
+    xhr.setRequestHeader(AUTHORIZATION.headers[0], '123');
+    xhr.setRequestHeader(AUTHORIZATION.headers[1], '123');
+    xhr.setRequestHeader('x-header1', '456');
+    xhr.setRequestHeader('x-header2', '789');
+    xhr.addEventListener('readystatechange', function () {
+        if (this.readyState === this.DONE) {
+            var headers = JSON.parse(this.responseText);
+
+            Object.keys(headers).forEach(function (header) {
```
churkin: Just check 4 headers, without `foreach` function

test/server/proxy-test.js
```diff
@@ -1525,5 +1526,34 @@ describe('Proxy', function () {
                 done();
             });
         });
+
+        it('Should pass authorization headers which are defined by client for cross-domain request without credentials (GH-1016)', function (done) {
+            var options = {
+                url:     proxy.openSession('http://127.0.0.1:2002/echo-headers-with-credentials', session),
+                headers: AUTHORIZATION.headers.reduce(function (obj, header, index) {
```
churkin: Use more simple code - without `reduce`, set all headers manualy

PR #1027, opened 2017-01-30
'setSelectionRange' method fixed (closes #1025)

### LINKED ISSUE #1025: Input type should be changed while setting selection range in Firefox
https://html.spec.whatwg.org/multipage/forms.html#do-not-apply
This [rule](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/event/selection.js#L107) should  also include Firefox browser since 51 version, cause it doesn't allow to set selectionRange originally, which leads to this [issue](https://github.com/DevExpress/testcafe/issues/1187)

georgiy-abbasov:
\cc @churkin, @AlexanderMoskovkin
I think, there is no need for regression test here, cause it was placed in the testcafe tests, when we faced with it first time - [B254340 ](https://github.com/DevExpress/testcafe/blob/378c6c71ac0df8fd9eae8af21040b2e7f1ad0571/test/client/fixtures/automation/regression-test.js#L594). It's green, cause Sauce Labs doesn't install FF 51yet.

### DIFF
--- a/src/client/sandbox/event/selection.js
+++ b/src/client/sandbox/event/selection.js
@@ -104,7 +104,8 @@ export default class Selection {
     }
 
     static _needChangeInputType (el) {
-        return domUtils.isInputElement(el) && browserUtils.isWebKit && /^(number|email)$/.test(el.type);
+        return (browserUtils.isWebKit || browserUtils.isFirefox && browserUtils.version > 50)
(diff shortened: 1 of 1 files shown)


georgiy-abbasov:
Firefox 51 needs for another approach to solve issue with typing, i need some time to investigate it.

PR #1030, opened 2017-02-01
'setSelectionRange' method fixed (closes #1025)

### LINKED ISSUE #1025: Input type should be changed while setting selection range in Firefox

georgiy-abbasov:
\cc @churkin, @AlexanderMoskovkin
With this fix testcafe qunit tests passes in Firefox 51 locally and in Firefox beta on Sauce Labs. Currently i'm checking functional tests with this set.

### DIFF
--- a/src/client/sandbox/event/selection.js
+++ b/src/client/sandbox/event/selection.js
@@ -3,6 +3,7 @@ import Listeners from './listeners';
 import nativeMethods from '../native-methods';
 import * as browserUtils from '../../utils/browser';
 import * as domUtils from '../../utils/dom';
+import INTERNAL_PROPS from '../../../processing/dom/internal-properties';
 
 export default class Selection {
     constructor (eventSandbox) {
@@ -26,12 +27,22 @@ export default class Selection {
             var isElementActive    = false;
 
             var selectionSetter = () => {
-                var changeType = Selection._needChangeInputType(el);
-                var savedType  = el.type;
+                var changeType        = Selection._needChangeInputType(el);
+                var useInternalSelection = Selection._needForInternalSelection();
+                var savedType         = el.type;
                 var res;
 
-                if (changeType)
+                if (changeType) {
                     el.type = 'text';
+
+                    if (useInternalSelection) {
+                        el[INTERNAL_PROPS.selectionProperty] = {
+                            selectionStart:     selectionStart,
+                            selectionEnd:       selectionEnd,
+                            selectionDirection: selection
+                        };
+                    }
+                }
 
                 // NOTE: In MSEdge, an error occurs  when the setSelectionRange method is called for an input with
                 // 'display = none' and selectionStart !== selectionEnd in other IEs, the error doesn't occur, but
@@ -44,12 +55,19 @@ export default class Selection {
                 }
 
                 if (changeType) {
+                    if (useInternalSelection) {
+                        el[INTERNAL_PROPS.selectionProperty].selectionStart     = el.selectionStart;
+                        el[INTERNAL_PROPS.selectionProperty].selectionEnd       = el.selectionEnd;
+                        el[INTERNAL_PROPS.selectionProperty].selectionDirection = el.selectionDirection;
+                    }
+
                     el.type = savedType;
                     // HACK: (A problem with input type = 'number' after Chrome is updated to v.33.0.1750.117 and
                     // in Firefox 29.0.  T101195) To set right selection: if the input type is 'number' or 'email',
                     // we need to change the type to text, and then restore it after setting selection.(B254340).
                     // However, the type is changed asynchronously in this case. To force type changing,we need to
                     // call blur, Then raise the focus event to make the element active.
+
                     if (isElementActive) {
                         selection.focusBlurSandbox.blur(el, null, true);
                         selection.focusBlurSandbox.focus(el, null, true);
@@ -104,7 +122,15 @@ export default class Selection {
     }
 
     static _needChangeInputType (el) {
-        return domUtils.isInputElement(el) && browserUtils.isWebKit && /^(number|email)$/.test(el.type);
+        return (browserUtils.isWebKit || Selection._needForInternalSelection())
+               && domUtils.isInputElement(el) && /^(number|email)$/.test(el.type);
+    }
+
+    // NOTE: input elements in Firefox since version 51 are needing
+    // for alternative selection properties, because original are
+    // clearing while input type changes to url, email etc
+    static _needForInternalSelection () {
+        return browserUtils.isFirefox && browserUtils.version > 50;
     }
 
     setSelection (el, start, end, direction) {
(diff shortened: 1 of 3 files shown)


src/client/sandbox/event/selection.js
```diff
@@ -26,13 +27,23 @@ export default class Selection {
...
 
-                if (changeType)
+                if (changeType) {
                     el.type = 'text';
 
+                    if (useInternalSelection) {
+                        el[INTERNAL_PROPS.selectionProperty] = {
+                            selectionStart:     selectionStart,
+                            selectionEnd:       selectionEnd,
+                            selectionDirection: selection
```
AlexanderMoskovkin: remove this internal property setter

src/client/sandbox/event/selection.js
```diff
@@ -117,33 +143,46 @@ export default class Selection {
...
             el.type = 'text';
         }
 
-        selection = {
-            start:     el.selectionStart,
-            end:       el.selectionEnd,
-            direction: el.selectionDirection
-        };
+        if (useInternalSelection && changeType && el[INTERNAL_PROPS.selectionProperty]) {
+            selection = {
```
AlexanderMoskovkin: no need to change el.type when this conditions are true

src/client/sandbox/event/selection.js
```diff
@@ -117,33 +143,46 @@ export default class Selection {
...
+                start:     el[INTERNAL_PROPS.selectionProperty].selectionStart,
+                end:       el[INTERNAL_PROPS.selectionProperty].selectionEnd,
+                direction: el[INTERNAL_PROPS.selectionProperty].selectionDirection
+            };
+        }
+        else {
+            selection = {
+                start:     el.selectionStart,
```
AlexanderMoskovkin: need to return default values or real

src/client/sandbox/event/selection.js
```diff
@@ -117,33 +143,46 @@ export default class Selection {
...
+        }
+
 
         if (changeType) {
             el.type = savedType;
 
-            if (isElementActive)
+            // NOTE: In Firefox since version 51 element lost focus if we try to get its
+            // selection properties after changing its type and browser windows isn't in focus
+            if (isElementActive || domUtils.getActiveElement(domUtils.findDocument(el)) !== el)
```
AlexanderMoskovkin: the second condition breaks behavior

src/client/sandbox/event/selection.js
```diff
@@ -104,7 +122,15 @@ export default class Selection {
...
 
     static _needChangeInputType (el) {
-        return domUtils.isInputElement(el) && browserUtils.isWebKit && /^(number|email)$/.test(el.type);
+        return (browserUtils.isWebKit || Selection._needForInternalSelection())
+               && domUtils.isInputElement(el) && /^(number|email)$/.test(el.type);
+    }
+
+    // NOTE: input elements in Firefox since version 51 are needing
+    // for alternative selection properties, because original are
+    // clearing while input type changes to url, email etc
```
AlexanderMoskovkin: Something like: `we need to store element state because it's cleared when it's type is changed`.

src/client/sandbox/event/selection.js
AlexanderMoskovkin: tests required

src/client/sandbox/event/selection.js
```diff
@@ -117,27 +135,30 @@ export default class Selection {
...
             el.type = 'text';
         }
 
+        var internalSelection = el[INTERNAL_PROPS.selectionProperty];
+
         selection = {
-            start:     el.selectionStart,
-            end:       el.selectionEnd,
-            direction: el.selectionDirection
+            start:     internalSelection ? internalSelection.selectionStart : el.selectionStart,
```
AlexanderMoskovkin: if we didn't set internal selection to this element before it will return `{ null, null, null }`, but should some default values. I told you before about it

src/client/sandbox/event/selection.js
```diff
@@ -117,27 +135,30 @@ export default class Selection {
...
+        var savedType            = el.type;
+        var selection            = null;
 
         // HACK: (A problem with input type = ‘number’ after Chrome is updated to v.33.0.1750.117 and in
         // Firefox 29.0. T101195) To get selection, if the input type is  'number' or 'email', we need to change
         // the type to text (B254340). However, the type is changed asynchronously in this case. To force type changing,
         // we need to call blur.Then call focus to make the element active.
         if (changeType) {
-            if (isElementActive)
+            if (!useInternalSelection && isElementActive)
```
AlexanderMoskovkin:
```js
if (!useInternalSelection && isElementActive)
    this.focusBlurSandbox.blur(el, null, true);
```
but
```js
if (isElementActive)
    this.focusBlurSandbox.focus(el, null, true);

```
??

src/client/sandbox/event/selection.js
```diff
@@ -26,8 +27,9 @@ export default class Selection {
             var isElementActive    = false;
 
             var selectionSetter = () => {
-                var changeType = Selection._needChangeInputType(el);
-                var savedType  = el.type;
+                var changeType           = Selection._needChangeInputType(el);
+                var useInternalSelection = Selection._needForInternalSelection();
```
AlexanderMoskovkin: with this code if `input.type` is `text` and it's Firefox v51, the `useInternalSelection` var is true. But I suppose we should not use internal selection in this case...

src/client/sandbox/event/selection.js
AlexanderMoskovkin: And we need a test for this case

test/client/fixtures/sandbox/selection-test.js
```diff
@@ -0,0 +1,26 @@
...
+
+test('Set and get selection on input with "email" type', function () {
+    var input = document.createElement('input');
+
+    input.type  = 'email';
+    input.value = inputTestValue;
+
+    document.body.appendChild(input);
+
+    input.setSelectionRange(0, inputTestValue.length, 'backward');
```
AlexanderMoskovkin: use some custom selection in this test (not form start to end, but something like 1..3)

src/client/sandbox/event/selection.js
georgiy-abbasov: >> if we didn't set internal selection to this element before it will return { null, null, null }

It returns { input.valueLength, input.valueLength, 'forward' }, cause type is changed to text, Firefox set these values as selection props.
```html
<input id="number" type="email" value="test@host.net">
<script>
    var input = document.querySelector('input');

    console.log(input.selectionStart); // null
    console.log(input.selectionEnd); // null
    console.log(input.selectionDirection); // null

    input.type = 'text';

    console.log(input.selectionStart); // 13
    console.log(input.selectionEnd); // 13
    console.log(input.selectionDirection); // forward
</script>
```

AlexanderMoskovkin: ok

src/client/sandbox/event/selection.js
georgiy-abbasov: I have faced with situation while ran qunit tests in testcafe - firefox doesn't focus on the element again while executes (after blur is executed)
```js
if (isElementActive)
    this.focusBlurSandbox.focus(el, null, true);
```

georgiy-abbasov:
TestCafe functional test is blinking, but it's not related to this commit, so FPR

AlexanderMoskovkin:
@churkin Please review it again, it was reworked

AlexanderMoskovkin:
@testcafe-build-bot \retest

AlexanderMoskovkin:
@testcafe-build-bot \retest

test/client/fixtures/sandbox/selection-test.js
```diff
@@ -0,0 +1,51 @@
...
+        start:     2,
+        end:       9,
+        direction: 'backward'
+    };
+
+    input.setSelectionRange(testSelection.start, testSelection.end, testSelection.direction);
+
+    var selection = selectionSandbox.getSelection(input);
+
+    equal(selection.start, testSelection.start);
```
churkin: We are using `strictEqual` anywhere

test/client/fixtures/sandbox/selection-test.js
```diff
@@ -0,0 +1,51 @@
+var selectionSandbox = hammerhead.eventSandbox.selection;
+var browserUtils     = hammerhead.utils.browser;
+
+var inputTestValue  = 'test@host.net';
+var isIE            = browserUtils.isIE;
+var isMobileBrowser = browserUtils.isIOS || browserUtils.isAndroid;
+
+var createTestInput = function () {
+    var input = $('<input type="email">').appendTo(document.body)[0];
```
churkin: We try to not use `jquery` in our `qunit` tests

georgiy-abbasov: It's impossible to change input type in IE9, after object is created.

churkin: You can use `.innertHTML = "<input type="email">"` for example

AlexanderMoskovkin: what? jquery is just a wrapper around native js. what's the problem to do the same as jquery?

georgiy-abbasov:
@testcafe-build-bot \retest

PR #1039, opened 2017-02-10
Fix require() on Electron clients (closes #1038)

### LINKED ISSUE #1038: Improve Electron sandboxing when proxying a page with file:// protocol
1. `require('./some-relative-path')` works on the client side only if Electron was started with a `file://` URL. This means that requiring relative paths do not work under `testcafe-hammerhead`. We should get correct `__dirname` for the page, and transform the required module path to an absolute one.

2. `require('module-name')` works only if the page served through `file://` protocol. Likely we have to get the `__dirname` of the original page and then follow `require` resolving rules: https://nodejs.org/api/modules.html#modules_all_together

3. Electron supports `__filename` and  `__dirname` inside HTML-pages on the client side, but only if the page is served through `file://` protocol. Similar to the `node`, these globals should contain absolute paths to the page and to the directory of the page. If the page was loaded through `http://`, these globals points to some internal Electron script. So we have to supply the correct values to the proxied page.

AndreyBelym:
\cc @churkin @LavrovArtem @kirovboris

### DIFF
--- a/src/client/sandbox/electron.js
+++ b/src/client/sandbox/electron.js
@@ -12,7 +12,7 @@ export default class ElectronSandbox extends SandboxBase {
         };
     }
 
-    static _fixModuleName (window) {
+    static _overrideElectronModulePaths (window) {
         // NOTE: Need this to avoid Webmake require
         /* eslint-disable no-eval */
         var electronRequire = eval('require');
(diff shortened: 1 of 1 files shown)


src/client/sandbox/electron.js
```diff
@@ -30,6 +53,8 @@ export default class ElectronSandbox extends SandboxBase {
             vm.runInContext      = ElectronSandbox._createFnWrapper(vm, nativeMethods.runInContext);
             vm.runInNewContext   = ElectronSandbox._createFnWrapper(vm, nativeMethods.runInNewContext);
             vm.runInThisContext  = ElectronSandbox._createFnWrapper(vm, nativeMethods.runInThisContext);
+
+            ElectronSandbox._fixModuleName(window);
```
kirovboris: I think `_fixModuleName` is bad name for method. We should use `_overrideElectronPaths` or sth like that. @LavrovArtem @churkin What do You think?

churkin: Agree

PR #1056, opened 2017-03-02
fix `We need to use the instanceToString function from dom utils instead of toString function` (close #1055)

### LINKED ISSUE #1055: We need to use the instanceToString function from dom utils instead of toString function
In [this](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/style.js#L36) function occurred the error: "Maximum call stack size exceeded"

LavrovArtem:
/cc @churkin @miherlosev

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -96,7 +96,7 @@ function hasClassFallback (el, className) {
     return preparedElementClassName.indexOf(className) !== -1;
 }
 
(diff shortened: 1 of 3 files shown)


LavrovArtem:
@testcafe-build-bot \retest

LavrovArtem:
@testcafe-build-bot \retest

PR #1061, opened 2017-03-15
Get activeElement from Shadow DOM

AndreyBelym:
\cc @churkin @miherlosev

### DIFF
--- a/src/client/utils/dom.js
+++ b/src/client/utils/dom.js
@@ -110,7 +110,18 @@ export function getActiveElement (currentDocument) {
     // https://github.com/DevExpress/testcafe-hammerhead/issues/768
     var doc = currentDocument || document;
 
-    return isDomElement(doc.activeElement) ? doc.activeElement : doc.body;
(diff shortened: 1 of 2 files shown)


src/client/utils/dom.js
```diff
@@ -110,7 +110,18 @@ export function getActiveElement (currentDocument) {
     // https://github.com/DevExpress/testcafe-hammerhead/issues/768
     var doc = currentDocument || document;
 
-    return isDomElement(doc.activeElement) ? doc.activeElement : doc.body;
+    var el = isDomElement(doc.activeElement) ? doc.activeElement : doc.body;
+
+    while (el && el.shadowRoot) {
```
miherlosev: Is it possible to create nested shadowRoots?
Can active element be not `isDomElement`?

If it is true, then add tests for following cases.

AlexanderMoskovkin:
Tests are ok (besides a blinking test), so I'll merge this

PR #1063, opened 2017-03-16
fix 'Should not raise file download if resource is fetched by setting script src' (close #1062)

### LINKED ISSUE #1062: Should not raise file download if resource is fetched by setting script src
We have a some script resource with file download response headers (`content-disposition:attachment; filename="f.txt"`).

If we fetch this resource via user action (for example, by click on link) then browser performs download.
If we do it by setting script src then browser process resource normally.

miherlosev:
@churkin


miherlosev:
@testcafe-build-bot \retest

PR #1067, opened 2017-03-17
Implemented 'focusIn,focusOut' events emulation (close  #1044)

### LINKED ISSUE #1044: Emulate focusin and focusout events in background
When a window is in background, browser stops raising focus events even after `.focus()` call. Currently in this case we manually raise `focus` and `blur` events. We need to do  the same for `focusin` and `focusout` events.

miherlosev:


miherlosev:
@testcafe-build-bot \retest

miherlosev:
FPR
@helen-dikareva @AndreyBelym

src/client/sandbox/event/focus-blur.js
```diff
@@ -64,85 +63,117 @@ export default class FocusBlurSandbox extends SandboxBase {
...
-            FocusBlurSandbox._getNativeMeth(el, type).call(el);
-
-            if (preventScrolling)
-                FocusBlurSandbox._restoreElementScroll(this.window, windowScroll);
-
-            if (browserUtils.isIE && nonScrollableParents.length) {
-                for (var j = 0, len = nonScrollableParents.length; j < len; j++)
-                    FocusBlurSandbox._restoreElementScroll(nonScrollableParents[j], nonScrollableParentsScrollValues[j]);
-            }
+            // NOTE: To guarantee that all focus/blur events are raised, we need to raise them manually.
```
helen-dikareva: This comment is not related with scrolling. 
I don't really like that `_ensureWindowScroll` call real `focus`.

src/client/sandbox/event/focus-blur.js
```diff
@@ -64,85 +63,117 @@ export default class FocusBlurSandbox extends SandboxBase {
...
+            }
+        }
+    }
+
+    _getElementNonScrollableParentsScrollState (el) {
+        var scrollState    = [];
+        var elementParents = domUtils.getParents(el);
+        var elementParent  = null;
+
+        for (var i = 0; i < elementParents.length; i++) {
```
helen-dikareva: You can use `for ... of` in circles

src/client/sandbox/event/focus-blur.js
```diff
@@ -172,7 +203,25 @@ export default class FocusBlurSandbox extends SandboxBase {
...
+        else if (focusInOutEventType === 'focusout')
+            return 'blur';
+
+        return null;
+    }
+
+    static getFocusInOutEventType (focusBlurEventType) {
+        if (focusBlurEventType === 'focus')
+            return 'focusin';
+        else if (focusBlurEventType === 'blur')
```
helen-dikareva: `getFocusBlurEventType`
`getFocusInOutEventType`
These two methods look very similar and confusing

test/client/fixtures/sandbox/event/focus-blur-change-test.js
```diff
@@ -879,6 +879,71 @@ asyncTest("focus() must not scroll to the element if 'preventScrolling' argument
...
+        var nativeEventLog = '';
+        var input          = document.createElement('input');
+        var nativeInput    = nativeMethods.createElement.call(document, 'input');
+
+        document.body.appendChild(input);
+        nativeInput.focus = nativeMethods.focus;
+        nativeInput.blur  = nativeMethods.blur;
+        nativeMethods.appendChild.call(document.body, nativeInput);
+
+        var focusHandler                    = function () {
```
helen-dikareva: It can be one handler:
```js
function (e) {
    eventLog +=  e.type + '|';
};
```

miherlosev: Ok

src/client/sandbox/event/focus-blur.js
miherlosev: > redundant else

Ok

src/client/sandbox/event/focus-blur.js
miherlosev: > This comment is not related with scrolling. 

This comment is placed on first line in brackets in which  we perform `_raiseEvent` calls.
https://github.com/miherlosev/testcafe-hammerhead/blob/74b2da15b2634991b3f66d762b90946d5d1e4ec6/src/client/sandbox/event/focus-blur.js#L175

src/client/sandbox/event/focus-blur.js
miherlosev: Ok

src/client/sandbox/event/focus-blur.js
miherlosev: >  getFocusBlurEventType
>  getFocusInOutEventType
>  These two methods look very similar and confusing

I need to perform two way convesion:
`blur <-> focusout`
`focus <-> focusin`

Propose a better way by your opinion

src/client/sandbox/event/focus-blur.js
miherlosev: >  I don't really like that _ensureWindowScroll call real focus

See the current code from
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/event/focus-blur.js#L104
to 
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/event/focus-blur.js#L145
I don't change an old logic
I just split the large code part on simple logical pars

miherlosev:
FPR

miherlosev:
FPR @helen-dikareva

src/client/sandbox/event/focus-blur.js
```diff
@@ -181,7 +148,36 @@ export default class FocusBlurSandbox extends SandboxBase {
...
+            let windowScroll                           = null;
+            let elementNonScrollableParentsScrollState = null;
+
+            if (preventScrolling)
+                windowScroll = styleUtils.getElementScroll(this.window);
+
+            if (browserUtils.isIE)
+                elementNonScrollableParentsScrollState = this._getElementNonScrollableParentsScrollState(el);
+
+            if (this._shouldUseLabelHtmlForElement(el, type)) {
```
helen-dikareva: I think it was better with ` this._ensureWindowScroll(el, preventScrolling, type, callback);`. Only without `callback`
What about
```js
el =  this._ensureWindowScroll(el, preventScrolling, type);

if (!el) {
    callback();
    return;
}

File: src/client/sandbox/event/focus-blur.js [hunk 6]
```diff
@@ -181,7 +148,36 @@ export default class FocusBlurSandbox extends SandboxBase {
 
         if (el[type]) {
             // NOTE: To guarantee that all focus/blur events are raised, we need to raise them manually.
-            this._ensureWindowScroll(el, preventScrolling, type, callback);
+            let windowScroll                           = null;
+            let elementNonScrollableParentsScrollState = null;
+
+            if (preventScrolling)
```
helen-dikareva: May be we can extract in method storing scroll state `saveScrollStateIfNecessary` and for storing valuesuse` this`.
And after calling `focus` use method` restoreScrollStateIfNecessary`

churkin: Agree

src/client/sandbox/event/focus-blur.js
miherlosev: I don't like this way.
It is not clear why `_ensureWindowScroll` method retruns `el` or `null`.

helen-dikareva: I'm sorry `_ensureLabelHtmlForElement` of course

miherlosev: It is a bad solution too.
You should returns two fields: `el` and flag that equals `it is  label with htmlFor attribute`.
Then we will return is only if it is label with htmlFor attribute and element is null.
So, the current way is simpler.

helen-dikareva: We don't need any flags. We should just return `el`. What do you mean?

miherlosev:
@testcafe-build-bot \retest

miherlosev:
@testcafe-build-bot \retest

miherlosev:
Tests are green.
@helen-dikareva merge this

PR #1077, opened 2017-03-31
fix focus blur event order test

miherlosev:


miherlosev:
I set the fixed browser version.
Elements doesn't raise focusin, focusout events for current latest firefox for Mac on SauceLabs.
Waiting for 52.0.2 or later.

miherlosev:
FPR
@churkin @LavrovArtem

PR #1100, opened 2017-04-25
fix 'on calling without parameters we should return the native resultof 'fetch' function instead of window.Promise.reject' (close #1099)

### LINKED ISSUE #1099: On calling without parameters we should return the native result of 'fetch' function instead of window.Promise.reject
It blocks the `core-js` library loading and, accordingly, `angular2` applications.

miherlosev:
@churkin @LavrovArtem

### DIFF
--- a/src/client/sandbox/fetch.js
+++ b/src/client/sandbox/fetch.js
@@ -127,6 +127,9 @@ export default class FetchSandbox extends SandboxBase {
             window.Request.prototype = nativeMethods.Request.prototype;
 
             window.fetch = function (...args) {
+                if (!args.length)
+                    return nativeMethods.fetch.apply(this, args);
+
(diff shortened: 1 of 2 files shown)


src/client/sandbox/fetch.js
```diff
@@ -127,6 +127,9 @@ export default class FetchSandbox extends SandboxBase {
             window.Request.prototype = nativeMethods.Request.prototype;
 
             window.fetch = function (...args) {
+                if (!args.length)
+                    return nativeMethods.fetch.apply(this, args);
```
LavrovArtem: just `return nativeMethods.fetch.call(this);`

miherlosev: Ok

miherlosev:
FPR

PR #1113, opened 2017-05-05
fix `EventSource requests are broken under proxy` (close #1106)

### LINKED ISSUE #1106: EventSource requests are broken under proxy
Based on the https://github.com/DevExpress/testcafe/issues/1458.

1) Open the Network tab in the browser devtools.
2) Open the https://testcafe-spike.qantasassure.com/health-insurance/quote page under playground
3) Fill all fields and press the `Continue` button
4) Press the `Continue As A Guest button`
5) See in the Network tab that requests are broken. It seems something went wrong during server-side processing

LavrovArtem:
/cc @churkin @miherlosev

### DIFF
--- a/src/client/sandbox/node/window.js
+++ b/src/client/sandbox/node/window.js
@@ -7,7 +7,7 @@ import { processScript } from '../../../processing/script';
 import styleProcessor from '../../../processing/style';
 import * as destLocation from '../../utils/destination-location';
 import { processHtml } from '../../utils/html';
-import { isSubDomain, parseUrl, getProxyUrl, convertToProxyUrl } from '../../utils/url';
+import { isSubDomain, parseUrl, getProxyUrl, convertToProxyUrl, stringifyResourceType } from '../../utils/url';
 import { isFirefox } from '../../utils/browser';
 import { isCrossDomainWindows, isImgElement, isBlob } from '../../utils/dom';
 import INTERNAL_ATTRS from '../../../processing/dom/internal-attributes';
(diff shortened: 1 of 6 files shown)


src/client/sandbox/node/window.js
```diff
@@ -166,8 +166,12 @@ export default class WindowSandbox extends SandboxBase {
         }
 
         if (window.EventSource) {
-            window.EventSource           = url => new nativeMethods.EventSource(getProxyUrl(url));
-            window.EventSource.prototype = nativeMethods.EventSource.prototype;
+            window.EventSource            = url => new nativeMethods.EventSource(
```
miherlosev: `window.EventSource` has two arguments in constructor - https://html.spec.whatwg.org/multipage/comms.html#the-eventsource-interface

test/client/fixtures/sandbox/node/classes-test.js
```diff
@@ -56,6 +56,26 @@ if (window.EventSource) {
...
         ok(eventSource instanceof EventSource);
     });
+
+    test('should have static constants (GH-1106)', function () {
+        ok(EventSource.hasOwnProperty('CONNECTING'));
+        ok(EventSource.hasOwnProperty('OPEN'));
+        ok(EventSource.hasOwnProperty('CLOSED'));
+    });
+
+    test('should process url to proxy with special flag (GH-1106)', function () {
```
miherlosev: Add tests for the cases when EventSource is closed:
-  from client - `eventSource.close()`
- from server - status code does not equal `200`

LavrovArtem:
FPR

test/client/fixtures/sandbox/node/classes-test.js
```diff
@@ -56,6 +56,59 @@ if (window.EventSource) {
...
```
miherlosev: Add [CORS handling](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/request-pipeline/index.js#L50) and replace proxied url for `origin` property for EventSource requests.
Example:
```js
var eventSource= new EventSource("http://cross-domain.com", {
  withCredentials: true
});
eventSource.addEventListener('message', function(e) {
  if (e.origin === 'http://origin-site-url') ...;
});
```

Also let's add more real life tests:
- `EventSource` for same domain request.
- `EventSource` for cross domain request. 
- checking `origin` property

LavrovArtem: It sounds like an unrelated problem. May be we need to create a separate issue.

LavrovArtem: What do you think?

miherlosev: Ok. I've created the separate issue - https://github.com/DevExpress/testcafe-hammerhead/issues/1114

Issue #350, opened 2016-01-11
Implement basic XHTML support


miherlosev:
![1](https://cloud.githubusercontent.com/assets/4133518/12230660/f6a1aeb6-b862-11e5-9f75-3efbe20caecc.png)

Presumably, problem grows from non closed our `meta` tag (`<meta xmlns="http://www.w3.org/1999/xhtml" class="charset-hammerhead-shadow-ui" charset="iso-8859-1">`).

inikulin:
This is issue is caused by the fact that page has `XHTML`

miherlosev:
I'm working on this

miherlosev:
To fix bug we need to change our `meta` tags (`X-UA-Compatible` and `charset`) to selfclosed.

inikulin:
@miherlosev Just thought it will not work if we will do it just for our tags. Since we serialize entire page we break it anyway serializing incorrectly user's `<meta>` tags. For now we can use the following solution:
1. Determine if page is an XML page (by analyzing doctype's publidId, details here: https://html.spec.whatwg.org/multipage/xhtml.html#parsing-xhtml-documents)
2. Use https://www.npmjs.com/package/sax for parsing and build XML-tree using it (like jsdom does:  https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/browser/htmltodom.js#L106 but with the proper XML entities, like ProcessingInstruction and stuff).
3. Implement a XML serializer (using https://github.com/jindw/xmldom/blob/master/dom.js#L913 as starting point).
4. Implement PageProcessor adapter for XML tree

2 and 3 can be refactored to the separate package.

husolukas:
is this issue solved? i have version 0.20.4 but i need to perform test on xhtml page

willscripted:
> is this issue solved?

This appears to still be an issue. 

Anyone have a workaround? Is there a way to bypass hammerhead for urls, similar to the `bypass-proxy` setting?

We are seeing valid xml read in by hammerhead and returned invalid.

```
.
.
  <link [...] /><link [...] /><link [...] />
.
.
```
to
```
.
.
  <link [...]><link [...]><link [...]>
```
Notice that the tags are no longer closed after hammerhead processing.

Farfurix:
@willscripted 

Hello,

We have plans to fix this issue, but I cannot provide you with any estimates on when it will be done.

rhymiz:
[...]

PR #1135, opened 2017-05-18
Override 'DOMParser.parseFromString' function (close #1133)

### LINKED ISSUE #1133: We should override the 'DOMParser.DOMParser.parseFromString'
https://developer.mozilla.org/en-US/docs/Web/API/DOMParser

LavrovArtem:
/cc @churkin @miherlosev

### DIFF
--- a/src/client/sandbox/native-methods.js
+++ b/src/client/sandbox/native-methods.js
@@ -173,6 +173,10 @@ class NativeMethods {
         // Object
         this.objectToString = win.Object.prototype.toString;
 
+        // DOMParser
+        if (win.DOMParser)
+            this.DOMParserParseFromString = win.DOMParser.prototype.parseFromString;
+
         this.refreshClasses(win);
     }
 
--- a/src/client/sandbox/node/window.js
+++ b/src/client/sandbox/node/window.js
@@ -326,6 +326,15 @@ export default class WindowSandbox extends SandboxBase {
             };
         }
 
+        if (window.DOMParser) {
+            window.DOMParser.prototype.parseFromString = function (...args) {
+                if (args.length > 1 && typeof args[0] === 'string' && args[1] === 'text/html')
+                    args[0] = processHtml(args[0]);
+
+                return nativeMethods.DOMParserParseFromString.apply(this, args);
+            };
+        }
+
         // NOTE: stab for ie9 and ie10 (GH-801)
         if (window.XDomainRequest)
             window.XDomainRequest = window.XMLHttpRequest;
--- a/src/client/sandbox/node/window.js
+++ b/src/client/sandbox/node/window.js
@@ -8,7 +8,7 @@ import styleProcessor from '../../../processing/style';
 import * as destLocation from '../../utils/destination-location';
 import { processHtml } from '../../utils/html';
 import { isSubDomain, parseUrl, getProxyUrl, convertToProxyUrl, stringifyResourceType } from '../../utils/url';
-import { isFirefox } from '../../utils/browser';
+import { isFirefox, isIE9 } from '../../utils/browser';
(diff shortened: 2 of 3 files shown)


src/client/sandbox/node/window.js
```diff
@@ -326,6 +326,15 @@ export default class WindowSandbox extends SandboxBase {
             };
         }
 
+        if (window.DOMParser) {
+            window.DOMParser.prototype.parseFromString = function (...args) {
+                if (args.length > 1 && typeof args[0] === 'string')
```
miherlosev: `args.length` is an uncessary check. Drop this.

Also, as I understand we should proxy a string only for the ` text/html` type.

For another supported types:
    "text/xml",
    "application/xml",
    "application/xhtml+xml",
    "image/svg+xml"
 we should not process string.

We have an issue for XHTML  support - https://github.com/DevExpress/testcafe-hammerhead/issues/350

@churkin

LavrovArtem: I'm disagree with
> args.length is an uncessary check. Drop this.

The `parseFromString` function throw an error `Failed to execute 'parseFromString' on 'DOMParser': 2 arguments required, but only 1 present.` and we don't need to process the string.

LavrovArtem:
@testcafe-build-bot \retest

src/client/sandbox/node/window.js
```diff
@@ -326,6 +326,15 @@ export default class WindowSandbox extends SandboxBase {
             };
         }
 
+        if (window.DOMParser) {
+            window.DOMParser.prototype.parseFromString = function (...args) {
```
miherlosev: Why do you override the `parseFromString` from `DOMParser.prototype`?

src/client/sandbox/native-methods.js
```diff
@@ -173,6 +173,10 @@ class NativeMethods {
         // Object
         this.objectToString = win.Object.prototype.toString;
 
+        // DOMParser
+        if (win.DOMParser)
+            this.DOMParserParseFromString = win.DOMParser.prototype.parseFromString;
```
miherlosev: Why do you store value from prototype?

src/client/sandbox/node/window.js
LavrovArtem: What do you meen?

LavrovArtem:
@testcafe-build-bot \retest

test/client/fixtures/sandbox/node/methods-test.js
```diff
@@ -320,6 +320,28 @@ if (!browserUtils.isFirefox) {
     });
 }
 
+if (window.DOMParser) {
+    test('DOMParser.parseFromString', function () {
+        var htmlStr        = '<a href="/path">Link</a>';
+        var parsedDocument = new DOMParser().parseFromString(htmlStr, 'text/html');
```
miherlosev: Small style issue.

Rewrite as
```js
var domParser           = new DOMParser();
var parsedDocument = domParser.parseFromString(htmlStr, 'text/html');
```
and remove `new DOMParser()`  below

LavrovArtem:
FPR

LavrovArtem:
ping @churkin

src/client/sandbox/node/window.js
```diff
@@ -326,6 +326,16 @@ export default class WindowSandbox extends SandboxBase {
             };
         }
 
+        // NOTE: DOMParser supports an HTML parsing for IE10 and later
+        if (window.DOMParser && !isIE9) {
+            window.DOMParser.prototype.parseFromString = function (...args) {
+                if (args.length > 1 && typeof args[0] === 'string' && args[1] === 'text/html')
```
churkin: `args.length` is extra check

churkin: Sorry, I was wrong

PR #1136, opened 2017-05-18
fix 'Should emulate native behaviour on calling 'Worker' constructor without parameters' (close #1132)

### LINKED ISSUE #1132: Should emulate native behaviour on calling 'Worker' constructor without parameters
Without proxy:
```js
new Worker()
Uncaught TypeError: Failed to construct 'Worker': 1 argument required, but only 0 present.
```

With proxy:
```js
new Worker()
Failed to load resource: the server responded with a status of 404 (Not Found)
http:<proxy-server-url>/<site-url>/undefined
```

See https://html.spec.whatwg.org/multipage/workers.html#dom-worker

miherlosev:
@churkin @LavrovArtem

### DIFF
--- a/src/client/sandbox/node/window.js
+++ b/src/client/sandbox/node/window.js
@@ -11,6 +11,7 @@ import { isSubDomain, parseUrl, getProxyUrl, convertToProxyUrl, stringifyResourc
 import { isFirefox } from '../../utils/browser';
 import { isCrossDomainWindows, isImgElement, isBlob } from '../../utils/dom';
 import INTERNAL_ATTRS from '../../../processing/dom/internal-attributes';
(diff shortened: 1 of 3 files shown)


miherlosev:
FPR

src/client/sandbox/node/window.js
```diff
@@ -128,11 +128,17 @@ export default class WindowSandbox extends SandboxBase {
         }
 
         if (window.Worker) {
-            window.Worker           = scriptURL => {
+            window.Worker           = function (scriptURL, options) {
+                if (this instanceof window.Worker === false)
```
miherlosev: After moving check to a separate method we with @churkin decide that comment is unnecessary.
So @LavrovArtem review this

PR #1152, opened 2017-05-29
Use 'for...of' loop for array

miherlosev:
@churkin @LavrovArtem 

We can not use `for ...of` loop for non array classes.

We use `loose=true` option value for [`babel-plugin-transform-es2015-for-of`](https://www.npmjs.com/package/babel-plugin-transform-es2015-for-of) plugin.
With this option babel generates simplified output that causes an error in browsers not supported `Symbol` implementation (for example, `IE`).

### DIFF
--- a/src/client/sandbox/code-instrumentation/location/wrapper.js
+++ b/src/client/sandbox/code-instrumentation/location/wrapper.js
@@ -101,8 +101,8 @@ export default class LocationWrapper extends EventEmitter {
             }));
         };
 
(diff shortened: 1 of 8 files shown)


miherlosev:
@testcafe-build-bot \retest

miherlosev:
@testcafe-build-bot \retest

miherlosev:
@testcafe-build-bot \retest

miherlosev:
@testcafe-build-bot \retest

miherlosev:
FPR

PR #1163, opened 2017-05-31
temporary fix for testcafe tests

miherlosev:
@churkin @LavrovArtem 

Our mechanism of the running testcafe tests on every PR doesn't work with `npm5`.
It is the temporary solution until the origin problem will be fixed


miherlosev:
@testcafe-build-bot \retest

PR #1170, opened 2017-06-08
Add detection of DataTransfer support

AndreyBelym:
\cc @churkin @miherlosev @AlexanderMoskovkin


src/client/sandbox/event/drag-and-drop/data-transfer.js
```diff
@@ -118,6 +118,11 @@ export default class DataTransfer {
                 itemListInternals.removeItem(format);
         };
     }
+
+    static isSupported () {
```
miherlosev: Move to the `client\utils\browser.js`

PR #1199, opened 2017-07-03
fixed the wrong script processing (close #1197)

### LINKED ISSUE #1197: Transformer post-message-get can create invalid Javascript
## Short Description

I encountered a bug with testcafe-hammerhead, or more precicely, this transform:

https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/processing/script/transformers/post-message-get.js

This transform produces invalid Javascript when we have an object method of the name `postMessage` on a plain object, and that code is transpiled by babel (preset used:babel-preset-env) and bundled with webpack.

While the code runs well without testcafe, when accessed through testcafe running a fixture, it turns this:
```javascript
/* harmony default export */ __webpack_exports__["default"] = ({

  methods: {
    postMessage: function postMessage(event) {
      console.log('Hello!');
    }
  }
});
```

into this:

```javascript
/* harmony default export */ __webpack_exports__["default"] = ({

  methods: {
    postMessage: function  __get$PostMessage(null,postMessage)(event) {
      console.log('Hello!');
    }
  }
});
```

However, this doesn't seem to happen when the code is not bundled with webpack.

## Demo reproduction Repository

https://github.com/LinusBorg/testcafe-hammerhead-postMessage-bug

The repository's README has more information and detailed instructions about how to reproduce the behaviour.

LavrovArtem:
/cc @miherlosev @churkin

### DIFF
--- a/src/processing/script/transformers/eval-get.js
+++ b/src/processing/script/transformers/eval-get.js
@@ -35,6 +35,11 @@ export default {
             if (parent.type === Syntax.MemberExpression)
                 return false;
 
+            // Skip: function eval () { ... }
+            if ((parent.type === Syntax.FunctionExpression || parent.type === Syntax.FunctionDeclaration) &&
+                parent.id === node)
+                return false;
+
             // Skip: function (eval) { ... } || function func(eval) { ... }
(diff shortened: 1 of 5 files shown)


LavrovArtem:
ping @churkin

PR #1200, opened 2017-07-03
move feature detection code from browserUtils (close #1171)

### LINKED ISSUE #1171: Move feature detection code from client/utils/browser
Move feature detection code from `client/utils/browser.js` to the `client/utils/feature-detection.js`
```js
//Feature detection
export var hasTouchEvents = !!('ontouchstart' in window);
// NOTE: We need to check touch points only for IE, because it has PointerEvent and MSPointerEvent (IE10, IE11)
// instead of TouchEvent (T109295).
export var hasTouchPoints = isIE && (navigator.maxTouchPoints > 0 || navigator.msMaxTouchPoints > 0);
export var isTouchDevice  = !!(info.mobile || info.tablet) && hasTouchEvents;
```

and publish `feature-detection` in hammerhead

miherlosev:
@churkin @LavrovArtem

### DIFF
--- a/src/client/index.js
+++ b/src/client/index.js
@@ -15,6 +15,7 @@ import * as styleUtils from './utils/style';
 import trim from '../utils/string-trim';
 import { isRelativeUrl, parseProxyUrl } from '../utils/url';
 import * as urlUtils from './utils/url';
+import * as featureDetection from './utils/feature-detection';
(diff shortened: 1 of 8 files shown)


miherlosev:
It is a breaking change to the `testcafe`.
So, FPR

PR #1239, opened 2017-07-22
Implement listeners.getEventListeners (closes #1238)

### LINKED ISSUE #1238: Implement GetEventListeners method in sandbox.event.listeners
It will allow to know are there any listeners for the window.load event (it's necessary for TestCafe, for improving element waiting mechanizm)

AlexanderMoskovkin:
/cc @miherlosev

### DIFF
--- a/src/client/sandbox/event/listeners.js
+++ b/src/client/sandbox/event/listeners.js
@@ -296,4 +296,18 @@ export default class Listeners extends EventEmitter {
             eventListeningInfo.outerHandlersWrapper = wrapper;
         }
     }
+
+    getEventListeners (el, event) {
+        const eventCtx = this.listeningCtx.getEventCtx(el, event);
+
+        if (!eventCtx)
+            return null;
+
+        const outerHandlers = [];
+
+        for (let i = 0; i < eventCtx.outerHandlers.length; i++)
+            outerHandlers.push(eventCtx.outerHandlers[i].fn);
+
+        return outerHandlers;
+    }
 }
--- a/src/client/sandbox/event/listeners.js
+++ b/src/client/sandbox/event/listeners.js
@@ -307,6 +307,6 @@ export default class Listeners extends EventEmitter {
         if (!eventCtx)
             return null;
 
-        return arrayMap.apply(eventCtx.outerHandlers, [handler => handler.fn]);
+        return arrayMap.call(eventCtx.outerHandlers, handler => handler.fn);
     }
 }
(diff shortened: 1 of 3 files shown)


src/client/sandbox/event/listeners.js
```diff
@@ -296,4 +296,18 @@ export default class Listeners extends EventEmitter {
...
+
+    getEventListeners (el, event) {
+        const eventCtx = this.listeningCtx.getEventCtx(el, event);
+
+        if (!eventCtx)
+            return null;
+
+        const outerHandlers = [];
+
+        for (let i = 0; i < eventCtx.outerHandlers.length; i++)
```
miherlosev: use the `Array.map` function. 
See example of the array' methods using - https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/event-emitter.js#L3

src/client/sandbox/event/listeners.js
```diff
@@ -296,4 +296,18 @@ export default class Listeners extends EventEmitter {
             eventListeningInfo.outerHandlersWrapper = wrapper;
         }
     }
+
+    getEventListeners (el, event) {
+        const eventCtx = this.listeningCtx.getEventCtx(el, event);
+
+        if (!eventCtx)
+            return null;
```
miherlosev: I propose to return an empty array in this case.
It allows to simplify code that will be use it.

test/client/fixtures/sandbox/event/internal-listeners-test.js
```diff
@@ -157,6 +157,32 @@ test('initElementListening', function () {
     checkHandlerCounters(1, 2, 2, 1);
 });
 
+test('getEventListeners', function () {
+    var onClick        = function () {
+    };
+
+    strictEqual(listeners.getEventListeners(container, 'click'), null);
```
miherlosev: For consistency use the `strictEqual` for `null` check and `deepEqual` with array check.

src/client/sandbox/event/listeners.js
AlexanderMoskovkin: I don't think so. It opens a way to fail in the code.
In fact we have two states: 
1) we don't listen element, so we don't know does it have handlers
2) we listen it and return results.

Imagine I'll use it in `testcafe` after some time for some element. And I get `[]` every time. How can I be sure is it element without event listeners or it's just not listened by hammerhead (and I should call `initElementListening`)?

miherlosev: Agree

AlexanderMoskovkin:
FPR

miherlosev:
@testcafe-build-bot \retest

miherlosev:
@testcafe-build-bot \retest

miherlosev:
@testcafe-build-bot \retest

PR #1285, opened 2017-08-29
fix `Page throws an error when handler passed to the addEventListener function equals undefined` (close #1251)

### LINKED ISSUE #1251: Page throws an error when handler passed to the addEventListener function equals undefined
The error does not occurs on page without `hammerhead`.
```javascript
window.addEventListener("message", void 0);
```
Page: http://zaycev.net/

LavrovArtem:
/cc @miherlosev  @churkin

### DIFF
--- a/src/client/sandbox/event/listeners.js
+++ b/src/client/sandbox/event/listeners.js
@@ -87,9 +87,8 @@ export default class Listeners extends EventEmitter {
         return true;
     }
 
-    static _isCorrectListener (listener) {
-        return listener && (typeof listener === 'function' ||
-                            typeof listener === 'object' && typeof listener.handleEvent === 'function');
+    static _isValidEventListener (listener) {
+        return typeof listener === 'function' || isObjectEventListener(listener);
     }
 
     _createEventHandler () {
@@ -148,7 +147,7 @@ export default class Listeners extends EventEmitter {
                 const eventListeningInfo     = listeningCtx.getEventCtx(el, type);
                 const nativeAddEventListener = Listeners._getNativeAddEventListener(el);
 
-                if (!eventListeningInfo || !Listeners._isCorrectListener(listener))
+                if (!eventListeningInfo || !Listeners._isValidEventListener(listener))
                     return nativeAddEventListener.apply(el, args);
 
                 // NOTE: T233158
@@ -177,7 +176,7 @@ export default class Listeners extends EventEmitter {
                 const nativeRemoveEventListener = Listeners._getNativeRemoveEventListener(el);
                 const eventCtx                  = listeningCtx.getEventCtx(el, type);
 
-                if (!eventCtx || !Listeners._isCorrectListener(listener))
+                if (!eventCtx || !Listeners._isValidEventListener(listener))
                     return nativeRemoveEventListener.apply(el, args);
 
                 args[1] = listeningCtx.getWrapper(eventCtx, listener, useCapture);
@@ -201,7 +200,7 @@ export default class Listeners extends EventEmitter {
                 const docEventListeningInfo = listeningCtx.getEventCtx(doc, type);
                 const eventListeningInfo    = listeningCtx.getEventCtx(this, type);
 
-                if (!docEventListeningInfo || !Listeners._isCorrectListener(listener))
+                if (!docEventListeningInfo || !Listeners._isValidEventListener(listener))
                     return nativeAddEventListener.call(this, type, listener, useCapture);
 
                 // NOTE: T233158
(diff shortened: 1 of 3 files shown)


src/client/sandbox/event/listeners.js
```diff
@@ -87,6 +87,11 @@ export default class Listeners extends EventEmitter {
         return true;
     }
 
+    static _isCorrectListener (listener) {
```
miherlosev: Rename to the `_isValidEventListener (listener)`

src/client/sandbox/event/listeners.js
```diff
@@ -87,6 +87,11 @@ export default class Listeners extends EventEmitter {
...
+        return listener && (typeof listener === 'function' ||
```
miherlosev: `return typeof listener === 'function' || eventUtils.isObjectEventListener(listener)`;

test/client/fixtures/sandbox/event/event-test.js
```diff
@@ -301,3 +302,21 @@ test('SVGElement.dispatchEvent should be overriden (GH-614)', function () {
 
     svg.parentNode.removeChild(svg);
 });
+
+test('should not wrapped invalid handlers (GH-1251)', function () {
```
miherlosev: test('should not wrap invalid event handlers (GH-1251)`

test/client/fixtures/sandbox/event/event-test.js
```diff
@@ -301,3 +302,21 @@ test('SVGElement.dispatchEvent should be overriden (GH-614)', function () {
...
+    var handlers = [void 0, 1, null, 'str', {}];
+
+    var testHandlers = function (obj) {
```
miherlosev: `var testHandlers = function (target) {`

src/client/utils/event.js
```diff
@@ -53,7 +53,7 @@ export function stopPropagation (ev) {
 }
 
 export function isObjectEventListener (listener) {
-    return typeof listener === 'object' && typeof listener.handleEvent === 'function';
+    return listener && typeof listener === 'object' && typeof listener.handleEvent === 'function';
```
miherlosev: First check needs only for preventing error on access to the `listener.handleEvent` property then `listener` is null.
Let's rephrase as 
`return typeof listener === 'object' && typeof (listener && listener.handleEvent) === 'function'`;

test/client/fixtures/sandbox/event/event-test.js
```diff
@@ -301,3 +302,31 @@ test('SVGElement.dispatchEvent should be overriden (GH-614)', function () {
 
     svg.parentNode.removeChild(svg);
 });
+
+test('should not wrap invalid event handlers (GH-1251)', function () {
+    var handlers = [void 0, 1, null, 'str', {}];
+
+    // NOTE: some handlers in some browsers throw the "Invalid argument" error
```
miherlosev: Don't exclude handlers that raise exception from testing.
Let's change test scenario to:
- memory handlers that raises an error (nativeAddEventListener)
- add handler to overriden addEventListener. If it raises an error then try to find approprite error in stored errors. If it called without error then check the `listeningCtx`.

Also add `Symbol` type checking (if browser supports this type)

src/client/utils/event.js
LavrovArtem: Then `return typeof listener === 'object' && listener && listener.handleEvent === 'function';`

miherlosev: I don't understand what do you mean

LavrovArtem: Sorry I am confused, that's right variant:
`return typeof listener === 'object' && listener && typeof listener.handleEvent === 'function';`

miherlosev: Ok

LavrovArtem:
FPR

test/client/fixtures/sandbox/event/event-test.js
```diff
@@ -301,3 +302,44 @@ test('SVGElement.dispatchEvent should be overriden (GH-614)', function () {
...
+        catch (e) {
+            if (!errorText)
+                errorText = e.toString();
+
+            invalidHandlers.push(handlers[i]);
+        }
+    }
+
+    var testHandlers = function (target) {
+        var storedHandlersCount = listeningCtx.getEventCtx(target, 'click').wrappers.length;
```
miherlosev: `storedHandlerWrappersCount`

test/client/fixtures/sandbox/event/event-test.js
```diff
@@ -301,3 +302,44 @@ test('SVGElement.dispatchEvent should be overriden (GH-614)', function () {
...
+
+test('should not wrap invalid event handlers (GH-1251)', function () {
+    var handlers        = [void 0, 1, null, 'str', {}];
+    var invalidHandlers = [];
+    var errorText       = null;
+
+    if ('Symbol' in window)
+        handlers.push(Symbol('foo'));
+
+    // NOTE: some handlers in some browsers throw the "Invalid argument" error
```
miherlosev: // NOTE:  on adding some type of handlers an "Invalid argument" error can be raised

PR #1294, opened 2017-09-05
fix `Onclick handler probably doesn't work after click on 'Submit' button in form` (close #1291)

### LINKED ISSUE #1291: Onclick handler probably doesn't work after click on 'Submit' button in form
Converted from a [question](https://testcafe-discuss.devexpress.com/t/problem-with-form-submit-onclick-function/436) on the Discuss forum. Can be reproduced in playground.

test.js
```js
import { ClientFunction } from 'testcafe';

fixture `f1`.page `./temp.html`;

test('t1', async t => {
 await t
        .typeText('#text', 'Testing')
        .click('#save_button');

 const getSearch = await ClientFunction(() => location.search);

 await t.expect(getSearch()).eql('?action=Save&text=Testing&button=Save');

});
```

temp.html
```html
<html>
    <head>
        <title>Tets page</title>
    </head>
    <body>
        <form action="temp.html" method="GET">
            <input id="action" name="action" value="Edit" hidden />
            <input id="text" name="text" type="text" />
            <input title="Save" onclick="this.form.action.value='Save'" type="submit" name="button" id="save_button" value="Save" />
        </form>
    </body>
</html>

```

LavrovArtem:
/cc @miherlosev @helen-dikareva

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -136,7 +136,12 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
             action: {
                 condition: el => domUtils.isDomElement(el) && domProcessor.isUrlAttr(el, 'action'),
 
-                get: el => PropertyAccessorsInstrumentation._getUrlAttr(el, 'action'),
+                get: el => {
+                    if (domUtils.isDomElement(el.action))
+                        return el.action;
+
+                    return PropertyAccessorsInstrumentation._getUrlAttr(el, 'action');
+                },
                 set: (el, value) => {
                     el.setAttribute('action', value);
 
--- a/test/client/fixtures/sandbox/code-instrumentation/getters-test.js
+++ b/test/client/fixtures/sandbox/code-instrumentation/getters-test.js
@@ -607,26 +607,31 @@ test('document.activeElement when it equals null (GH-1226)', function () {
     div.parentNode.removeChild(div);
 });
 
-if (!browserUtils.isIE) {
-    test('form.action should return element when the form contains element with the "action" attribute name (GH-1291)', function () {
-        var form  = document.createElement('form');
-        var input = document.createElement('input');
+test('form.action should return element when the form contains element with the "action" attribute name (GH-1291)', function () {
+    var form        = document.createElement('form');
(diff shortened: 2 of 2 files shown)


LavrovArtem:
@testcafe-build-bot \retest

src/client/sandbox/code-instrumentation/properties/index.js
```diff
@@ -136,7 +136,12 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
             action: {
                 condition: el => domUtils.isDomElement(el) && domProcessor.isUrlAttr(el, 'action'),
 
-                get: el => PropertyAccessorsInstrumentation._getUrlAttr(el, 'action'),
+                get: el => {
+                    if (domUtils.isDomElement(el.action))
```
miherlosev: I propose to invert logic:
```js
if typeof el.action === 'string'
   return PropertyAccessorsInstrumentation._getUrlAttr(el, 'action');

return el.action;
```
It helps to prevent edge cases when form.action is not dom element.
And add test for this case

LavrovArtem:
FPR

test/client/fixtures/sandbox/code-instrumentation/getters-test.js
```diff
@@ -606,3 +606,27 @@ test('document.activeElement when it equals null (GH-1226)', function () {
 
     div.parentNode.removeChild(div);
 });
+
+if (!browserUtils.isIE) {
```
miherlosev: Why do you exclude IE?

test/client/fixtures/sandbox/code-instrumentation/getters-test.js
```diff
@@ -606,3 +606,27 @@ test('document.activeElement when it equals null (GH-1226)', function () {
...
+    test('form.action should return element when the form contains element with the "action" attribute name (GH-1291)', function () {
+        var form  = document.createElement('form');
+        var input = document.createElement('input');
+
```
miherlosev: Use native element behaviour to get expected values.
```js
var nativeForm = nativeMethods.createElement....
var nativeInput = nativeMethods.createElement...
```

LavrovArtem:
FPR

PR #1301, opened 2017-09-08
fix 'Should not raise an error if processed element not extensible' (close #1300)

### LINKED ISSUE #1300: Should not raise an error if processed element not extensible
There are 3 ways to 'freeze' element: 
- [Object.preventExtentions](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions)
- [Object.freeze](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze)
- [Object.seal](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Object/seal)

If we create an element using a non-overriden way (for example, `document.registerElement`), 'freeze' this element and add to a some element in `DOM`, then we get the `TypeError: can't define property "hammerhead|processed-context": <some_element_class> is not extensible`

Based on https://github.com/DevExpress/testcafe/issues/1767

miherlosev:
@LavrovArtem @churkin

### DIFF
--- a/src/client/sandbox/native-methods.js
+++ b/src/client/sandbox/native-methods.js
@@ -178,6 +178,8 @@ class NativeMethods {
         this.objectKeys           = win.Object.keys;
         this.objectDefineProperty = win.Object.defineProperty;
         this.objectCreate         = win.Object.create;
+        this.objectIsExtensible   = win.Object.isExtensible;
+        this.objectIsFrozen       = win.Object.isFrozen;
 
         // DOMParser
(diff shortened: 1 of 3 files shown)


LavrovArtem:
@testcafe-build-bot \retest

miherlosev:
FPR

miherlosev:
ping @churkin

PR #1307, opened 2017-09-11
TestCafe qunit tests for input type "number" fixed

helen-dikareva:
see comment [here](https://github.com/DevExpress/testcafe/pull/1768#issuecomment-327791998)


miherlosev:
Add test for fixed case on hammerhead side.

PR #1310, opened 2017-09-13
Use `iltorb` module instead of 'brotli`

LavrovArtem:
/cc @miherlosev @AlexanderMoskovkin

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,11 @@
   },
   "dependencies": {
     "babel-runtime": "^6.3.19",
-    "brotli": "^1.3.1",
     "crypto-md5": "^1.0.0",
     "css": "2.2.1",
     "dedent": "0.4.0",
     "iconv-lite": "0.4.11",
+    "iltorb": "1.3.6",
     "lodash": "4.16.4",
     "lru-cache": "2.6.3",
     "mime": "~1.3.4",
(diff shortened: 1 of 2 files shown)


AlexanderMoskovkin:
let's test it on health-monitor before merging

miherlosev:
> let's test it on health-monitor before merging

Yes, we discussed in private this.
Also, please provide information about:
- decode/encode large file (about 1Mb) with old and new library and provide speed and memory usage comparison
- measure memory allocation on library initialization
- encode

LavrovArtem:
Size of the source hammerhead.js file is 1mb.

| | brotli | iltorb
--- | ---: | ---:
encoding time | 5100ms | 2400ms
size after encoding | 531kb | 149kb
decoding time | 35ms | 5ms
memory allocation on library initialization | 310mb | 2.5mb
memory usage | 315mb | 4.1mb

miherlosev:
Two health monitor runs:
https://testcafe-hhhm.devexpress.com/history?timestamp=1505311354244
https://testcafe-hhhm.devexpress.com/history?timestamp=1505333732609

The `itorb` module is faster and requires less memory on module initialization and during decoding/encoding.
There is no new probles or performance degradation.

So, there is no reasons to don't merge it.
/cc @AlexanderMoskovkin

AlexanderMoskovkin:
awesome!

PR #1312, opened 2017-09-14
Implement `ConsoleSandbox`. Add the `consoleMethCalled` event.

AlexanderMoskovkin:
It's for https://github.com/DevExpress/testcafe/issues/1738

### DIFF
--- a/src/client/sandbox/console.js
+++ b/src/client/sandbox/console.js
@@ -1,0 +1,27 @@
+import SandboxBase from './base';
+
+export default class ConsoleSandbox extends SandboxBase {
+    constructor () {
+        super();
+
+        this.CONSOLE_METH_CALLED = 'hammerhead|console|console-meth-called-event';
+    }
+
+    attach (window) {
+        super.attach(window);
+
+        const sandbox = this;
+
+        const proxyConsoleMeth = meth => {
+            sandbox.window.console[meth] = function () {
+                sandbox.emit(sandbox.CONSOLE_METH_CALLED, { meth, args: arguments });
+                sandbox.nativeMethods.consoleMeths[meth].apply(sandbox.nativeMethods.console, arguments);
+            };
+        };
+
+        proxyConsoleMeth('log');
+        proxyConsoleMeth('info');
+        proxyConsoleMeth('error');
+        proxyConsoleMeth('warn');
+    }
+}
--- a/src/client/sandbox/console.js
+++ b/src/client/sandbox/console.js
@@ -1,15 +1,28 @@
 import SandboxBase from './base';
+import { isCrossDomainWindows } from '../utils/dom';
 
 export default class ConsoleSandbox extends SandboxBase {
-    constructor () {
+    constructor (messageSandbox) {
         super();
 
         this.CONSOLE_METH_CALLED_EVENT = 'hammerhead|event|console-meth-called';
+
+        this.messageSandbox = messageSandbox;
     }
 
     _proxyConsoleMeth (meth) {
         this.window.console[meth] = (...args) => {
-            this.emit(this.CONSOLE_METH_CALLED_EVENT, { meth, args: args });
+            if (!isCrossDomainWindows(window, window.top)) {
+                const sendToTopWindow = window !== window.top;
+
+                if (sendToTopWindow) {
+                    this.emit(this.CONSOLE_METH_CALLED_EVENT, { meth, args, inIframe: true });
+                    this.messageSandbox.sendServiceMsg({ meth, args, cmd: this.CONSOLE_METH_CALLED_EVENT }, window.top);
+                }
+                else
+                    this.emit(this.CONSOLE_METH_CALLED_EVENT, { meth, args });
+            }
+
             this.nativeMethods.consoleMeths[meth].apply(this.nativeMethods.console, args);
         };
     }
@@ -25,5 +38,14 @@ export default class ConsoleSandbox extends SandboxBase {
             this._proxyConsoleMeth('error');
             this._proxyConsoleMeth('warn');
         }
+
+        const messageSandbox = this.messageSandbox;
+
+        messageSandbox.on(messageSandbox.SERVICE_MSG_RECEIVED_EVENT, e => {
+            const message = e.message;
+
+            if (message.cmd === this.CONSOLE_METH_CALLED_EVENT)
+                this.emit(this.CONSOLE_METH_CALLED_EVENT, { meth: message.meth, args: message.args });
+        });
     }
 }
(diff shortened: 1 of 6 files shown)


test/client/fixtures/sandbox/console-test.js
```diff
@@ -0,0 +1,66 @@
...
+        info:  console.info
+    };
+
+    function addToLog (meth, args) {
+        log += argsToArray(args).join('');
+
+        originMethods[meth].apply(console, args);
+    }
+
+    window.console = {
```
miherlosev: To capture console trace you override methods that already overriden in sandbox.
Usually, we use an another way - we override the appropriate native methods.
See tests in https://github.com/DevExpress/testcafe-hammerhead/blob/master/test/client/fixtures/sandbox/node/methods-test.js
Use this way too.

test/client/fixtures/sandbox/console-test.js
```diff
@@ -0,0 +1,66 @@
...
+        logFromEvents += argsToArray(e.args).join('');
+    });
+
+    window.console.log('1', '2');
+    window.console.warn('3', '4');
+    window.console.error('5', '6');
+    window.console.info('7', '8');
+    /* eslint-enable no-console */
+
+    deepEqual(handledEvents, ['log', 'warn', 'error', 'info']);
```
miherlosev: `handledConsoleMethodCalls`

test/client/fixtures/sandbox/console-test.js
```diff
@@ -0,0 +1,66 @@
...
+    deepEqual(log, '12345678');
+    deepEqual(logFromEvents, '12345678');
```
miherlosev: `handledConsoleMethodArgs`

src/client/sandbox/console.js
```diff
@@ -0,0 +1,27 @@
...
+
+        this.CONSOLE_METH_CALLED = 'hammerhead|console|console-meth-called-event';
+    }
+
+    attach (window) {
+        super.attach(window);
+
+        const sandbox = this;
+
+        const proxyConsoleMeth = meth => {
```
LavrovArtem: Make this function as private member of ConsoleSandbox, I think it is more redable.

src/client/sandbox/native-methods.js
```diff
@@ -212,6 +212,16 @@ class NativeMethods {
             this.CSS2PropertyRemoveProperty   = win.CSS2Property.prototype.removeProperty;
         }
 
+        // Console
+        this.console = win.console;
+
+        this.consoleMeths = {
```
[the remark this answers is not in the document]
AlexanderMoskovkin: Why? Is there a significant reason for this? I believe it's better to leave it as is to use
```js
sandbox.nativeMethods.consoleMeths[meth]
```
instead of
```js
const nativeMeth = console + meth[0].toUpperCase() + meth.slice(1);
sandbox.nativeMethods[nativeMeth];
```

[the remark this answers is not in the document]
miherlosev: Ok. If you use a console method as key then let you leave it as is.

src/client/sandbox/console.js
```diff
@@ -0,0 +1,31 @@
...
+        this.window.console[meth] = (...args) => {
+            this.emit(this.CONSOLE_METH_CALLED_EVENT, { meth, args: args });
+            this.nativeMethods.consoleMeths[meth].apply(this.nativeMethods.console, args);
+        };
+    }
+
+    attach (window) {
+        super.attach(window);
+
+        const sandbox = this;
```
LavrovArtem: remove this variable

AlexanderMoskovkin: nice catch

AlexanderMoskovkin:
FPR

src/client/sandbox/console.js
```diff
@@ -0,0 +1,29 @@
+import SandboxBase from './base';
+
+export default class ConsoleSandbox extends SandboxBase {
+    constructor () {
+        super();
+
+        this.CONSOLE_METH_CALLED_EVENT = 'hammerhead|console|console-meth-called';
```
miherlosev: Rename to 'hammerhead|event|console-meth-called'

src/client/sandbox/console.js
```diff
@@ -0,0 +1,29 @@
...
+        super.attach(window);
+
+        if (this.window.console && typeof this.window.console.log.apply === 'function') {
+            // NOTE: ie9 has no the `window.console` object when developer tools are closed.
+            // If they are opened there is no `console.log.apply`, so we skip this case.
+            this._proxyConsoleMeth('log');
+            this._proxyConsoleMeth('info');
+            this._proxyConsoleMeth('error');
+            this._proxyConsoleMeth('warn');
+        }
```
miherlosev: Should we override the `console.dir` method too?

src/client/sandbox/index.js
```diff
@@ -198,6 +200,7 @@ export default class Sandbox extends SandboxBase {
         this.node.attach(window);
         this.upload.attach(window);
         this.cookie.attach(window);
+        this.console.attach(window);
```
miherlosev: Should we reattach `ConsoleSandbox` after `document.write` is called?
See `reattach` method and `DOCUMENT_CLEANED_EVENT` event.

src/client/sandbox/console.js
AlexanderMoskovkin: I guess not.

https://developer.mozilla.org/en-US/docs/Web/API/Console/dir

> Non-standard
This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user.

test/client/fixtures/sandbox/console-test.js
```diff
@@ -47,4 +47,16 @@ if (window.console && typeof window.console.log !== 'undefined') {
             info:  originMethods.info
         };
     });
+
+    test('consoleMethCalled event after document.write', function () {
```
miherlosev: `'consoleMethod' event should be raised after document.write`

AlexanderMoskovkin: 'consoleMethCalled' event...

miherlosev: Yep

src/client/sandbox/index.js
AlexanderMoskovkin: It seems, we should

AlexanderMoskovkin:
FPR

AlexanderMoskovkin:
@testcafe-build-bot \retest

AlexanderMoskovkin:
I've rewritten things about iframes, take a look
/cc @miherlosev @LavrovArtem

PR #1331, opened 2017-09-28
Add fix for crossing shadow boundary in getParents()

varunkumar:
Add fix for scrolling elements inside ShadowDOM. https://github.com/DevExpress/testcafe/issues/1808


miherlosev:
I've checked your PR and found out that it does not completely fix the related TestCafe's issue.

If you run the following test code against your PR, it will fail.
```js
test('"getParents" should work properly for elements inside shadowDOM', function () {
    var host  = document.createElement('div');
    var root  = host.createShadowRoot();
    var input = document.createElement('input');

    document.body.appendChild(host);
    root.appendChild(input);

    var parents = domUtils.getParents(input);

    deepEqual(parents, [root, host, document.body, document.documentElement]);
   document.body.removeChild(host);
});
```

It happens because we check that element's parent is `ELEMENT_NODE` (see https://github.com/DevExpress/testcafe-hammerhead/blob/e0e9e8c4d2ca600183282ef9ca41becbf4c758b2/src/client/utils/dom.js#L747)
But element's parent is `DOCUMENT_FRAGMENT_NODE`.
Also we should work properly with document fragment nodes that are not in shadowDOM.

AlexanderMoskovkin:
Guys, since the `getParents` function became more complex please add a simple qunit test for the ShadowDOM case

miherlosev:
Guys, we will improve this functionality in context of support web components.

varunkumar:
Is there any timeline for this? We use `web-components` heavily and having a full fledged support from TestCafe would be helpful.

miherlosev:
We are planning to start working with your PR next week.

miherlosev:
@testcafe-build-bot \retest

miherlosev:
FPR
@LavrovArtem @AndreyBelym

miherlosev:
ping @LavrovArtem @AndreyBelym

PR #1332, opened 2017-09-28
Revert "Use `iltorb` module instead of 'brotli` (#1310)"

LavrovArtem:


LavrovArtem:
We have to revert the fix since we've faced a major issue there. The [iltorb](https://www.npmjs.com/package/iltorb) module requires the c++ compiler. There are compiled binaries for linux, win and mac platforms for nodejs versions 4, 6 and 8. But if you use it with nodejs v5 or v7 node tries to compile c++ files. C++ compiler is not available on most operating systems by default. In this case the iltorb module installation fails.
We will try to find the best way to resolve this and fix the initial issue.

PR #1335, opened 2017-10-03
fix `The elements processing does not work for documentFragment after it was added to DOM` (close #1334)

### LINKED ISSUE #1334: The elements processing does not work for documentFragment after it was added to DOM
The `appendChild` and `insertBefore` wrapped functions are called the [_onElementAdded (el)](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/element.js#L551) function after `el` was added to DOM.
If `el` is `documentFragment`, the empty `documentFragment` is passed into the `_onElementAdded` function .

LavrovArtem:
/cc @miherlosev @AlexanderMoskovkin

### DIFF
--- a/src/client/sandbox/node/element.js
+++ b/src/client/sandbox/node/element.js
@@ -296,7 +296,7 @@ export default class ElementSandbox extends SandboxBase {
         return result;
     }
 
-    _insertNodeCore (parentNode, args, nativeFn, needToCheckThatBodyElIsPresent) {
+    _addNodeCore ({ parentNode, args, nativeFn, checkBody }) {
         const newNode = args[0];
 
         this._prepareNodeForInsertion(newNode, parentNode);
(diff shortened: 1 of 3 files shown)


src/client/sandbox/node/element.js
```diff
@@ -292,6 +296,37 @@ export default class ElementSandbox extends SandboxBase {
...
+        // certain manipulations and then remove it.
+        // Therefore, we need to check if the body element is present in DOM
+        if (needToCheckThatBodyElIsPresent && domUtils.isBodyElementWithChildren(parentNode) &&
+            domUtils.isElementInDocument(parentNode))
+            result = this.shadowUI.insertBeforeRoot(newNode);
+        else
+            result = nativeFn.apply(parentNode, args);
+
+        if (childNodes) {
+            for (const child of childNodes)
```
miherlosev: instead of `for..of` loop use the `Array.forEach` method

src/client/sandbox/node/element.js
```diff
@@ -551,11 +551,15 @@ export default class ElementSandbox extends SandboxBase {
     _onElementAdded (el) {
         if ((domUtils.isElementNode(el) || domUtils.isDocumentNode(el)) && domUtils.isElementInDocument(el)) {
             const iframes = domUtils.getIframes(el);
+            const scripts = domUtils.getScripts(el);
```
miherlosev: I propose to move this line to `561` row (before `for..of` loop for `scripts`).

PR #1343, opened 2017-10-11
Replace shortid to nanoid

ai:
Since `shortid` [will be replaced](https://github.com/DevExpress/testcafe/pull/1866) to `nanoid` in main `testcafe` repo, I replaced it here as well.

I used full 22 symbols in ID (`shortid` used only 7), since here (as I understood) we need better collision protection. 22 symbols is equal to UUID v4 collision probability, so we can be sure in it.

### DIFF
--- a/src/session/index.js
+++ b/src/session/index.js
@@ -35,7 +35,7 @@ export default class Session extends EventEmitter {
 
     static _generateSessionId () {
         // NOTE: GH-116
-        return nanoid();
+        return nanoid(7).replace(/~/g, '-');
     }
 
     // State


miherlosev:
Hi, @ai .
Thank you for your PR.
At present, we use a unique identifier with 9 symbols.
We want to leave it as is.
Could you please change your PR accordingly ?
```js
nanonid() -> nanoid(9)
```

ai:
Sure. I will update it, when will return to home.

miherlosev:
@testcafe-build-bot \retest

ai:
Do you need any help?

miherlosev:
I've researched the failed tests and found the problem with our proxy.
If sessionId ends with tilda char (`~`) and we open a service `about:blank` page then our proxy failed.
Example of a proxied url: `http://localhost:1401/JSKLJJLASJKJSQ~/about:blank`.

I need an additional time to investigate it.

ai:
We can replace `~` to `-` to be compatible with shortid. I will update PR (we use ~ because trailing - had problems on old Android, but maybe it is not big deal for test cafe)

ai:
@miherlosev I fixed `nanoid` usage, not it is more compatible with `shortid`.

1. I reduced ID size back to 7 as in `shortid`.
2. I replaced `~` to `-` to have `shortid` alphabet.

miherlosev:
Thanks for you work.
I will merge your PR after fix the problem with `~` symbol.

ai:
I removed `~` symbol from PR (it will be replaced to `-`)

miherlosev:
Yes, I saw it. But your PR shows us a problem with the `~` character in proxied urls.
First, we will fix this problem.

ai:
Sure. Maybe then I can revert my ~ to - fix since ~ is more URL safe

miherlosev:
Yes, of course.

miherlosev:
I've created an [another PR](https://github.com/DevExpress/testcafe-hammerhead/pull/1414) because @ai removed own fork.

PR #1352, opened 2017-10-27
fix 'should properly set value of instrumented property if it is readonly for DOM elements' (close #1351)

### LINKED ISSUE #1351: Can't set the `nextSibling` properly of a plain object to null
**Tested page:**
```js
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
</head>
<body>
    <input id="input" />
    <script>
        function MyClass () {
            this.nextSibling = document.getElementById('input');
        }

        var instance = new MyClass();

        instance.nextSibling = null;

        alert(instance.nextSibling);
    </script>
</body>
</html>
```

**Without hammerhead:** `alert(null)`.
**With hammerhead:** `alert(<input-element-description>)`.

**Reason:**
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/code-instrumentation/properties/index.js#L406

```js
nextSibling: {
  condition: node => node.nextSibling && domUtils.isDomElement(node.nextSibling),
  get:       node => domUtils.isShadowUIElement(node.nextSibling) ? null : node.nextSibling,
  set:       () => void 0
},
```

It seems we should add one more check to the condition like: `domUtils.isDomElement(node),`.
Additionally take a look at other properties. It possible there similar things there.

miherlosev:
@AlexanderMoskovkin @churkin

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -180,9 +180,7 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
             },
 
             attributes: {
-                condition: el => {
-                    return el.attributes instanceof window.NamedNodeMap;
-                },
+                condition: el => domUtils.isDomElement(el) && el.attributes instanceof window.NamedNodeMap,
 
                 get: el => getAttributesProperty(el),
                 set: (el, value) => value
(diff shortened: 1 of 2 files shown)


miherlosev:
FPR
@churkin

AlexanderMoskovkin:
ping @churkin

PR #1355, opened 2017-11-02
fix '"Refresh" header should be overriden' (close #1354)

### LINKED ISSUE #1354: 'Refresh' header should be overriden
related to https://github.com/DevExpress/testcafe/issues/1920

see header specification in https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

miherlosev:
@LavrovArtem @churkin


miherlosev:
FPR

Issue #1363, opened 2017-11-08
The href change of the iframe location works wrong


LavrovArtem:
When cross domain iframe changes location, it leaves proxy port as is, but parent window location and new iframe location may be the same domain.

Related with https://github.com/DevExpress/testcafe/issues/1922

LavrovArtem:
I'm working on this

LavrovArtem:
This problem also applies to the processing of the iframe page on server side and client side. We should pay attention to the parent's location when choosing a proxy port for url of elements that change the location of an iframe [for example anchor or area].

To begin with, I'll make a simplified fix that fixes the related issue from TestCafe repo (DevExpress/testcafe#1922).

PR #1365, opened 2017-11-09
Temporary solution for `The href change of the iframe location works wrong #1363`

LavrovArtem:
/cc @churkin @miherlosev


LavrovArtem:
ping @churkin

PR #1095, opened 2017-04-19
fix `Wrong svg element processing in IE` (close #1083)

LavrovArtem:
/cc @churkin @miherlosev

PR #1387, opened 2017-11-24
fix typo

aleks-pro:
/cc @churkin @LavrovArtem

### DIFF
--- a/test/client/fixtures/sandbox/event/unload-test.js
+++ b/test/client/fixtures/sandbox/event/unload-test.js
@@ -30,7 +30,7 @@ asyncTest('BEFORE_UNLOAD_EVENT must be called last (GH-400)', function () {
         });
 });
 
-test('hammerhead UNLOAD_EVENT must be called', function () {
+test('UNLOAD_EVENT should be called', function () {
     var unloadSandbox        = hammerhead.sandbox.event.unload;
     var unloadEventWasCalled = false;
     var handler = function () {
--- a/test/client/fixtures/sandbox/event/unload-test.js
+++ b/test/client/fixtures/sandbox/event/unload-test.js
@@ -30,6 +30,17 @@ asyncTest('BEFORE_UNLOAD_EVENT must be called last (GH-400)', function () {
         });
 });
 
+test('Hammerhead UNLOAD_EVENT must be called (GH-1095)', function () {
+    var unloadSandbox   = hammerhead.sandbox.event.unload;
+    var unloadWasCalled = false;
+
+    hammerhead.on(hammerhead.EVENTS.unload, () => {
+        unloadWasCalled = true;
+    });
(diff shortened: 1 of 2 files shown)


test/client/fixtures/sandbox/event/unload-test.js
```diff
@@ -30,6 +30,17 @@ asyncTest('BEFORE_UNLOAD_EVENT must be called last (GH-400)', function () {
         });
 });
 
+test('Hammerhead UNLOAD_EVENT must be called (GH-1095)', function () {
```
miherlosev: Why do you add `(GH-1095)` suffix to the test name?
test name should be in lowercase.

test/client/fixtures/sandbox/event/unload-test.js
```diff
@@ -30,6 +30,17 @@ asyncTest('BEFORE_UNLOAD_EVENT must be called last (GH-400)', function () {
...
+    var unloadSandbox   = hammerhead.sandbox.event.unload;
+    var unloadWasCalled = false;
```
miherlosev: `var unloadEventWasCalled = `

test/client/fixtures/sandbox/event/unload-test.js
```diff
@@ -30,6 +30,17 @@ asyncTest('BEFORE_UNLOAD_EVENT must be called last (GH-400)', function () {
...
+
+    hammerhead.on(hammerhead.EVENTS.unload, () => {
+        unloadWasCalled = true;
```
miherlosev: Don't use arrow functions in tests.

Need to unsubscribe a test handler
```js
var handler = function () {
      unloadWasCalled = true;
       hammerhead.off(hammerhead.EVENTS.unload, handler);
};
 hammerhead.on(hammerhead.EVENTS.unload, handler);
```

test/client/fixtures/sandbox/event/unload-test.js
aleks-pro: I thought, if you fix an issue and write a test for this case, you need to add such a suffix with issue number. Am i wrong?

test/client/fixtures/sandbox/event/unload-test.js
```diff
@@ -30,6 +30,19 @@ asyncTest('BEFORE_UNLOAD_EVENT must be called last (GH-400)', function () {
         });
 });
 
+test('hammerhead UNLOAD_EVENT must be called', function () {
```
miherlosev: We don't use a `must` work in test names (from recent time ).
So, please, rename your test to `UNLOAD_EVENT should be called`.

Also, we have a small piece of unstable tests in hammerhead and testcafe.
You PR don't affect tests.
To restart tests just add a comment `@testcafe-build-bot  \retest`

aleks-pro:
@testcafe-build-bot \retest

aleks-pro:
FPR

Issue #1384, opened 2017-11-23
Hammerhead emit an error "Failed to find a DNS-record for the resource"


LavrovArtem:
https://testcafe-discuss.devexpress.com/t/failed-to-find-a-dns-record-for-the-resource-at/574

Server for reproducing:
```js
const net = require('net');

let mainPageSocket = null;

const server = net.createServer(socket => {
    socket.on('data', data => {
        const stringData = data.toString();
        const url = stringData.match(/GET ([\s\S]+) HTTP/)[1];
        const isKeepAlive = /Connection: keep-alive/gi.test(stringData);

        if (url === '/') {
            mainPageSocket = socket;

            socket.setKeepAlive(isKeepAlive);
            socket.write([
                'HTTP/1.1 302 Object Moved',
                'Location: /redirect',
                'Content-Length: 0',
                '',
                ''
            ].join('\r\n'));
        } else if (url === '/redirect') {
            mainPageSocket && mainPageSocket.end();

            if (mainPageSocket !== socket) {
                socket.write([
                    'HTTP/1.1 200 Ok',
                    'Content-Length: 5',
                    '',
                    'Hello'
                ].join('\r\n'));
            }
        }
    });
}).listen(8124, () => console.log('server bound'));
```
The error occurs when I try to open `http://localhost:8124/` through hammerhead.
* The hammerhead creates connection for `http://localhost:8124/` with the `keep-alive` header.
* Server responds with the `302` redirect and `keep alive` connection.
* The hammerhead uses a created connection for `http://localhost:8124/redirect`.
* Server closes the connection.
* The hammerhead socket emits the `Error: socket hang up` error.
* The hammerhead emits the `Failed to find a DNS-record for the resource` error

Hammerhead logs:
```
>>>>>>>>>>>>>>>> request created <<<<<<<<<<<<<<<<
http://localhost:8124
>>>>>>>>>>>>>>>> lookup <<<<<<<<<<<<<<<<
http://localhost:8124
>>>>>>>>>>>>>>>> connect <<<<<<<<<<<<<<<<
http://localhost:8124
>>>>>>>>>>>>>>>> data <<<<<<<<<<<<<<<<
http://localhost:8124
>>>>>>>>>>>>>>>> request created <<<<<<<<<<<<<<<<
http://localhost:8124/redirect
>>>>>>>>>>>>>>>> end <<<<<<<<<<<<<<<<
http://localhost:8124
{ Error: socket hang up
[...]

PR #1394, opened 2017-11-30
fix `Hammerhead emit an error "Failed to find a DNS-record for the resource"` (#1384 part 2)

LavrovArtem:
/cc @miherlosev @churkin


LavrovArtem:
FPR

PR #1396, opened 2017-11-30
drop ie9, 10 related code (part 2)

miherlosev:
@LavrovArtem

### DIFF
--- a/src/client/sandbox/event/message.js
+++ b/src/client/sandbox/event/message.js
@@ -83,14 +83,7 @@ export default class MessageSandbox extends SandboxBase {
             host:     parsedDest.host
         });
 
-        const result = {
-            message:   message,
-            originUrl: originUrl,
-            targetUrl: targetUrl,
-            type:      type
-        };
-
-        return result;
+        return { message, originUrl, targetUrl, type };
     }
 
     _removeInternalMsgFromQueue (sendFunc) {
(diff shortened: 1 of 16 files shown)


src/client/sandbox/event/message.js
```diff
@@ -88,15 +84,13 @@ export default class MessageSandbox extends SandboxBase {
         });
 
         const result = {
-            isStringMessage: typeof message === 'string',
-            message:         message,
-            originUrl:       originUrl,
-            targetUrl:       targetUrl,
-            type:            type
+            message:   message,
```
LavrovArtem:
```js
const result = { message, originUrl, targetUrl, type }
```

test/client/fixtures/sandbox/event/message-test.js
```diff
@@ -103,37 +102,33 @@ asyncTest('message types', function () {
...
-            })
-            .then(function () {
-                return checkValue(void 0);
-            })
-            .then(function () {
-                return checkValue('{a:0}');
-            })
-            .then(start);
-    }
+    checkValue(true)
```
LavrovArtem: `asyncTest` -> `test`
`checkValue(true)` -> `return checkValue(true)`
remove  `.then(start);`

miherlosev:
FPR

PR #1431, opened 2017-12-25
Custom proxy protocol

rudanzona:
@LavrovArtem @miherlosev

Currently the poxy is always created with http protocol. I would like to customize proxy protocol with https and keep its default value as http. Would you help to review the changes and to check if there is any concerns? Thank you.

### DIFF
--- a/src/proxy/index.js
+++ b/src/proxy/index.js
@@ -22,24 +22,25 @@ function parseAsJson (msg) {
     }
 }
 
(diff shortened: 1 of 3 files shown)


rudanzona:
The related PR on testcafe: https://github.com/DevExpress/testcafe/pull/2024

miherlosev:
https://github.com/DevExpress/testcafe/pull/2024#issuecomment-367353430

PR #1461, opened 2018-01-26
Use stored 'hasOwnProperty' for internal purposes (close #1223)

### LINKED ISSUE #1223: Use stored 'hasOwnProperty' for internal purposes

Farfurix:


src/client/sandbox/cookie/index.js
```diff
@@ -28,7 +29,7 @@ export default class CookieSandbox extends SandboxBase {
         const parsedCookieCopy = {};
 
         for (const prop in parsedCookie) {
-            if (parsedCookie.hasOwnProperty(prop))
+            if (nativeMethods.objectHasOwnProperty.call(parsedCookie, prop))
```
miherlosev: Specify valid context for native method call:
`nativeMethods.objectHasOwnProperty.call(window.Object, parsedCookie, prop)`
there and above.

PR #1485, opened 2018-02-09
Refactor the 'files' and 'value' properties of the 'Code instumentaion' subsystem

LavrovArtem:


src/client/sandbox/node/window.js
```diff
@@ -564,6 +569,41 @@ export default class WindowSandbox extends SandboxBase {
...
+                return UploadSandbox.getUploadElementValue(this);
+
+            return nativeMethods.inputValueGetter.call(this);
+        }, function (value) {
+            if (this.type.toLowerCase() === 'file')
+                return windowSandbox.uploadSandbox.setUploadElementValue(this, value);
+
+            nativeMethods.inputValueSetter.call(this, value);
+
+            if (isTextEditableElementAndEditingAllowed(this) && !isShadowUIElement(this))
```
miherlosev: Replace with `if (!isShadowUIElement(this) && isTextEditableElementAndEditingAllowed(this))` because `!isShadowUIElement(this)` is faster.
Here and below.

src/client/utils/dom.js
```diff
@@ -353,8 +353,10 @@ export function isIframeWithoutSrc (iframe) {
     // NOTE: When an iframe has an empty src attribute (<iframe src></iframe>) or has no src attribute (<iframe></iframe>),
     // the iframe.src property is not empty but has different values in different browsers.
     // Its document location is 'about:blank'. Therefore, we should check the src attribute.
+    /*eslint-disable no-restricted-properties*/
```
miherlosev: Can we replace with `nativeMethods.getAttribute.call(iframe, 'src')`?

LavrovArtem:
@testcafe-build-bot \retest

PR #1495, opened 2018-02-14
Refactor the 'src' property of the 'Code instrumentation' subsystem

LavrovArtem:


src/client/sandbox/node/window.js
```diff
@@ -113,6 +113,20 @@ export default class WindowSandbox extends SandboxBase {
         return resolveUrlAsDest(attrValue);
     }
 
+    _overrideUrlAttrDescriptors (attr, constructors) {
```
miherlosev: `_overrideUrlAttrDescriptors (attr, elementConstructors)`

PR #1509, opened 2018-02-26
fix `Call of bind function from postMessage and eval should be instrumented` (close #1498)

### LINKED ISSUE #1498: Call of bind function from postMessage and eval should be instrumented
Client script may contains code as below:
```js 
var w = a.postMessage ? a.postMessage.bind(a) : function() {}
```
And corently it processed:
```js
var w = __get$PostMessage(a) ? a.postMessage.bind(a) : function() {}
```

LavrovArtem:


LavrovArtem:
@testcafe-build-bot \retest

Issue #1379, opened 2017-11-21
Refactor 'Code instrumentation' subsystem


miherlosev:
We will use a native getter from property descriptor.
It allows to drop server and client script processing, improves stack traces for client errors and etc.

It will be done after updating mobile browsers `Android 6.0+`  and `Safari 10+`.

miherlosev:
- [x] `Safari 10`
- [ ] Android 6 (after `testcafe 1.0`)

miherlosev:
Start from `data` property, it allows to fix https://github.com/DevExpress/testcafe-hammerhead/issues/1337

LavrovArtem:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

LavrovArtem:
Done!

PR #1515, opened 2018-02-27
Refactor the 'target' property of the 'Code instrumentation' subsystem

LavrovArtem:
part of #1379


test/client/fixtures/sandbox/node/window-test.js
```diff
@@ -298,6 +298,18 @@ if (window.history.replaceState && window.history.pushState) {
     });
 }
 
+test('overridden "target" attribute getter must return same value as origin getter for elements without attribute', function () {
```
miherlosev: 'an overridden "target" attribute getter should return the same value as origin getter (attribute value is not defined)'

PR #1516, opened 2018-02-27
Refactor 'status' and 'responseURL' xhr properties of the 'Code instrumentation' subsystem and refactor 'status', 'type' and 'url' response properties of fetch sandbox

LavrovArtem:
part of #1379


src/client/sandbox/fetch.js
```diff
@@ -123,35 +93,52 @@ export default class FetchSandbox extends SandboxBase {
     attach (window) {
         super.attach(window, window.document);
 
+        if (!window.fetch)
```
miherlosev: It is wrong.  A page can contain an own fetch implementation.
We should check this condition using the stored in `nativeMethods` value.

```js
if (!nativeMethods.fetch)
     return;
```

src/client/sandbox/xhr.js
```diff
@@ -1,11 +1,13 @@
...
+import { getProxyUrl, parseProxyUrl } from '../utils/url';
 import XHR_HEADERS from '../../request-pipeline/xhr/headers';
 import AUTHORIZATION from '../../request-pipeline/xhr/authorization';
 import { getOriginHeader } from '../utils/destination-location';
 import reEscape from '../../utils/regexp-escape';
 import * as JSON from '../json';
+import overrideDescriptor from '../utils/override-descriptor';
+import { SAME_ORIGIN_CHECK_FAILED_STATUS_CODE } from '../../request-pipeline/xhr/same-origin-policy';
 
 const IS_OPENED_XHR               = 'hammerhead|xhr|is-opened-xhr';
```
miherlosev: Do we still need `XhrSandbox.isOpenedXhr` and `domUtils.isXMLHttpRequest` functions?

src/client/sandbox/xhr.js
```diff
@@ -160,5 +158,20 @@ export default class XhrSandbox extends SandboxBase {
 
             return nativeMethods.xhrSetRequestHeader.call(this, header, value);
         };
+
+        overrideDescriptor(window.XMLHttpRequest.prototype, 'status', function () {
+            const status = nativeMethods.xhrStatusGetter.call(this);
+
+            return this[IS_OPENED_XHR] && status === SAME_ORIGIN_CHECK_FAILED_STATUS_CODE ? 0 : status;
```
miherlosev: I guess the condition `this[IS_OPENED_XHR]` is not necessary.
And constant too.

PR #1521, opened 2018-03-02
fix `document.all property doesn't overridden` (close #1046)

### LINKED ISSUE #1046: document.all property doesn't overridden
The issue is reproducing in playground.
https://www.screencast.com/t/KwMlQHC6NS

LavrovArtem:


src/client/sandbox/native-methods.js
```diff
@@ -39,10 +41,13 @@ class NativeMethods {
         this.documentCreateTouchList     = doc.createTouchList || docProto.createTouchList;
 
         // getters/setters
-        const docPrototype = win.Document.prototype;
+        const docPrototype         = win.Document.prototype;
```
miherlosev: Let's move getters and setters related code to the separate method

src/client/sandbox/native-methods.js
```diff
@@ -39,10 +39,13 @@ class NativeMethods {
...
+        const htmlDocPrototype     = win.HTMLDocument.prototype;
+        const documentAllPropOwner = docPrototype.hasOwnProperty('all') ? docPrototype : htmlDocPrototype;
```
miherlosev: Move to `utils/feature-detection.js` as `featureDetection.getDocumentAllPropertyOwner` function

LavrovArtem:
@testcafe-build-bot \retest

src/client/sandbox/native-methods.js
LavrovArtem: We cannot import the `feature-detection` module into the `nativeMethods` because it is circular dependency.

miherlosev: Sorry. ‘NativeMethods.getDocumentAllPropertyOwner’

PR #1525, opened 2018-03-07
An active input should not be blurred after a call of the focus on a disabled input

LavrovArtem:
related with https://github.com/DevExpress/testcafe/issues/2123


test/client/fixtures/sandbox/event/focus-blur-change-test.js
```diff
@@ -277,6 +277,49 @@ asyncTest('disabled element', function () {
         });
 });
 
+asyncTest('an active input should not be blurred after a call of the focus on a disabled input', function () {
```
helen-dikareva: Please specify gh issue id in title of this test. I think we can event place it in regression section

PR #1529, opened 2018-03-13
Refactor the 'baseURI' property of the 'Code instrumentation' subsystem

LavrovArtem:
part of #1379


src/client/sandbox/native-methods.js
```diff
@@ -307,6 +306,12 @@ class NativeMethods {
         if (cssStyleSheetHrefDescriptor)
             this.cssStyleSheetHrefGetter = cssStyleSheetHrefDescriptor.get;
 
+        const nodeBaseURIDescriptor = win.Object.getOwnPropertyDescriptor(win.Node.prototype, 'baseURI');
+
+        // NOTE: only for IE11
```
miherlosev: It's difficult to understand by your comment what only IE11 has `baseURI` or only `IE11` does not have this property.
Let's change to `// NOTE: IE11 does not have the 'window.Node.prototype.baseURI' property`

LavrovArtem: `// NOTE: IE11 does not have the 'baseURI' property` 
I think it better)))

And I will replace simular comments into current method.

miherlosev: Ok. Rebase this PR

LavrovArtem:
@testcafe-build-bot \retest

PR #1532, opened 2018-03-14
Set a proxy url to an iframe location if iframe is not initialized (close #1531)

### LINKED ISSUE #1531: "Content of the iframe to which you are switching did not load" error; failure to proxy?
As the title says, our test encounters the dreaded “Content of the iframe to which you are switching did not load”. It’s not at all clear why: running with the Chrome devtools open, no requests are failing or timing out. (The devtools show the websocket connection as forever “pending”, but forcing it to fail did not make TestCafe think the frame was done loading.)

If I may speculate, I notice that while the parent frame is proxied, the iframe URL is pointing straight to our server, not a proxy. Maybe this is the issue? If so, maybe it is in turn caused by the fact that our iframe `src` is set by Javascript, not hardcoded in the page? (And if so, maybe a simple `MutationObserver` on `{attributeFilter:['src']}` could help?)

Minimal repro:

```javascript
import { Selector, t } from 'testcafe';

fixture `Login from login page Elk`
	.page `https://web.elk.co/login`;

test('Login with test account', async t => {
	await t
		.switchToMainWindow()
		.expect(Selector('iframe[id=Content]').exists).ok()
		.switchToIframe('#Content')
		.expect(Selector('#email').exists).ok()
		.expect(Selector('#password').exists).ok();
});
```

LavrovArtem:


LavrovArtem:
@testcafe-build-bot \retest

LavrovArtem:
@testcafe-build-bot \retest

PR #1534, opened 2018-03-15
Refactor the 'scripts' property of the 'Code instrumentation' subsystem (close #1229)

### LINKED ISSUE #1229: Self removing scripts entries refactoring
There are four self removing scripts with shared client code: tag, self-executing function and remove scripts parts. Need to create the utility function `createSelfRemovingScript (content)` and use it for these scripts.

LavrovArtem:
part of #1379


miherlosev:
I will merge this PR after next hammerhead version will be released.

Issue #1428, opened 2017-12-20
Support of the new window simulation into iframe for the login via social networks purposes


LavrovArtem:
I've figured out that we have simple way to login via social networks. I've tried to log in to the site from an [issue #1385](https://github.com/DevExpress/testcafe-hammerhead/issues/1385) via a Facebook account and I've [got it](https://www.screencast.com/t/cNGnnhYdtu) in two steps:
* skipped the `x-frame-options` header because it disallowed a login page loading in an iframe.
 https://github.com/DevExpress/testcafe-hammerhead/blob/bf87b68374d88c6f9e409268c31c3422aba8295c/src/request-pipeline/header-transforms.js#L143-L143
* added the `patchAuth` ClientFunction into the test.
```js
const patchAuth = ClientFunction(() => {
    window.open = function (url) {
        var iframe = document.createElement('iframe');

        iframe.style.position = 'fixed';
        iframe.style.left = '200px';
        iframe.style.top = '150px';
        iframe.style.width = '400px';
        iframe.style.height = '300px';
        iframe.style['z-index'] = '99999999';;		
        iframe.src = url;
        iframe.id = 'auth-iframe';

        document.body.appendChild(iframe);
    };
});
```

**Final test:**
```js
fixture `fixture`
    .page `https://www.soudfa.com/signup`;

test('test', async t => {
    await patchAuth();

    await t
        .click('button.facebook')
        .switchToIframe('#auth-iframe')
        .typeText('#email', 'email***')
        .typeText('#pass', 'pass***')
        .click('#u_0_0')
        .wait(30e3);
});
```

neilmatillano:
@LavrovArtem 

Tried this proposal but I am blocked with few errors:

_**const patchAuth = new ClientFunction(() => {**_
Error: TypeScript compilation failed.
Only a void function can be called with the 'new' keyword.
----------
Note that I already have this : import { Selector, Role, ClientFunction } from "testcafe"
So I removed the "new' keyword but I still get the error below.

For this line --  **_window.open = function (url) {_**
Error: TypeScript compilation failed.
/Users/admin/ark-e2e/tests/testCampaign_flow.ts (41, 9): Type '(url: string) => void' is not assignable to type '(url?: string, target?: string, features?: string, replace?: boolean) => Window'.
  Type 'void' is not assignable to type 'Window'.

LavrovArtem:
Fixed type script test:
```ts
[...]

PR #1539, opened 2018-03-20
[WIP]Add special url flag for the authentication via social networks in iframe

LavrovArtem:
related with https://github.com/DevExpress/testcafe-hammerhead/issues/1428

### DIFF
--- a/src/client/dom-processor/client-dom-adapter.js
+++ b/src/client/dom-processor/client-dom-adapter.js
@@ -83,6 +83,10 @@ export default class ClientDomAdapter extends BaseDomAdapter {
         }
     }
 
+    hasAuthIframeParent () {
+        return false;
+    }
+
     attachEventEmitter (domProcessor) {
         const eventEmitter = new EventEmitter();
 
--- a/src/processing/dom/base-dom-adapter.js
+++ b/src/processing/dom/base-dom-adapter.js
@@ -77,6 +77,10 @@ export default class BaseDomAdapter {
         throw new Error('Not implemented');
     }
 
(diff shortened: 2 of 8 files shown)


LavrovArtem:
This PR is unnecassary because the `x-frame-options` header can be replaced via the [Request Hooks](https://github.com/DevExpress/testcafe/pull/2200) feature.

gbatz:
Hi @LavrovArtem 
Could you please share how exactly can the `x-frame-options` header be replaced via Request Hooks?
I implemented the hook, it fires up for my specific request, but still that header seems to be unaffected.
I'm using your workaround for Facebook Login popup. Setting the `'x-frame-options': skip`, in `header-transforms` works good, but I don't want to keep locally that change in `node_modules`.
Thanks!

LavrovArtem:
Hi @gbatz,
For your scenario, you should use the `_onConfigureResponse` method. Unfortunately, it is not documented. The event object has the `setHeader (name, value)` and `removeHeader (name)` methods. For example:

```js
class MyRequestHook extends RequestHook {
    constructor (requestFilterRules, responseEventConfigureOpts) {
        super(requestFilterRules, responseEventConfigureOpts);
    }
    onRequest (event) {
    }
    onResponse (event) {
    }
    _onConfigureResponse (event) {
        event.removeHeader('x-frame-options');
    }
}
```

gbatz:
Thanks @LavrovArtem for your quick response! That worked like a charm!
I can only hope, still, that new window popups will be supported by testcafe in the near future. There's still lots of other pain with this.
Thanks for your hard work!

Issue #659, opened 2016-07-05
Script error for https://www.getresponse.ru/


miherlosev:
see https://testcafe-hhhm.devexpress.com/history?timestamp=1467659469614&siteUrl=http://www.getresponse.ru/

Reproduced in playground.

churkin:
I'm working on it

churkin:
Сookie validation problem:

`"param=value; domain=.ru"` must not be valid for the `www.getresponse.ru`

churkin:
@inikulin  We need to client version of the `tough-cookie` to take into account public suffix list.

LavrovArtem:
This issue is part of #1137

miherlosev:
Close in favor https://github.com/DevExpress/testcafe-hammerhead/issues/1787

PR #1563, opened 2018-04-10
fix `Cookie processing problem related to applications running on 'localhost' with non-default port` (close #1491)

### LINKED ISSUE #1491: Cookie processing problem related to applications running on 'localhost' with non-default port
Steps to reproduce:
Click button 'XHR: Get cookie using ...' and observe 'XSRF-TOKEN=<number>' message in browser console (if run site without proxy).
With proxy, the message does not show. It indicates that cookie sets incorrectly.

[Files to reproduce](https://github.com/DevExpress/testcafe-hammerhead/files/1716312/index.zip)

Farfurix:
### Example
https://github.com/DevExpress/testcafe-hammerhead/issues/1491

### Changes:
1. Refactor default port omitting (the default port is omitted in Chrome, Firefox, Edge, Internet Explorer browsers)
2. localhost/127.0.0.1 (http/https) with default port (80/443) case. If the cookie domain and the current url hostname are equal to localhost/127.0.0.1, we should remove Domain=... form cookie (see tough-cookie: getPublicSuffix('localhost') check)
3. Non-default localhost/127.0.0.1 (http/https) case. Host name check in `_isValidCookie()` instead of same origin check

### Reference:
https://github.com/DevExpress/testcafe-hammerhead/issues/659

### DIFF
--- a/src/client/utils/destination-location.js
+++ b/src/client/utils/destination-location.js
@@ -31,14 +31,11 @@ export function sameOriginCheck (location, checkedUrl, rejectForSubdomains) {
     if (checkedUrl)
         checkedUrl = resolveUrl(checkedUrl);
 
-    // NOTE: exclude the default port from 'same origin check'
-    location = location.replace(sharedUrlUtils.DEFAULT_PORT, '');
-
     return sharedUrlUtils.sameOriginCheck(location, checkedUrl, rejectForSubdomains);
 }
 
 export function resolveUrl (url, doc) {
-    url = sharedUrlUtils.prepareUrl(url);
+    url = sharedUrlUtils.processSpecialChars(url);
 
     /*eslint-disable no-restricted-properties*/
     if (url && url.indexOf('//') === 0)
--- a/src/client/utils/destination-location.js
+++ b/src/client/utils/destination-location.js
@@ -31,6 +31,9 @@ export function sameOriginCheck (location, checkedUrl, rejectForSubdomains) {
     if (checkedUrl)
         checkedUrl = resolveUrl(checkedUrl);
 
+    // NOTE: exclude the default port from 'same origin check'
+    location = location.replace(sharedUrlUtils.DEFAULT_PORT, '');
+
     return sharedUrlUtils.sameOriginCheck(location, checkedUrl, rejectForSubdomains);
 }
 
--- a/src/session/cookies.js
+++ b/src/session/cookies.js
@@ -3,12 +3,19 @@ import BYTES_PER_COOKIE_LIMIT from './cookie-limit';
 import { castArray } from 'lodash';
 import { parseUrl } from '../utils/url';
 
-const LOCALHOST_NAME = 'localhost';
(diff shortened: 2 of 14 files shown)


src/utils/url.js
```diff
@@ -20,6 +20,8 @@ export const REQUEST_DESCRIPTOR_VALUES_SEPARATOR = '!';
 export const TRAILING_SLASH_RE                   = /\/$/;
 export const SPECIAL_PAGES                       = ['about:blank', 'about:error'];
 
+export const DEFAULT_PORT = ':80';
```
miherlosev: Don't duplicate constants.
Create a shared function `omitDefaultPort ()` and use it on client and server.
Use implementation from server side - https://github.com/DevExpress/testcafe-hammerhead/blob/4ea7ec3b78993f48db4863eee8852391dc12f46e/src/request-pipeline/context.js#L82.

src/client/utils/destination-location.js
```diff
@@ -31,6 +31,9 @@ export function sameOriginCheck (location, checkedUrl, rejectForSubdomains) {
     if (checkedUrl)
         checkedUrl = resolveUrl(checkedUrl);
 
+    // NOTE: exclude the default port from 'same origin check'
+    location = location.replace(sharedUrlUtils.DEFAULT_PORT, '');
```
miherlosev: We should omit default port inside `sameOriginCheck` function.

src/utils/url.js
miherlosev: Add tests for `http` and `https` protocol values.

Farfurix:
@testcafe-build-bot \retest

src/session/cookies.js
```diff
@@ -1,6 +1,10 @@
-import { CookieJar } from 'tough-cookie';
+import { CookieJar, Cookie } from 'tough-cookie';
 import BYTES_PER_COOKIE_LIMIT from './cookie-limit';
 import { castArray } from 'lodash';
+import { parseUrl } from '../utils/url';
+
+const LOCALHOST_NAME = 'localhost';
```
miherlosev: `const LOCALHOST_DOMAIN = ...`

src/session/cookies.js
```diff
@@ -14,6 +18,19 @@ export default class Cookies {
             if (cookieStr.length > BYTES_PER_COOKIE_LIMIT)
                 return;
 
+            const cookie = Cookie.parse(cookieStr);
+
+            // NOTE: If cookie.domain and url hostname are equal to localhost/127.0.0.1,
+            // we should remove 'Domain=...' form cookieStr (GH-1491)
+            if (cookie && cookie.domain) {
+                const isCookieDomainLocalhost = cookie.domain === LOCALHOST_NAME || cookie.domain === LOCALHOST_IP;
```
miherlosev: Move to separate function `_hasLocalhostDomain (cookie)`

Farfurix:
@testcafe-build-bot \retest

PR #1564, opened 2018-04-11
textInput event should be supported correctly

AlexKamaev:
connected with https://github.com/DevExpress/testcafe/issues/1956

### DIFF
--- a/src/client/sandbox/event/simulator.js
+++ b/src/client/sandbox/event/simulator.js
@@ -30,6 +30,9 @@ const MOUSE_EVENT_NAME_RE        = /^((mouse\w+)|((dbl)?click)|(contextmenu)|(dr
 const TOUCH_EVENT_NAME_RE        = /^touch\w+$/;
 const FOCUS_IN_OUT_EVENT_NAME_RE = /^focus(in|out)$/;
 
+// NOTE: initTextEvent method required INPUT_METHOD param in IE
+const DOM_INPUT_METHOD_KEYBOARD = 1;
+
 export default class EventSimulator {
     constructor () {
         this.DISPATCHED_EVENT_FLAG = 'hammerhead|dispatched-event';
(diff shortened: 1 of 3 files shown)


src/client/sandbox/event/simulator.js
```diff
@@ -600,6 +600,23 @@ export default class EventSimulator {
         return null;
     }
 
+    _dispatchTextEvent (el, text) {
+        const name = browserUtils.isIE ? 'textinput' : 'textInput';
+
+        let event = null;
```
miherlosev: Specification https://webplatform.github.io/docs/dom/TextEvent/initTextEvent.

Rewrite as
```js
 if (nativeMethods.WindowTextEvent && nativeMethods.documentCreateEvent) {
           const event = nativeMethods.documentCreateEvent.call(document, 'TextEvent');         
           const textEventArgs = {
                 eventType: browserUtils.isIE ? 'textinput' : 'textInput',
                 ...other parameters
           };
          
           event.initTextEvent(textEventArgs.eventType, ....);

           return this._raiseDispatchEvent(el, event);
 }

return null;
```

Also there are few questions about ` event.initTextEvent(name, true, true, null, text, 0, '')` parameters:
* `null` specified for  window on which this event is occurring. Maybe it should be the current window instead of null?
* 0 specified how text is entered. Now it equals `TextEvent.DOM_INPUT_METHOD_UNKNOWN = 0`. Maybe we should use `TextEvent.DOM_INPUT_METHOD_KEYBOARD = 1`?
* '' specified the locale name. As I understand it should be the current locale name - `navigator.language`.

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -674,3 +674,26 @@ test('wrong type of the blur event (GH-947)', function () {
     eventSimulator.blur(domElement);
     ok(raised);
 });
+
```
miherlosev: Add comment that firefox does not support `textInput` properties.
Rename test to `text input` and place it right after the `drag and drop events` test.

src/client/sandbox/event/simulator.js
AlexKamaev: I agree

AlexKamaev:
@testcafe-build-bot \retest

PR #1583, opened 2018-04-24
fix `Hammerhead overrides some attributes, that are used in non-standard way, with "null"` (close #2347)

### LINKED ISSUE #2347: Chrome crashes with STATUS_ACCESS_VIOLATION error message
After setting cookies in `Request Hook`.
Original issue: https://supportcenter.devexpress.com/internal/ticket/details/T898409

intermike:


miherlosev:
Thank you for your contribution. We will publish a new module version on `npm` immediately.

PR #1627, opened 2018-05-18
fix `Function called twice when location keyword used`  (close #1612)

### LINKED ISSUE #1612: Function called twice when location keyword used
Consider test ``page.html`` with following content:

```
<script>
var getMyLocation = function() {
    document.write("getMyLocation();");
    return "1355 Market Street";
};

var updateLocation = function(location) {
    location = getMyLocation();
    return location;
};

updateLocation("399 Fremont Street");
</script>
```

When this page is opened directly in a browser, it outputs:

``getMyLocation();``.

When this page is opened via TestCafe, it outputs:

``getMyLocation();getMyLocation();``.

Function ``getMyLocation()`` is called twice. Depending on it's nature it can turn the application under test into a completely unpredictable state.

LavrovArtem:


LavrovArtem:
@testcafe-build-bot \retest

PR #1638, opened 2018-05-31
fix `Resource with the 304 status code and the 'content-length: x' header does not load` (close #1602)

### LINKED ISSUE #1602: Resource with the 304 status code and the 'content-length: x' header does not load
Server for reproducing:

```js
const net = require('net');

net.createServer(socket => {
    socket.on('data', data => {
        const stringData = data.toString();
        const url = stringData.match(/GET ([\s\S]+) HTTP/)[1];

        if (url === '/') {
            if (!/If\-None\-Match/ig.test(stringData)) {
                socket.write([
                    'HTTP/1.1 200 OK',
                    'ETag: "eyJ2ZXJzaW9uIjowLCJ1cmkiOiJwYWdlcy9ob21lIiwiaGFzaCI6InpuRDdTQT09In0="',
                    'Content-Length: 5',
                    '',
                    'Hello'
                ].join('\r\n'));
            } else {
                socket.write([
                    'HTTP/1.1 304 Not Modified',
                    'Content-Length: 5',
                    '',
                    ''
                ].join('\r\n'));
            }
        } else {
            socket.write('HTTP/1.1 404 Not Found');
            socket.end();
        }
    });
}).listen(8124, () => console.log('server bound'));
```
Proxy works fine if we remove the `Content-Length` header from server above from cached response or if we add body to this response.

Workaround is `disable cache in browser`.

LavrovArtem:


LavrovArtem:
@testcafe-build-bot \retest

src/request-pipeline/index.js
```diff
@@ -96,7 +96,11 @@ const stages = {
                         ctx.onResponseEventDataWithoutBody.push({ rule, opts: configureResponseEvent.opts });
                 });
 
-                ctx.destRes.pipe(ctx.res);
+                if (ctx.contentInfo.isNotModified)
+                    ctx.res.end('');
```
miherlosev: Can we simplify this to `ctx.res.end('');`

I guess we need do same for requests with processed content too.

miherlosev:
@testcafe-build-bot retest

PR #1659, opened 2018-06-22
fix `buttons` property for mouse events and refactor `which` property (closes #1658)

### LINKED ISSUE #1658: MouseEvent does not have corrent `buttons` property in Chrome

AlexKamaev:


test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -475,6 +475,81 @@ if (!browserUtils.isFirefox) {
...
+    domElement.addEventListener('contextmenu', eventHandler);
+    domElement.addEventListener('mousedown', eventHandler);
+    domElement.addEventListener('mouseup', eventHandler);
+    domElement.addEventListener('mouseover', eventHandler);
+    domElement.addEventListener('mouseout', eventHandler);
+    domElement.addEventListener('mouseenter', eventHandler);
+    domElement.addEventListener('mouseleave', eventHandler);
+    domElement.addEventListener('mousemove', eventHandler);
+
+    eventSimulator.click(domElement);
```
helen-dikareva: Let reorganize this test in data driven style, e.g.:
```js
const testData = [
    {
        eventSimulator: () => eventSimulator.dblclick(domElement),
        expectedLog: getExpectedLog('click', eventUtils.BUTTON.left, eventUtils.BUTTONS_PARAMETER.leftButton, 
        eventUtils.WHICH_PARAMETER.leftButton);
    },
...
testData.forEach(function (item) {
     item.eventSimulator();
     deepEqual(actualLog, item.expectedLog);
     clearLogs();
});
}
```
IMHO It's easier to understand and debug. What do you think?

src/client/sandbox/event/simulator.js
```diff
@@ -150,7 +150,7 @@ export default class EventSimulator {
             button:        options.button === void 0 ? eventUtils.BUTTON.left : options.button,
             buttons:       options.buttons === void 0 ? eventUtils.BUTTONS_PARAMETER.leftButton : options.buttons,
             relatedTarget: options.relatedTarget || null,
-            which:         options.which === void 0 ? eventUtils.WHICH_PARAMETER.leftButton : options.which
+            which:         options.which
```
helen-dikareva: Did you check that we send `which` param for each simulated event? (for creation by `nativeMethods.documentCreateEvent`)

test/client/fixtures/sandbox/event/event-simulator-test.js
AlexKamaev: Agree. I'organize the test in this manner

src/client/sandbox/event/simulator.js
AlexKamaev: We don't need to send it for each event, even created via `document.createEvent`. The `which` property is calculated correctly depending on `button` property. We need `which` param for events with not working `which` property, i.e. `mousemove`. So we pass `which` from testcafe to hammerhead on `mousemove`. But, I've mentioned that in any cases we do not have proper set of `button`, `buttons`, `which` prioperties for some events even now. Let's discuss it

AlexKamaev:
@testcafe-build-bot \retest

src/client/sandbox/event/simulator.js
```diff
@@ -181,6 +181,27 @@ export default class EventSimulator {
         return modifiersString;
     }
 
+    static _updateMouseEventButtonProperties (options) {
```
helen-dikareva: we can use fancy `static _updateMouseEventButtonProperties (options = {}) {`

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -210,7 +210,8 @@ test('event with coords (clientX, clientY, screenX, screenY)', function () {
     var screenY = 4;
 
     domElement.onmousedown = function (e) {
-        var coordsAreCorrect = e.clientX === clientX && e.clientY === clientY && e.screenX === screenX && e.screenY === screenY;
+        var coordsAreCorrect = e.clientX === clientX && e.clientY === clientY && e.screenX === screenX && e.screenY ===
```
helen-dikareva: It doesn't really matter, but it looks better if we write
```
var foo = bar1 === bar1 &&
          bar1 === bar2;
```

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -475,6 +476,149 @@ if (!browserUtils.isFirefox) {
...
+    var getExpectedMouseMoveEvents = function (type) {
+        return [{
+            expected: getArgs(type, eventUtils.BUTTON.left, eventUtils.BUTTONS_PARAMETER.noButton,
+                browserUtils.isWebKit ? eventUtils.WHICH_PARAMETER.noButton : eventUtils.WHICH_PARAMETER.leftButton),
+            action: function () {
+                eventSimulator[type](domElement);
+            }
+        }, {
+            expected: getArgs(type, eventUtils.BUTTON.left, eventUtils.BUTTONS_PARAMETER.leftButton,
+                browserUtils.isWebKit ? eventUtils.WHICH_PARAMETER.leftButton : eventUtils.WHICH_PARAMETER.leftButton),
```
helen-dikareva: I don't get this condition

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -475,6 +476,149 @@ if (!browserUtils.isFirefox) {
...
+        return {
+            type:    type,
+            button:  button,
+            buttons: getButtonsProperty(buttons),
+            which:   which
+        };
+    };
+
+    var getExpectedMouseMoveEvents = function (type) {
+        return [{
```
helen-dikareva: I prefer this style:
```js
return [
    {
         foo: bar,
         foo: bar,
    },
    {
         foo: bar,
         foo: bar,
    }
];
```
What do ypu think?

test/client/fixtures/sandbox/event/event-simulator-test.js
AlexKamaev: looks like it was autoformatted accidentally

test/client/fixtures/sandbox/event/event-simulator-test.js
AlexKamaev: it's a mistype

test/client/fixtures/sandbox/event/event-simulator-test.js
AlexKamaev: if it's clearer, ok

AlexKamaev:
https://github.com/DevExpress/testcafe/pull/2581

AlexKamaev:
testcafe tests are passed in https://github.com/DevExpress/testcafe/pull/2581

src/client/sandbox/event/simulator.js
```diff
@@ -697,49 +741,47 @@ export default class EventSimulator {
...
-        return this._simulateEvent(el, 'mouseup', options);
+        return this._simulateEvent(el, 'mouseup', options, {
+            button,
+            buttons: eventUtils.BUTTONS_PARAMETER.noButton
+        });
     }
 
     mouseover (el, options) {
-        return this._simulateEvent(el, 'mouseover', options);
+        return this._simulateEvent(el, 'mouseover', EventSimulator._updateMouseEventButtonProperties(options));
```
miherlosev: It's difficult to read. Move to separate variable.
Also, `EventSimulator._updateMouseEventButtonProperties(options)` call is not clean.
The function calculates `button`, `buttons` and `which` option values. And it's an implementation detail.
I propose to rename it as `_prepareMouseEventOptions (option)`.
So the final call will look like:
```js
mouseover (el, options) {
  const options = EventSimulator._prepareMouseEventOptions(options);
 
  return return this._simulateEvent(el, 'mouseover', options);
}
```

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -475,6 +475,159 @@ if (!browserUtils.isFirefox) {
...
+            {
+                expected: getArgs(type, eventUtils.BUTTON.left, eventUtils.BUTTONS_PARAMETER.rightButton,
+                    browserUtils.isWebKit ? eventUtils.WHICH_PARAMETER.rightButton : eventUtils.WHICH_PARAMETER.leftButton),
+                action: function () {
+                    eventSimulator[type](domElement, { buttons: eventUtils.BUTTONS_PARAMETER.rightButton });
+                }
+            }];
+    };
+
+    const testData = [
```
miherlosev: ` const testCases = `

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -475,6 +475,159 @@ if (!browserUtils.isFirefox) {
...
+    domElement.addEventListener('contextmenu', eventHandler);
+    domElement.addEventListener('mousedown', eventHandler);
+    domElement.addEventListener('mouseup', eventHandler);
+    domElement.addEventListener('mouseover', eventHandler);
+    domElement.addEventListener('mouseout', eventHandler);
+    domElement.addEventListener('mouseenter', eventHandler);
+    domElement.addEventListener('mouseleave', eventHandler);
+    domElement.addEventListener('mousemove', eventHandler);
+
+    testData.forEach(function (item) {
```
miherlosev:
```js
testCases.forEach(function (testCase) {
    testCase.action();
+ 
    deepEqual(actualLog[testCase.expected.type], testCase.expected);
}
```

PR #1689, opened 2018-07-18
fix `Adding a 'load' listener to an image triggers the listener` (close #1688)

### LINKED ISSUE #1688: Adding a 'load' listener to an image triggers the listener
URLs of `src` attributes in `` elements aren't proxied by default. When an event listener for the `load` event is added to an image element, [this code](https://github.com/DevExpress/testcafe-hammerhead/blob/7e94a2cdf3e46ee25642bf60b35502480894219b/src/client/sandbox/node/element.js#L741) replaces the original URL in the `src` attribute with a proxied one. It causes triggering of the event listener, that wouldn't been raised without `testcafe-hammerhead`.

Source: http://builder03/Demos-17.2/ASPxImageAndDataNavigationDemos/ImageGallery/CustomImageViewer.aspx

###### GIF:
dxr

###### index.html:
```html
<html>
    <head>
        <title>Image</title>
    </head>
    <body>
        <img src="https://user-images.githubusercontent.com/4479386/42766437-266c35e2-8923-11e8-94f8-6008d5d651d4.gif">
        <script>
            setTimeout(() => document.querySelector('img').addEventListener('load', () => { alert('"load" raised')}), 1000)
        </script>
    </body>
</html>
```

The alert mustn't be raised, however it is raised with  `testcafe-hammerhead`.

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

PR #1692, opened 2018-07-20
fix `Problem with file upload function.` (related to DevExpress/testcafe#2642)

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

Issue #1702, opened 2018-07-28
Cookie domain validation rule on client side is strange


b12031106:
for example,

if current domain is `https://www.google.com`,
and you try set cookie with domain `.google.com`：
```js
document.cookie = 'a=b; domain=.google.com;';
```
In normal browser, this code is working fine,
I can get this cookie value by access `document.cookie`,

but in Testcafe browser, this code is not working,
and I find out maybe is because client side cookie validation is strange:

https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/cookie/index.js#L85
```js
return !parsedCookie.domain || parsedDestLocation.hostname === parsedCookie.domain;
```

this line means that when you set a cookie with domain attribute,
it should be **exactly the same** with current hostname,
but this behavior is different to a normal browser.

PR #1703, opened 2018-07-28
fix client cookie domain attribute validation issue (#1702)

b12031106:
fix client side cookie validation about domain attribute issue #1702


LavrovArtem:
Hi @b12031106,
Good job, thanks so much for your contribution. I will review this pull request and add some tests.

PR #1708, opened 2018-08-02
fix cookie synchronization regression

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

src/client/sandbox/cookie/index.js
```diff
@@ -221,6 +224,10 @@ export default class CookieSandbox extends SandboxBase {
     }
 
     syncWindowCookie (parsedCookies, win) {
+        // NOTE: Current function can be called before the 'attach' function was called
```
miherlosev: // NOTE: This function can be called before the 'attach' call.

src/client/sandbox/cookie/index.js
```diff
@@ -241,5 +248,16 @@ export default class CookieSandbox extends SandboxBase {
         super.attach(window);
 
         this.windowSync = new WindowSync(window, this, this.messageSandbox);
+
+        for (const { parsedCookies, win, resolve } of this.storedWindowSync) {
```
miherlosev: Let's move this code to the 'syncCookieForNotAttachedWindows' function.

src/client/sandbox/cookie/index.js
```diff
@@ -21,6 +22,8 @@ export default class CookieSandbox extends SandboxBase {
         this.messageSandbox = messageSandbox;
         this.cookieSync     = new CookieSync();
         this.windowSync     = null;
+
+        this.storedWindowSync = [];
```
miherlosev: I propose to rename as `pendingWindowSync`

test/client/fixtures/sandbox/cookie-test.js
```diff
@@ -357,6 +357,28 @@ test('actual cookie in iframe even if a synchronization message does not receive
         });
 });
 
+if (!browserUtils.isFirefox) {
```
miherlosev: Add comment, why we skip the test for Firefox browser

src/client/sandbox/cookie/index.js
LavrovArtem: what about `_processPendingWindowSync`?

LavrovArtem:
@testcafe-build-bot retest

PR #1709, opened 2018-08-06
Use the native "then" and "reject" functions (related to DevExpress/testcafe#2686)

LavrovArtem:


test/client/fixtures/sandbox/fetch-test.js
```diff
@@ -565,6 +565,23 @@ if (window.fetch) {
             strictEqual(origin, 'https://example.com');
             strictEqual(fetchRequestCredentials, 'same-origin');
         });
+
+        test('should use the native "then" function (GH-TC-2686)', function () {
```
miherlosev: Use the PR identificator (`1704`) instead of  `(GH-TC-2686)`

src/client/sandbox/native-methods.js
```diff
@@ -489,6 +489,9 @@ class NativeMethods {
             this.responseUrlGetter    = win.Object.getOwnPropertyDescriptor(win.Response.prototype, 'url').get;
         }
 
+        if (win.Promise)
```
miherlosev: You can remove this condition.
If browser supports `fetch` requests then it means `Promise` is supported too.

Write a separate issue to store `win.Promise.prototy.reject` method and use it in this places: [1](https://github.com/DevExpress/testcafe-hammerhead/pull/1709/files#diff-6a868b3a2dc4967657f6f7dcb7f65f19R116), [2](https://github.com/DevExpress/testcafe-hammerhead/pull/1709/files#diff-6a868b3a2dc4967657f6f7dcb7f65f19R120)

PR #1741, opened 2018-08-28
Refactor transport; Allow to adjust static content cache (closes #1718)

### LINKED ISSUE #1718: Driver can't connect to the server on https://reportserver.devexpress.com/ in IE11
Was reproduced on [Report Server](https://reportserver.devexpress.com/)'s query editor page in IE11 in CI.

Driver sends an XHR to the server just after driver's script is loaded, but the request fails. Server doesn't even log request's payload so I assume it's purely client-side problem.  It is still unclear what the reason of this failure, because IE crashes immediately after opening DevTools. 

Retrying request with a small delay (e.g. 5 seconds) helps. However, I think we should refactor the `testcafe-hammerhead`'s [transport](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/transport.js) module to enable better control over request errors handling.

AndreyBelym:


AndreyBelym:
@testcafe-build-bot retest

src/client/transport.js
```diff
@@ -159,19 +158,28 @@ class Transport {
...
+            this.msgQueue[msg.cmd] = Promise.resolve();
+
+        const isRejectingAllowed = msg.allowRejecting;
 
-            this.msgQueue[msg.cmd].push({
-                msg:      msg,
-                callback: resolve
+        msg.allowRejecting = true;
+
+        const sendMsgPromise = this.msgQueue[msg.cmd]
```
LavrovArtem: May be:
```js
this.msgQueue[msg.cmd]
    .catch(noop)
    .then(() => this.asyncServiceMsg(msg));
```

```js
return this.msgQueue[msg.cmd]
    .catch(err => {
        ...
    });
```

PR #1748, opened 2018-09-05
fix 'It's possible to get shadowUI root using 'for' loop' (close #1747)

### LINKED ISSUE #1747: It's possible to get shadowUI root using 'for' loop
Example to reproduce:
```js
<html>
	<body>
		<h1>Page</h1>
		<script>
		    var hammerhead = window['%hammerhead%'];
		    
			hammerhead && hammerhead.shadowUI.getRoot();
		
			for(var i = 0, childNode; childNode = document.body.childNodes[i]; i++) {
				console.log(childNode);
			}
		</script>
	</body>
</html>
```

miherlosev:
@churkin 

I've changed structure of tests.
Only one new test will be added - https://github.com/DevExpress/testcafe-hammerhead/compare/master...miherlosev:i1747?expand=1#diff-52837961726917bc92aefa76983a3eceR242.

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -111,6 +111,11 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
                     accessors[propName].condition(owner))
                     return accessors[propName].get(owner);
 
+                const propertyValue = owner[propName];
+
+                if (propertyValue && domUtils.isShadowUIElement(propertyValue))
+                    return void 0;
+
                 return owner[propName];
             },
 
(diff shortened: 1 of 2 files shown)


src/client/sandbox/code-instrumentation/properties/index.js
```diff
@@ -111,6 +111,11 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
                     accessors[propName].condition(owner))
                     return accessors[propName].get(owner);
 
+                const propertyValue = owner[propName];
+
+                if (propertyValue && domUtils.isShadowUIElement(propertyValue))
+                    return void 0;
+
                 return owner[propName];
```
churkin: `return propertyValue;` ?

miherlosev:
> What about for...of loop? Will it work?

Yes, it will work. `for..of` loop uses the `length` property internally. I've added a test for this case.

miherlosev:
@testcafe-build-bot retest

PR #1752, opened 2018-09-10
[Request hooks]: Should not hung on large reasponse

miherlosev:
@churkin 

Connected with https://github.com/DevExpress/testcafe/issues/2725


miherlosev:
@testcafe-build-bot retest

PR #1774, opened 2018-10-01
init pointer event in ctor instead of initPointerEvent in Edge

AlexKamaev:


AlexKamaev:
@testcafe-build-bot retest

AlexKamaev:
All tests with `click` or other pointer actions will fail without this fix

helen-dikareva:
They will after updating Edge on browser stack.

PR #1778, opened 2018-10-02
fix `Uncaught TypeError: Cannot convert object to primitive value in console.js` (close #1750)

### LINKED ISSUE #1750: Uncaught TypeError: Cannot convert object to primitive value in console.js
https://github.com/DevExpress/testcafe-hammerhead/blob/bbc833fc911aed34b8cd64fbd0c2df377b0b9449/src/client/sandbox/console.js#L21

### Are you requesting a feature or reporting a bug?
Bug

### What is the current behavior?

On this line, when trying to console.log/.error/etc an array containing a "derived" object, an exception is thrown:

```
Uncaught TypeError: Cannot convert object to primitive value
    at String (<anonymous>)
    at Array.map (native)
```

I'm using AngularJS's http service, and with a (failed) http response, the return value of `response.header()` in the following example code has a header that looks like an Object. However, its `.__proto__` property is missing, and so it doesn't have the `.toString()` method, causing the above error.

```js
$http(options)
.then(function successCallback(response) {
    if (!response) {
        deferred.reject(response);
    } else {
        console.debug('Success call URL', options.url, 'Status:', response.status, 'Data:', response.data, 'Headers:', response.headers(), 'Config:', response.config)
        deferred.resolve(response.data);
    }
}, function errorCallback(response) {
    console.error('Failed call URL', options.url, 'Status:', response.status, 'Error:', response.data, 'Headers:', response.headers(), 'Config:', response.config)
    deferred.reject(response);
});
```

### What is the expected behavior?

The proxied `console.*` methods are able to work as expected without throwing an exception. For a comparison, instead of the `Array.prototype.map.call` line in the code below, simply run `console.log(args)`. 

### How would you reproduce the current behavior (if this is a bug)?

```js
let header1 = {"Date": Date()};
let header2 = {"other_data": "some data"};
let header2.__proto__ = null;
let args = [header1, header2];
Array.prototype.map.call(args, String);
```
#### Provide the test code and the tested page URL (if applicable)

Tested page URL:

Test code

```js

```

### Specify your

* operating system: Windows 10
* testcafe-hammerhead version: 14.2.7
* node.js version: 8.11.3

Farfurix:
https://github.com/DevExpress/testcafe-hammerhead/issues/1750

### Changes
1. Fix overridden console methods (`log`, `warn`, `error`, `info`) logic: try `String(arg)` converting, **otherwise return "object" string**.

### Reference
https://github.com/DevExpress/testcafe-hammerhead/issues/1779

### DIFF
--- a/src/client/sandbox/console.js
+++ b/src/client/sandbox/console.js
@@ -1,7 +1,5 @@
 import SandboxBase from './base';
 import { isCrossDomainWindows } from '../utils/dom';
-import { isConvertibleToString } from '../utils/types';
-import { stringify as stringifyJSON } from '../json';
 
 // NOTE: We should avoid using native object prototype methods,
 // since they can be overridden by the client code. (GH-245)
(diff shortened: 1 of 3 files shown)


src/client/sandbox/console.js
```diff
@@ -14,11 +16,18 @@ export default class ConsoleSandbox extends SandboxBase {
         this.messageSandbox = messageSandbox;
     }
 
+    _convertToStringOrStringify (obj) {
```
miherlosev: _stringify (obj) {
   return obj.toString ? String(obj) : stringifyJSON(obj);
}

PR #1784, opened 2018-10-10
[registerElement] the lifecycle callbacks should not be called twice (fix #695)

### LINKED ISSUE #695: Problem with custom elements
This test will fail because the element creation handler (`createdCallback`) will be called immediately after `document.registerElement`, because an instance of a `new-tag` element has already been created by this time during html processing.

``` javascript
test('', function () {

    iframe.id = 'test8904834';

    //nativeMethods.documentWrite.call(iframe.contentDocument,
    iframe.contentDocument.write(
        '<head>',
        '    <script>',
        '        var newTagProto = Object.create(HTMLElement.prototype);',
        '        newTagProto.createdCallback = function() { if (!document.body) window.isTestFail = true; };',
        '        document.registerElement("new-tag", { prototype: newTagProto });',
        '    <\/script>',
        '</head>',
        '<body>',
        '    <new-tag></new-tag>',
        '</body>'
    );

    ok(!iframe.contentWindow.isTestFail);
});
```

LavrovArtem:


src/client/utils/html.js
```diff
@@ -59,6 +59,7 @@ const STORED_ATTRS_SELECTOR = (() => {
 const SHADOW_UI_ELEMENTS_SELECTOR                    = `[class*="${SHADOW_UI_CLASSNAME.postfix}"]`;
 const HOVER_AND_FOCUS_PSEUDO_CLASS_ELEMENTS_SELECTOR = `[${INTERNAL_ATTRS.hoverPseudoClass}],[${INTERNAL_ATTRS.focusPseudoClass}]`;
 const FAKE_ELEMENTS_SELECTOR                         = `${FAKE_HEAD_TAG_NAME}, ${FAKE_BODY_TAG_NAME}`;
+const PARSING_CONTAINER_FLAG                         = 'hammerhead|parsing-container-flag';
```
miherlosev: `const HTML_PARSER_ELEMENT_FLAG = ...`

src/client/utils/html.js
```diff
@@ -257,6 +262,13 @@ export function dispose () {
     htmlDocument = null;
 }
 
+export function isParsingElement (el) {
```
miherlosev: `isInternalHtmlParserElement (el)`

LavrovArtem:
@testcafe-build-bot retest

Issue #1715, opened 2018-08-09
Internal listeners in iframe are added two and more times


LavrovArtem:
related to https://github.com/DevExpress/testcafe-hammerhead/issues/1649
site for reproducing: https://www.mapion.co.jp/
error in console:
```
hammerhead.js:3453 Uncaught TypeError: _this.resolversMap[message.id] is not a function
    at MessageSandbox.<anonymous> (hammerhead.js:3453)
    at MessageSandbox.emit (hammerhead.js:15980)
    at MessageSandbox._onMessage (hammerhead.js:5671)
    at fastApply (hammerhead.js:16148)
    at onMessageHandler (hammerhead.js:5735)
    at hammerhead.js:5248
```

miherlosev:
Fixed in https://github.com/DevExpress/testcafe-hammerhead/commit/646a6dcba5128f879cda6204d82e85967d6953df

LavrovArtem:
No, in the 646a6dc was fixed the symptom, but not the reason.

PR #1785, opened 2018-10-10
fix the `i.resolversMap[t.id] is not a function` error (related to #1715, fix DevExpress/testcafe#2956)

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

PR #1790, opened 2018-10-12
[RequestHook]: add methods for manipulation with response headers (Closes #1657)

### LINKED ISSUE #1657: [RequestHook]: add methods for manipulation with response headers
https://devexpress.github.io/testcafe/documentation/test-api/intercepting-http-requests/creating-a-custom-http-request-hook.html#the-onresponse-method

`ResponseEvent.setHeader(name, value)`
`ResponseEvent.removeHeader(name)`

NickCis:
The idea es to implement #1657 .

@miherlosev proposed:
> Having discussed this with the team, we can suggest the following API:
> 
> * `removeHeader(name)`
> * `setHeader(name, value)`
> 
> These methods will be defined as methods of the [ConfigureResponseEvent](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/session/events/configure-response-event.js) class.
> 
> The feature implementation can be divided into 2 parts:
> 
> * split the [header-transforms.js
>   ](https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/request-pipeline/header-transforms.js) file into two parts:
>   
>   1. `header-transforms.js` - contains only transformations for request and response headers
>   2. `header-transform-api.js` - contains the `forRequest`, `forResponse` and `transformHeadersCaseToRaw` methods
> * change response headers transformation order.
>   
>   1. apply all header changes from request hooks
>   2. apply header transformation
>   3. write updated headers to the response.


miherlosev:
Sometimes our tests can be unstable due to Saucelabs infrastructure problems.
You can rerun the tests using the @testcafe-build-bot retest.

miherlosev:
@testcafe-build-bot retest

NickCis:
@miherlosev Please review :)

src/request-pipeline/header-transforms/index.js
```diff
@@ -0,0 +1,59 @@
+import {
+    requestTransforms,
+    forcedRequestTransforms,
+    responseTransforms,
+    forcedResponseTransforms
+} from './transforms';
+
+// Transformation routine
+function transformHeaders (srcHeaders, ctx, transformList, forcedTransforms) {
```
NickCis: @miherlosev 
> * Move `transformHeaders` function to the `src/request-pipeline/header-transforms/index.js` file.
> * Rename function parameters as follows: `(headers, ctx, transforms, forcedTransforms)`


I renamed `forced` to `forcedTransforms`. But if i rename `srcHeaders` to `headers` and `transformList` to `transforms` i'll get lint errors: `'...' is already declared in the upper scope no-shadow` because of the `const applyTransforms = function (...)` of line 12.

I left `srcHeaders` and `transformList` parameters just like it was, but what is the preferable approach?

src/request-pipeline/index.js
```diff
@@ -257,16 +266,12 @@ function isDestResBodyMalformed (ctx) {
...
-    const responseInfo           = requestEventInfo.createResponseInfo(ctx);
-    const configureResponseEvent = new ConfigureResponseEvent(rule, ConfigureResponseEventOptions.DEFAULT);
-
-    ctx.session.callRequestEventCallback(REQUEST_EVENT_NAMES.onConfigureResponse, rule, configureResponseEvent);
-
+function callResponseEventCallbackForProcessedRequest (ctx, configureResponseEvent) {
+    const responseInfo         = requestEventInfo.createResponseInfo(ctx);
     const preparedResponseInfo = requestEventInfo.prepareEventData(responseInfo, configureResponseEvent.opts);
-    const responseEvent        = new ResponseEvent(rule, preparedResponseInfo);
+    const responseEvent        = new ResponseEvent(configureResponseEvent._requestFilterRule, preparedResponseInfo);
```
NickCis: @miherlosev 
i've refactored in order to split this function.

I really don't like passing the `configureResponseEvent._requestFilterRule,` because `_requestFilterRule` is like a _private_ property.

What is it preferable?:
1. Leave it like this.
2. This function needs the `configureResponseEvent` object because, in order to call `requestEventInfo.prepareEventData`, i need `configureResponseEvent.opts`. This dependency could be removed as this `opts` are always `ConfigureResponseEventOptions.DEFAULT`. The `callResponseEventCallbackForProcessedRequest` interface could be `callResponseEventCallbackForProcessedRequest (ctx, rule)`, in `sendProxyResponse`, i would iterate through the rules array (instead of the `configureResponseEvents`) and here i would directly use ``ConfigureResponseEventOptions.DEFAULT`` (insead of `configureResponseEvent.opts`).
3. In [ConfigureResponseEvent](https://github.com/DevExpress/testcafe-hammerhead/blob/0dcf519e75e21b6535708df53fac0f8a5e5cc6fb/src/session/events/configure-response-event.js) class rename the `_requestFilterRule` property to `requestFilterRule`.
4. In [ConfigureResponseEvent](https://github.com/DevExpress/testcafe-hammerhead/blob/0dcf519e75e21b6535708df53fac0f8a5e5cc6fb/src/session/events/configure-response-event.js) class add a getter: `get requestFilterRule() { return this._requestFilterRule; }`

src/request-pipeline/header-transforms/index.js
miherlosev: OK, let it as is. This code part requires more complex refactoring. We will do it later.

src/request-pipeline/index.js
miherlosev: I suggest you leave it as is, because we will keep the `_requestFilterRule` property as a private field. 
In the future, we will mark the `_onConfigreResponse` event as public. This is why, the `_requestFilterRule` property should not be exposed as a public property of the `ConfigureResponseEvent`.

miherlosev:
I will merge the PR after the `testcafe@0.23.0` will be released.

NickCis:
@miherlosev should I update testcafe's RequestHook documentation? Or as `_onConfigureResponse` is a private method, is preferable to leave it as an undocumented functionality? (Afaik, this will be mitigated when this method is marked as public)

miherlosev:
> is preferable to leave it as an undocumented functionality?

At present, we will leave it as an undocumented functionality.

PR #1841, opened 2018-11-16
fix 'Hammerhead breaks scripts that save WindowPrototype to a property' (close #1828)

### LINKED ISSUE #1828: Hammerhead breaks scripts that save WindowPrototype to a property
###### page.html
```html
<html>
 <head>
 </head>
 <body>
 <script>
 let prop = 0;
 const arr = [window.constructor.prototype];
 console.log(arr[prop]);
 </script>
 </body>
</html>
```
The problem is in the [isShadowUIElement](https://github.com/DevExpress/testcafe-hammerhead/blob/4eeb86fb9fe90b3a8b9c7b3b39e4c3441cef31c7/src/client/utils/dom.js#L486) function: IE throws an error when accessing the WindowPrototype.window property.

isShadowUIElement is invoked in the property getter function here: https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/code-instrumentation/properties/index.js#L119

miherlosev:
@LavrovArtem

### DIFF
--- a/src/client/sandbox/code-instrumentation/properties/index.js
+++ b/src/client/sandbox/code-instrumentation/properties/index.js
@@ -27,7 +27,7 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
 
     static _safeIsShadowUIElement (el) {
         try {
-            return el && domUtils.isShadowUIElement(el);
+            return domUtils.isShadowUIElement(el);
         }
         catch (e) {
             return false;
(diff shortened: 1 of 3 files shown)


miherlosev:
@testcafe-build-bot

miherlosev:
@LavrovArtem  FPR

miherlosev:
FPR

miherlosev:
ping @LavrovArtem

PR #1919, opened 2019-02-06
fix process styles when stylesheets get updated

DevSide:
Some css-in-js librairies, like styled-components, which rewrite its stylesheets on page load, it is necessary to process again the styles on **CSSStyleSheet.insertRule**.

(for css-in-js user: it fixes css hover, urls and so on)

### DIFF
--- a/src/client/sandbox/style.js
+++ b/src/client/sandbox/style.js
@@ -225,9 +225,10 @@ export default class StyleSandbox extends SandboxBase {
             }
         });
 
-        window.CSSStyleSheet.prototype.insertRule = function (value, index) {
-            value = styleProcessor.process(value, getProxyUrl);
-            nativeMethods.styleSheetsInsertRule.call(this, value, index);
+        window.CSSStyleSheet.prototype.insertRule = function (rule, index) {
+            const newRule = styleProcessor.process(rule, getProxyUrl);
+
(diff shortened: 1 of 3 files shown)


src/client/sandbox/style.js
```diff
@@ -225,6 +225,11 @@ export default class StyleSandbox extends SandboxBase {
             }
         });
 
+        window.CSSStyleSheet.prototype.insertRule = function (value, index) {
+            value = styleProcessor.process(value, getProxyUrl);
+            nativeMethods.styleSheetsInsertRule.call(this, value, index);
```
LavrovArtem: Please add the blank line before the native method call and return the result of the native method execution.
Also, the returned value should be tested :)
https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule#Return_value

DevSide:
Without SSR, i guess we also need to listen to <style> tag injection. I don't know if the hammerhead already does it. I will check this.

Anyone know why the tests fail ? CSSStyleSheet is maybe not available on some browsers.

LavrovArtem:
@testcafe-build-bot retest

LavrovArtem:
@testcafe-build-bot retest

DevSide:
Perfect ! Thanks @LavrovArtem

PR #1929, opened 2019-02-11
Move client to typescript 3

miherlosev:
@LavrovArtem


src/client/utils/feature-detection.ts
```diff
@@ -11,8 +11,8 @@ export const emptyActionAttrFallbacksToTheLocation = nativeMethods.formActionGet
 
 // NOTE: In Chrome, toString(window) equals '[object Window]' and toString(Window.prototype) equals '[object Blob]',
 // this condition is also satisfied for Blob, Document, XMLHttpRequest, etc
-export const instanceAndPrototypeToStringAreEqual = nativeMethods.objectToString.call(window) ===
-                                                    nativeMethods.objectToString.call(Window.prototype);
+export const instanceAndPrototypeToStringAreEqual = nativeMethods.objectToString(window) ===
```
LavrovArtem:
```js
Object.prototype.toString.call(window) // "[object Window]"
Object.prototype.toString(window) // "[object Object]"
```

src/client/sandbox/node/window.ts
```diff
@@ -137,7 +150,7 @@ export default class WindowSandbox extends SandboxBase {
 
     static _formatUnhandledRejectionReason (reason) {
         if (!isPrimitiveType(reason)) {
-            const reasonStr = nativeMethods.objectToString.call(reason);
+            const reasonStr = nativeMethods.objectToString(reason);
```
LavrovArtem: return `call`

PR #1939, opened 2019-02-14
Full Session typings

LavrovArtem:


src/session/events/names.ts
```diff
@@ -1,5 +1,7 @@
-export default {
-    onRequest:           'onRequest',
-    onConfigureResponse: 'onConfigureResponse',
-    onResponse:          'onResponse'
-};
+enum RequestEventNames {
+    onRequest = 'onRequest', // eslint-disable-line no-unused-vars
```
miherlosev: You can omit values and rewrite as
```js
enum RequestEventNames {
    onRequest,
   onConfigureResponse,
   onResponse
}
```

src/session/index.ts
```diff
@@ -9,51 +20,88 @@ import generateUniqueId from '../utils/generate-unique-id';
...
-    id: string;
-    cookies: any;
-    proxy: any;
-    externalProxySettings: any;
-    pageLoadCount: number;
-    pendingStateSnapshot: any;
-    injectable: any;
-    requestEventListeners: any;
-    mocks: any;
+interface Injectable {
```
miherlosev: `interface InjectableResources`

Issue #1935, opened 2019-02-13
Support the spread syntax for object literals (new in ECMAScript 2018)


othree:
Hi

I have a private project using testcafe to do test.
And a strange behavior occurs but I am not able to create a simple sample to reproduce it.
So I am here to asking for some advice.

The issue is:

I have an iframe append to the body after the document is ready.
And I will postMessage to it with some user action.
But inside the iframe's message event, the value of `event.data` is undefined.

I did some investigation. And found there is a MessageSandbox.
Outside the iframe, the message data didn't wrapped by it.
But inside the iframe, the message is extracted by it, so the value becomes `undefined`.
I use breakpoint and a working case to do A/B test to identify this.

I try to build a simple sample try to reproduce this issue.
But I failed, do you have any idea is there any chance this issue could happen.

LavrovArtem:
Thank you for providing this information. However, it is not sufficient to find the cause of the issue and provide a solution. There can be different causes and without having access to your page or code, I cannot give you any advice.

nico29:
Hello,

We are facing a similar issue. We are using testcafe to run tests on our website built using [NextJS](http://nextjs.org), version 8.0.2 (and React version 16.8.2).
Recently our tests would not run and we received the following stacktrace.

```
 1) - Error in fixture.beforeEach hook -
      A JavaScript error occurred on "https://dev.gymlib.com/AXDefoasg/https://dev.gymlib.com/".
      At this moment, TestCafe tracks uncaught JavaScript errors on the page. Try to manually perform the test
      scenario.
      If this error still occurs then it means you site has uncaught JavaScript errors. To disable JavaScript error
      tracking you can turn the --skip-js-errors option on.
      If the error occurs only with TestCafe then it is a bug. Write a new issue about it at:
      "https://github.com/DevExpress/testcafe/issues/new?template=bug-report.md".

      JavaScript error details:
      SyntaxError: Unexpected token )
          at exports.parse (http://192.168.1.21:56369/hammerhead.js:4:16155)
          at Function.t._getMessageData (http://192.168.1.21:56369/hammerhead.js:7:2165)
          at t._onMessage (http://192.168.1.21:56369/hammerhead.js:7:2215)
          at n (http://192.168.1.21:56369/hammerhead.js:13:27178)
          at n (http://192.168.1.21:56369/hammerhead.js:7:3525)
          at http://192.168.1.21:56369/hammerhead.js:6:27272
```
[...]

PR #1952, opened 2019-03-01
prevent error when empty string is processed

nico29:
As stated in my comment in #1935, I think the `JSON.parse` must be surrounded with a `try/catch` block.

Thank you


LavrovArtem:
Thank you for contributing. The #1935 issue has already been fixed so I will close this pull request.

PR #1968, opened 2019-03-21
Optimization of the cookie syncronization between windows (close #1775, fix #1885)

### LINKED ISSUE #1775: Optimization of the cookie syncronization between windows

### LINKED ISSUE #1885: Server crashes if request headers size is more than 8Kb
Looks like the cause of the issue is in the adv which is on the page.
```JS
fixture('fixture')
    .page(`https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_select_options2`)
    test(`test1`, async (t) => {
        await t.wait(15000);
    });
    test(`test2`, async (t) => {
        await t.wait(15000);
    });
```
Also refer to this link: https://nodejs.org/api/errors.html#errors_hpe_header_overflow

The error happens even in the simple nodejs http-server, but only hammerhead crashes

LavrovArtem:


src/client/sandbox/cookie/index.ts
```diff
@@ -21,11 +20,13 @@ export default class CookieSandbox extends SandboxBase {
     messageSandbox: any;
     windowSync: any;
     pendingWindowSync: Array<any>;
+    unloadSandbox: any;
```
miherlosev: Add types wherever necessary

src/client/sandbox/cookie/index.ts
```diff
@@ -173,52 +172,28 @@ export default class CookieSandbox extends SandboxBase {
...
-        return this.windowSync.syncBetweenWindows(actualCookies, win);
     }
 
     attach (window: Window): void {
         super.attach(window);
 
-        this.windowSync = new WindowSync(window, this, this.messageSandbox);
-
-        this._processPendingWindowSync();
+        this.windowSync = new WindowSync(window, this, this.messageSandbox, this.unloadSandbox);
```
miherlosev: It looks like we should create the target `windowSync` in the `CookieSandbox` constructor and start/attach it here.

src/client/sandbox/cookie/window-sync.ts
```diff
@@ -1,113 +1,173 @@
...
 
-            return null;
+            if (this._win !== this._win.top)
+                this._messageSandbox.sendServiceMsg({ id: message.id, cmd: SYNC_COOKIE_DONE_CMD }, source);
+            else
+                this.syncBetweenWindows(message.cookies, source);
+        }
+        else if (message.cmd === SYNC_COOKIE_DONE_CMD) {
+            if (this._resolversMap.has(message.id))
+                this._resolversMap.get(message.id)();
```
miherlosev: It's difficult to read. Let's split in two statements:
```js
            if (this._resolversMap.has(message.id)) {
                const resolver = this._resolversMap.get(message.id);

                resolver();
            }
```

src/client/sandbox/cookie/window-sync.ts
```diff
@@ -1,113 +1,173 @@
...
 
-        const windowsForSync = this._getWindowsForSync(initiator);
-        const syncMessages   = [];
+    private _getWindowsForSync (initiator: Window, currentWindow: Window = this._win.top, windows: Array<Window> = []) {
+        if (currentWindow !== initiator && currentWindow !== this._win.top)
+            windows.push(currentWindow);
 
-        for (const win of windowsForSync) {
-            const cookieSandbox = WindowSync._getCookieSandbox(win);
+        for (let i = 0; i < currentWindow.frames.length; i++)
```
miherlosev: Can we use `for ..of` loop?

LavrovArtem: `error TS2488: Type 'Window' must have a '[Symbol.iterator]()' method that returns an iterator.`
But we have the plugin for `for .. of` statement. I can use `for .. of` with the `@ts-ignore` instruction.

PR #2008, opened 2019-04-25
fix ``setFilesToUpload` not raising `change` event on Linux/Chrome?` (close #2007)

### LINKED ISSUE #2007: `setFilesToUpload` not raising `change` event on Linux/Chrome?
Since testcafe v1.1.3 our tests are failing on Linux/Chrome while they are passing on Linux/Firefox or MacOS/Chrome. Manual testing also works fine on Linux/Chrome.

It seems that on Linux/Chrome the `change` event is fired only for the first call to `setFilesToUpload`. Any subsequent call to `setFilesToUpload` (even with a different `filePath`) don't seem to be firing the `change` event.

It feels this issue could be related or even was introduced with #1844.

Our frontend code looks like this (Vue.js):
```
<input
id="file-upload"
ref="file"
class="uploader"
type="file"
accept=".txt"
@change="handleFileUpload"
>
```

Our test code looks like this:
```js
let files = [
  'file1.txt',
  'file2.txt',
  'file3.txt',
];

for (let i = 0; i < files.length; i += 1) {
    await t.setFilesToUpload(Selector('#file-upload'), files[i]);
}
```

Farfurix:
https://github.com/DevExpress/testcafe-hammerhead/issues/2007

### Changes
1. Call `_shouldRaiseChangeEvent` before `setUploadInfo`.

### DIFF
--- a/test/client/fixtures/sandbox/upload-test.js
+++ b/test/client/fixtures/sandbox/upload-test.js
@@ -544,12 +544,15 @@ asyncTest('change event', function () {
 });
 
 test('change event in the case of another file selection (GH-2007)', function () {
-    expect(2);
+    var changeEventCounter = 0;
+    var fileInput          = document.createElement('input');
 
-    var fileInput = $('<input type="file" name="test" id="id">')[0];
+    fileInput.type = 'file';
+
+    document.body.appendChild(fileInput);
 
     var changeHandler = function () {
-        ok(true);
+        changeEventCounter++;
     };
 
     fileInput.onchange = changeHandler;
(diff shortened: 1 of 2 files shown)


Farfurix:
@testcafe-build-bot retest

test/client/fixtures/sandbox/upload-test.js
```diff
@@ -543,6 +543,23 @@ asyncTest('change event', function () {
     uploadSandbox.doUpload(fileInput, './file.txt');
 });
 
+test('change event in the case of another file selection (GH-2007)', function () {
+    expect(2);
```
miherlosev: In my opinion, using of the counter is more clear.

PR #2015, opened 2019-05-06
Provide capability to pass timeStamp as mouseClick argument (closes #2014)

### LINKED ISSUE #2014: Provide capability to pass timeStamp as mouseClick argument
required for https://github.com/DevExpress/testcafe/issues/3645

AlexKamaev:


AlexKamaev:
https://github.com/DevExpress/testcafe/pull/3772

src/client/sandbox/event/simulator.ts
```diff
@@ -610,6 +610,12 @@ export default class EventSimulator {
             });
         }
 
+        if (timeStamp && !browserUtils.isIE) {
```
miherlosev: I guess this condition should be changed to `if (timeStamp && !browserUtils.isIE11) {`
because MS Edge supports the `timestamp` property (https://developer.mozilla.org/ru/docs/Web/API/Event/timeStamp)
Also, an appropriate test should be changed.

AlexKamaev: All the browsers support the `timestamp` property. However, it is different. In MS browsers, the `timestamp` of the `mouseup` and the `click` events can be not equal to each other. However, in Chrome/Safari/FF the `timestamp` properties of these two events are the same. Thus the `browserUtils.isIE` condition is required for both IE and Edge.

miherlosev: Ok

AlexKamaev:
@testcafe-build-bot retest

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -629,6 +629,18 @@ test('mouse event buttons properties', function () {
     });
 });
 
+if (!browserUtils.isIE) {
+    test('timestamp', function () {
+        var options = { timeStamp: 'timestamp' };
```
miherlosev: `timeStamp` is value in milliseconds.
So, let test value be in milliseconds too.
```js
var options = { timeStamp: 1000 };
```

AlexKamaev:
@testcafe-build-bot retest

AlexKamaev:
@testcafe-build-bot retest

AlexKamaev:
@testcafe-build-bot retest

Issue #1970, opened 2019-03-25
Constructor Worker requires 'new' in firefox


ZhijieZhang:
### What is your Test Scenario?
The main use of my app is to view different types of documents such as pdf, office files, images etc. It uses Web Worker to run some scripts that help processing the document so that it can be rendered/viewed. 

### What is the Current behavior?
When the app is running in the localhost, the Web Worker can be instantiated correctly so that the document can be processed and rendered. However if I'm using testcafe to test the app, the console will error `Constructor Worker requires 'new'`. After tracking the error down a bit, I think it is coming from `window.ts`:
```
if (constructorIsCalledWithoutNewKeyword(this, window.Worker))
    nativeMethods.Worker.apply(this, arguments);
```   

### What is the Expected behavior?
The Web Worker can be instantiated without any error.

### What is your web application and your TestCafe test code?
Please clone the repo from: https://github.com/ZhijieZhang/testcafe_sample, run npm install and then npm start. 
You will see the web application opened in a new tab. 
You can check the TestCafe test code in `test.js` and start the test by `npm run test`.

### Your Environment details:
testcafe version: 1.1.0
node.js version: 9.9.0
command-line arguments: npm run test
browser name and version: Firefox 66.0
platform and version: macOs 10.13.6

miherlosev:
Hi @ZhijieZhang 

Thank you for the detailed description. I've reproduced the problem.

franciscolosardo:
I just verified this is still not working for me. Anyway when the issue was fixed on another branch before the PR was approved and merged, i built my own build from that branch and it was working.
14.6.5 was working in my case.

Farfurix:
@ZhijieZhang 

There is another remaining issue related to this problematic scenario: https://github.com/DevExpress/testcafe-hammerhead/issues/1221#issuecomment-496446388. Once we address it, the original issue will be fixed completely.

#### Firefox results.
`14.6.5` version:
![1970-1](https://user-images.githubusercontent.com/30019338/58684083-8d868780-837f-11e9-8fa8-c615956239f3.PNG)

`master` branch (7d14498):
![1970-2](https://user-images.githubusercontent.com/30019338/58684134-bdce2600-837f-11e9-9e77-5756c6326e5f.PNG)

Farfurix:
I cannot reproduce this issue with testcafe@1.9.4. So, I'll close it.

[`npm run tests`](https://github.com/ZhijieZhang/testcafe_sample) results:
[...]

PR #2028, opened 2019-05-23
fix `calling overridden window.Worker should not cause the "use the 'new'..." error` (part of #1970)

Farfurix:
Part of https://github.com/DevExpress/testcafe-hammerhead/issues/1970

### Changes
1. Fix "constructor is called without the `new` keyword" check ("Constructor Worker requires 'new'..." user case).
2. Fix "calling without `new`" case (process `scriptURL`).


Farfurix:
@testcafe-build-bot retest

Farfurix:
@testcafe-build-bot retest

PR #2038, opened 2019-06-05
change Chrome 75 input[type=email] selection to internalSelection mechanism

AlexKamaev:


AlexKamaev:
https://github.com/DevExpress/testcafe/issues/3865

AlexKamaev:
The cause of the issue is in our Selection mechanism and Chrome behavior of input type=email and input type=number.
When we need to get  selection from input type=email, we cannot do it in normal way, since Chrome input type=email does not have the `selectionStart` and `selectionEnd` property. We used a hack with changing the `type` attribute of input. This hack helped us to restore the `selectionStart` property of input[email]. Starting from Chrome 75 this hack does not work, so we use our internal mechanism which I turn on by the `browserResetInputSelection` option

AlexKamaev:
@testcafe-build-bot retest

PR #2050, opened 2019-06-14
refix typing in email and number types (closes #3865)

AlexKamaev:


AlexKamaev:
I was forced to use internal selection mechanism for all browser except IE and EDGE so support Chrome and Chrome emulation. Since Chrome:emulation changes userAgent it's impossible to detect if we need to use internal selection mechanism or not, so now FF, Chrome and Safari use it.

PR #2057, opened 2019-06-20
fix security alert

miherlosev:
@LavrovArtem


LavrovArtem:
@testcafe-build-bot retest

PR #2086, opened 2019-07-19
fix: remove stripping of www-authenticate header (reverts #974)

jshor:
This reverts PR #974 

The aforementioned PR does not make clear why the stripping of `www-authenticate` is necessary.

Regardless, this header is not uncommonly sent in request responses. We found that the application that we were testing is breaking because of the removal of this header from the response.

Please see [this article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) for more information on the `www-authenticate` header.

### DIFF
--- a/src/request-pipeline/header-transforms/transforms.ts
+++ b/src/request-pipeline/header-transforms/transforms.ts
@@ -151,9 +151,6 @@ export const responseTransforms = {
     'x-content-security-policy-report-only': skip,
     'x-webkit-csp':                          skip,
 
-    // NOTE: Even if we are not able to be authorized, we should prevent showing the native credentials window.
-    'www-authenticate': skip,
-
     // NOTE: We perform CORS checks on our side, so we skip the related headers.
     'access-control-allow-origin': skip,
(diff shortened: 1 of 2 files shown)


LavrovArtem:
Thank you for this pull request. We will review it soon.

jshor:
@LavrovArtem thanks, our team is very much looking forward to getting this fixed.

LavrovArtem:
Hi @jshor, Please accept my apologies for the delayed response. 
As mentioned in the removed comment to this PR:
> Even if we are not able to be authorized, we should prevent showing the native credentials window.

It means that the browser can freeze until the real user types credentials in the native browser window. This test will no longer be automated. Could you provide us with a real scenario which demonstrates the described behavior?

LavrovArtem:
This pull request has been automatically closed because there has been no response to our request for more information from the original author. With only the information that is currently in the pull request, we don't have enough information to take action. Please reach out if you have or find the answers we need so that we can investigate further.

PR #2101, opened 2019-08-02
extractFileToReadStream: fix fileName separators (path.normalize)

Farfurix:
### Changes
1. Normalize a **resource path** in the case of asar (`AsarResource`).
2. Naming: `getFileInAsarNotFoundMessage` -> `getFileInAsarNotFoundErrorMessage`.

### Notes
https://github.com/electron/electron/blob/cec61d010b7f421870d07f00d2aae25bb3b05b61/lib/common/asar.js#L61
```js
archivePath = path.normalize(archivePath)
```
File path sample:
https://github.com/electron/electron/blob/cec61d010b7f421870d07f00d2aae25bb3b05b61/lib/common/asar.js#L70
```js
// "then/internal.file" (with a path separator excluded)
```


AndreyBelym:
@testcafe-build-bot \retest

src/utils/asar.ts
```diff
@@ -47,7 +47,7 @@ export default class Asar {
     parse (fullPath: string) : ParsedPath {
         for (const archivePath of this._archivePaths) {
             if (fullPath.startsWith(archivePath))
-                return { archive: archivePath, fileName: fullPath.replace(archivePath, '.') };
+                return { archive: archivePath, fileName: fullPath.replace(archivePath + '/', '') };
```
LavrovArtem:
```js
fullPath.substr(archivePath.length + 1)
```

PR #2111, opened 2019-08-14
Add return values for all public functions in RequestFilterRule class, ignore error for modules without TypeScript definitions

miherlosev:


src/request-pipeline/request-hooks/request-filter-rule.ts
```diff
@@ -103,24 +103,24 @@ export default class RequestFilterRule {
...
         return this._matchUsingObjectOptions(requestInfo);
     }
 
-    static get ANY () {
+    static get ANY (): RequestFilterRule {
         return new RequestFilterRule(MATCH_ANY_REQUEST_REG_EX);
     }
 
-    static isANY (instance: any) {
+    static isANY (instance: any): boolean {
```
LavrovArtem:
```ts
static isANY (instance?: RequestFilterRule): boolean {
```

src/request-pipeline/request-hooks/request-filter-rule.ts
```diff
@@ -103,24 +103,24 @@ export default class RequestFilterRule {
...
         return !!(instance &&
```
LavrovArtem: !!instance && instance.options.url === MATCH_ANY_REQUEST_REG_EX;

PR #2113, opened 2019-08-19
[WIP]

miherlosev:
Changes:
* fix in the `src/client/sandbox/iframe.ts` file
* update comment - `In Google Chrome, iframes whose src contains html code raise the 'load' event twice.` It's outdated information. Since 73 version, Chrome raised only 1 'load' event. But we still necessary override iframe several times.
* small amount of typings


miherlosev:
@testcafe-build-bot retest

PR #2127, opened 2019-08-29
fix `form.elements.length has the wrong length in IE` (close #2034)

### LINKED ISSUE #2034: form.elements.length has the wrong length in IE
The fix for https://github.com/DevExpress/testcafe-hammerhead/issues/2009 doesn't work sometimes in IE. It is similar to #1763 - sometimes the ShadowUI Collection flag gets lost in IE. There is even a comment in the code: https://github.com/DevExpress/testcafe-hammerhead/blob/cc26d8d2dd1f789ff4ab1c8d7a860826a89c9864/src/client/sandbox/node/window.ts#L756. The condition in the fix for #1763 looks only for `body` and `head` children, but any form with file inputs can cause problems too.

It can't be reproduced consistently. Connecting a debugger to the TestCafe process or opening DevTools in IE helps in reproducing the issue.

The original example:
```js
import { Selector, ClientFunction } from 'testcafe';

fixture `T203631_gridtemplate`
    .page(`http://sarmin-nb-w10/RegressionTestSite/ASPxEditors/ASPxBinaryImage/T203631_gridtemplate.aspx`);


test(`Preview changes on upload`, async t => {
    
    await t.click(Selector('#grid_DXCBtn0'));
    await t
        .click(Selector(".dxbButtonSys[data-args*='CancelEdit']"))
        .click(Selector('#grid_DXCBtn0'));
});
```

The minimal example (it's harder to reproduce the problem with, likely to low resources usage):
##### test.html
```html
<html>
    <head>
        <title>TEST</title>
    </head>
    <html>
        <form id="form">
            <input type="file">
        </form>
        <button id="button">Click me</button>
        <div id="div"></div>
        <script>
            var el = form.elements[1];

            button.onclick = function () { div.textContent = "form.elements.length: " + form.elements.length + " form.elements[1]: " + typeof el};
        </script>
    </html>
</html>
```

##### test.js
```js
import { Selector } from "testcafe";

fixture`form.elements.length`.page('./test.html');

test(`test`, async t => {
    await t
        .setFilesToUpload('input', 'test.js')
        .click('#button')
        .expect(Selector('#div').textContent).eql(`form.elements.length: 1 form.elements[1]: undefined`);
});
```

LavrovArtem:


src/client/sandbox/shadow-ui.ts
```diff
@@ -550,12 +551,32 @@ export default class ShadowUI extends SandboxBase {
...
+
     static isShadowContainer (el) {
         return ShadowUI._hasFlag(el, IS_SHADOW_CONTAINER_FLAG);
     }
 
-    static isShadowContainerCollection (collection) {
-        return ShadowUI._hasFlag(collection, IS_SHADOW_CONTAINER_COLLECTION_FLAG);
+    static isShadowContainerCollection (collection, length?: number) {
+        return isIE && length
+            ? ShadowUI._hasCollectionFlagForIE(collection, length, IS_SHADOW_CONTAINER_COLLECTION_FLAG)
```
miherlosev: Change signature as following `ShadowUI._hasFlagForIE(collection, IS_SHADOW_CONTAINER_COLLECTION_FLAG, , length)`

PR #2171, opened 2019-11-05
fix `An 'import' context is lost when it is used in the evaluated code` (close #2122)

### LINKED ISSUE #2122: An `import` context is lost when it is used in the evaluated code
An imported url will be resolved relative a page url instead of a script url.

```js
var f = new Function('w', 'return import(w);');
f('./123.js')
```

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest-all

LavrovArtem:
@testcafe-build-bot retest

Issue #2206, opened 2020-01-03
"Maximum call stack size exceeded" when create many Proxy instance using same handler


link89:
Hi, we found that there is a chance that when using testcafe to test a mbox based SPA, "Maximum call stack size exceeded" error will occur during case execution.

It turns out the root cause of this error is because hammerhand overwrite `window.Proxy`.
Here is a minimal reproduce script
```javascript
const handler = {
    get: function(obj, prop) {
        return prop in obj ?
            obj[prop] :
            37;
    }
};
let p;
for (let i = 1; i < 100000; i++) {
    p = new Proxy({}, handler);
}
console.log(p.a)
```
Start hammerhand play ground and visit any website in it, then execute the code in console, you will get this error:
 ![微软_Bing_搜索_-_国内版](https://user-images.githubusercontent.com/3314130/71705339-33203e80-2e1a-11ea-8a31-072468d05b4e.png)

Here is the related code
```typescript
        if (window.Proxy) {
            window.Proxy = function (target, handler) {
                if (handler.get) {
                    const storedGet = handler.get;

                    handler.get = function (getterTarget, name, receiver) {
                        if (name === IS_PROXY_OBJECT_INTERNAL_PROP_NAME)
                            return IS_PROXY_OBJECT_INTERNAL_PROP_VALUE;

                        return storedGet.call(this, getterTarget, name, receiver);
                    };
                }

                return new nativeMethods.Proxy(target, handler);
            };
            window.Proxy.toString = () => nativeMethods.Proxy.toString();
            window.Proxy.revocable = nativeMethods.Proxy.revocable;
        }
```

I guess this bug can be fixed by checking if the `handler` has already been overwrite.

Dmitry-Ostashev:
Thank you for your report and PR. We'll discuss it.

PR #2207, opened 2020-01-03
don't rewrite proxy handler if it has been overwritten

link89:
fix https://github.com/DevExpress/testcafe-hammerhead/issues/2206


LavrovArtem:
Thank you for this pull request, we really appreciate it. I think it's ready for the merge.

PR #2231, opened 2020-02-10
fix `A proxied request hangs if a request to the original site emits an error` (close #1961)

### LINKED ISSUE #1961: A proxied request hangs if a request to the original site emits an error
I don't know if this can be reproduced in production, but likely it can.

Take a looks at the error handlers of the DestinationRequest's `req` object:

[_onError at src/request-pipeline/destination-request/index.ts](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/src/request-pipeline/destination-request/index.ts#L171)
```ts
    _onError (err: Error): void {
        if (this._isSocketHangUpErr(err))
            this.emit('socketHangUp');

        else if (requestAgent.shouldRegressHttps(err, this.opts)) {
            requestAgent.regressHttps(this.opts);
            this._send();
        }

        else if (this._isTunnelingErr(err)) {
            //...
        }

        else if (this._isDNSErr(err)) {
            //...  
        }

        else
            this.emit('error', err);
    }
```

[error at src/request-pipeline/utils.ts](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/src/request-pipeline/utils.ts#L95)
```ts
    req.on('error', () => {
        ctx.hasDestReqErr = true;
        resolve();
    });
```

If an error can't be handled by any brach in the `_onError` handler, it will be handled by the generic `error` handler, that just sets the `ctx.hasDestReqErr` and doesn't set the `ctx.goToNextStage` flag. After that the first flag is not handled by any code because without the second request pipeline silently terminates [here](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/src/request-pipeline/index.ts#L210) just after the request is resolved.

How to reproduce:
1. (optional) Add logging statements or breakpoints:
[test/playground/server.js:24](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/test/playground/server.js#L24)
[src/request-pipeline/index.ts:210](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/src/request-pipeline/index.ts#L210)
[src/request-pipeline/utils.ts:95](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/src/request-pipeline/utils.ts#L95)
[src/request-pipeline/destination-request/index.ts:171](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/src/request-pipeline/destination-request/index.ts#L171)

1. `npm i nock`
2. Create `test.js`:
```js
const nock = require('nock');
const playground = require('./test/playground/server');

nock('http://example.com')
    .get('/')
    .replyWithError('this fails')
    .persist();

playground.start();
```
3. `node test.js`
4. When a browser opens, open DevTools.
5. Proxy `example.com`.
6. Request to the proxied page started and never finishes, no error is logged in [handlePageError at test/playground/server.js](https://github.com/DevExpress/testcafe-hammerhead/blob/v14.5.1/test/playground/server.js#L24)

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

src/request-pipeline/utils.ts
```diff
@@ -36,8 +37,14 @@ export function sendRequest (ctx: RequestPipelineContext) {
             resolve();
         });
 
-        req.on('error', () => {
-            ctx.hasDestReqErr = true;
+        req.on('error', err => {
+            // NOTE: Sometimes the underlying socket emits an error event. But if we have a response body,
+            // we can still process such requests. (B234324)
+            if (ctx.isDestResBodyMalformed()) {
+                error(ctx, getText(MESSAGE.destConnectionTerminated, ctx.dest.url, err.toString()));
```
miherlosev: Need to come up with a more generic message, not only about connection terminating.
Something like this
```j
Failed to perform a request for the resource at <a href="{url}">{url}</a> because of error.\n{message}
```

VasilyStrelyaev:
```
Failed to perform a request to the resource at <a href="{url}">{url}</a> because of an error.\n{message}
```

LavrovArtem:
@testcafe-build-bot retest

PR #2235, opened 2020-02-11
element is disabled only if has 'disabled' attribute(closes #2234)

### LINKED ISSUE #2234: Cannot click a button if it is in div which was disabled in JS
You cannot disable a div element. However, you can write `div.disabled=true`.
Site `https://www.afibel.com/fr/` uses this approach. We cannot click on elements which have the `disabled` property. We correct approach in this case will be to check the `disabled` attribute but not property.

This issue will not fix issues related to `disabled` element hierarchy

See: https://github.com/DevExpress/testcafe/issues/4736

AlexKamaev:
check the element `disabled` property is not enough
element is truly disabled only if it has attribute 'disabled'
example:
```JS
var div = document.querySelector('div')
div.disabled = true;
div.hasAttribute('disabled') === false
```
because div is not supposed to be disabled.
Oppositely:
```JS
var button = document.querySelector('button')
button.disabled = true;
button.hasAttribute('disabled') === true
```
because button is supposed to be disabled.

**This works for all browser except IE
In IE even div can be disabled by property.**

### DIFF
--- a/src/client/sandbox/event/simulator.ts
+++ b/src/client/sandbox/event/simulator.ts
@@ -605,7 +605,9 @@ export default class EventSimulator {
     }
 
     _dispatchMouseEvent (el, args, { dataTransfer, timeStamp }: any) {
-        const disabledParent = domUtils.findParent(el, true, node => node.disabled);
+        const disabledParent = domUtils.findParent(el, true, node => {
+            return domUtils.isElementNode(node) && nativeMethods.hasAttribute.call(node, 'disabled');
+        });
 
         if (disabledParent)
             return null;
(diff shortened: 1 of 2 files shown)


src/client/sandbox/event/simulator.ts
```diff
@@ -605,7 +605,9 @@ export default class EventSimulator {
     }
 
     _dispatchMouseEvent (el, args, { dataTransfer, timeStamp }: any) {
-        const disabledParent = domUtils.findParent(el, true, node => node.disabled);
+        const disabledParent = domUtils.findParent(el, true, node => {
+            return domUtils.isElementNode(node) && nativeMethods.hasAttribute.call(node, 'disabled');
```
LavrovArtem: `domUtils.isElementNode(node)` condition is unnecessary

`const disabledParent = domUtils.findParent(el, true, node => nativeMethods.hasAttribute.call(node, 'disabled'));`

AlexKamaev: In the current `findParent` implementation the `document` element can be parent. Document does not have the `hasAttribute` method

PR #2263, opened 2020-03-05
fix: reload in case of iframe unable to load to restart as a workaround

yashLadha:
When the iframe is unable to load, in chrome it starts cancelling the
request. This causes hangup of the testcafe session. Instead when it
gets an error on messaging, it will reload the browser window.

## Purpose

## Approach

## References

## Pre-Merge TODO

### DIFF
--- a/src/client/transport-worker/request.ts
+++ b/src/client/transport-worker/request.ts
@@ -1,10 +1,13 @@
 import { ServiceMessage } from '../../typings/proxy';
 
+const MAX_RETRIES = 3;
+
 interface RequestContext {
     url: string;
     msg: ServiceMessage;
     callback: (err?: string, response?: any) => void;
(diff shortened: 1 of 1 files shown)


LavrovArtem:
Sorry for the late response. Your pull request contains only a workaround. Therefore, we cannot accept it as is. So, please provide a a minimal example so that we can debug it and find the fix.

yashLadha:
@LavrovArtem This w.r.t to multiple chrome hangup issues that are currently opened in testcafe.

LavrovArtem:
@yashLadha We cannot accept this solution as a fix for the problem. Reloading the tested page can lead to clearing its state. Inputs will lose files, text, etc. JavaScript context will be cleared. So I close this pull request. You can reopen or create a new pull request if you get a new idea.

PR #2252, opened 2020-02-19
Improve cookies synchronization mechanism for multi-windows mode (close #2238)

LavrovArtem:

PR #2266, opened 2020-03-06
Fix regression of #2252

miherlosev:


miherlosev:
@testcafe-build-bot retest

PR #2290, opened 2020-04-11
correct server processing the `form` element with target=iframe

AlexKamaev:
The situation:
- we have form with target  = 'framename'
- we submit the form using `form.submit`
- swithingToIframe
- click/expect/any command in iframe leads to error

The cause that iframe window is reloaded after form submit, because of incomplete hh server processing.

**I checked two original cases and they work as expected**

### DIFF
--- a/src/processing/dom/parse5-dom-adapter.ts
+++ b/src/processing/dom/parse5-dom-adapter.ts
@@ -110,7 +110,12 @@ export default class Parse5DomAdapter extends BaseDomAdapter {
         return urlUtils.sameOriginCheck(location, checkedUrl);
     }
 
-    isExistingTarget (): boolean {
-        return false;
+    isExistingTarget (target: string, el: ASTNode): boolean {
+        while (el.parentNode)
+            el = el.parentNode;
+
+        return parse5Utils.findElement(el, e => {
+            return this.getAttr(e, 'name') === target;
+        });
     }
 }
--- a/src/processing/dom/parse5-dom-adapter.ts
+++ b/src/processing/dom/parse5-dom-adapter.ts
@@ -114,8 +114,6 @@ export default class Parse5DomAdapter extends BaseDomAdapter {
         while (el.parentNode)
             el = el.parentNode;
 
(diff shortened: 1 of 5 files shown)


AlexKamaev:
@testcafe-build-bot retest

src/processing/dom/parse5-dom-adapter.ts
```diff
@@ -110,7 +110,12 @@ export default class Parse5DomAdapter extends BaseDomAdapter {
         return urlUtils.sameOriginCheck(location, checkedUrl);
     }
 
-    isExistingTarget (): boolean {
-        return false;
+    isExistingTarget (target: string, el: ASTNode): boolean {
+        while (el.parentNode)
+            el = el.parentNode;
+
+        return parse5Utils.findElement(el, e => {
```
LavrovArtem, line 117:
```js
return parse5Utils.findElement(el, e => this.getAttr(e, 'name') === target);
```

src/utils/parse5.ts
```diff
@@ -75,9 +82,25 @@ export function findElementsByTagNames (root: ASTNode, tagNames: string[]): any
     return elements;
 }
 
+export function findElement (el: ASTNode, predicate: Function) {
```
LavrovArtem, line 85:
```ts
export function findElement (el: ASTNode, predicate: (el: ASTNode) => boolean) {
```

src/utils/parse5.ts
```diff
@@ -75,9 +82,25 @@ export function findElementsByTagNames (root: ASTNode, tagNames: string[]): any
...
+    if (isElementNode(el) && predicate(el))
+        return el;
+
+    if (!el.childNodes)
+        return null;
+
+    for (let i = 0; i < el.childNodes.length; i++) {
```
LavrovArtem, line 92:
```js
for (let child of el.childNodes)
```

AlexKamaev:
FPR
Tests passed. For some reasons Job is displayed as pending, but tests finished

PR #2294, opened 2020-04-16
fix `We should not skip the www-authenticate header if it doesn't show the native credentials window` (close #2286)

### LINKED ISSUE #2286: We should not skip the www-authenticate header if it doesn't show the native credentials window
Server for reproducing:

```js
require('http')
    .createServer((req, res) => {
        if (req.url === '/') {
            res.writeHead(200, { 'content-type': 'text/html' });
            res.end(`
                <script>
                    var xhr = new XMLHttpRequest();

                    xhr.addEventListener('load', function () {
                        document.body.appendChild(document.createTextNode(xhr.getResponseHeader('www-authenticate')));
                    });

                    xhr.open('post', '/auth');
                    xhr.send();
                </script>
            `);
        }
        else if (req.url === '/auth') {
            res.writeHead(401, {
                'content-type': 'text/plain',
                'www-authenticate': 'Bearer realm="Unidos Admin Login"'
            });
            res.end();
        }
        else
            res.end();
    })
    .listen(2021);
```

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

test/client/config-qunit-server-app.js
```diff
@@ -155,6 +155,16 @@ module.exports = function (app) {
         res.json(req.headers);
     });
 
+    app.post('/set-response-headers', function (req, res) {
```
miherlosev, line 158: `app.post('/echo-request-body-in-response-headers')`

PR #2295, opened 2020-04-17
Remove duplicated upload roots

miherlosev:
This change is necessary to avoid displaying duplicates for the `Scanned file paths` section for the updated `Can't find the uploaded file(s)` (https://github.com/devexpress/testcafe/issues/4860#issuecomment-615273760)


miherlosev:
FPR

PR #2296, opened 2020-04-20
Use the built-in functionality for compressing/decompressing of brotli (close #1917)

### LINKED ISSUE #1917: Use the built-in functionality for compressing/decompressing of brotli
https://nodejs.org/api/zlib.html#zlib_zlib_brotlicompress_buffer_options_callback

miherlosev:


adriancable:
Why no await in the brotliCompress line in the updated encoding/index.ts?

miherlosev:
Hi @adriancable 

> Why no await in the brotliCompress line in the updated encoding/index.ts?

The `encodeContent` function returns a `Promise` that will be awaited in the top-level code.
https://github.com/DevExpress/testcafe-hammerhead/blob/18aa93da2a1bc2bd557b1d162eb207fff41f38fa/src/processing/resources/index.ts#L60

PR #2346, opened 2020-06-11
Fire mouse events on custom elements with a 'disabled' attribute

Ogurecher:
## Purpose
_dispatchMouseEvent does not fire an event if the target element has a 'disabled' attribute even if this attribute is not native. This causes issues when using custom HTML elements (for example Ionic in DevExpress/testcafe#5141).

## Approach
I check if the element is an instance of any of the natively disableable HTML element classes and prevent the event from dispatching if it is. (May not be the best solution, open for feedback)

## References
DevExpress/testcafe#5141

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/event/simulator.ts
+++ b/src/client/sandbox/event/simulator.ts
@@ -593,6 +593,16 @@ export default class EventSimulator {
         this._raiseDispatchEvent(el, pointEvent);
     }
 
+    _isDisableableHTMLElement (el): boolean {
+        for (const isDisableableHTMLElementType of DISABLEABLE_HTML_ELEMENT_TYPE_CHECKERS) {
+            if (isDisableableHTMLElementType(el)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     _dispatchMouseRelatedEvents (el, args, userOptions = {}) {
         if (args.type !== 'mouseover' && args.type !== 'mouseenter' && shouldIgnoreMouseEventInsideIframe(el, args.clientX, args.clientY))
             return true;
@@ -620,16 +630,7 @@ export default class EventSimulator {
 
     _dispatchMouseEvent (el, args, { dataTransfer, timeStamp }: any) {
         const disabledParent = domUtils.findParent(el, true, node => {
-            let disableable = false;
-
-            for (const isDisableableHTMLElementType of DISABLEABLE_HTML_ELEMENT_TYPE_CHECKERS) {
-                if (isDisableableHTMLElementType(node)) {
-                    disableable = true;
-                    break;
-                }
-            }
-            
-            return disableable && node.hasAttribute && nativeMethods.hasAttribute.call(node, 'disabled');
+            return this._isDisableableHTMLElement(node) && node.hasAttribute && nativeMethods.hasAttribute.call(node, 'disabled');
         });
 
         if (disabledParent)
--- a/src/client/sandbox/event/simulator.ts
+++ b/src/client/sandbox/event/simulator.ts
@@ -55,7 +55,7 @@ const DISABLEABLE_HTML_ELEMENT_TYPE_CHECKERS = [
     domUtils.isOptionElement,
     domUtils.isSelectElement,
     domUtils.isTextAreaElement
-]
+];
 
 export default class EventSimulator {
     DISPATCHED_EVENT_FLAG: string = 'hammerhead|dispatched-event';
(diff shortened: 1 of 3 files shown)


src/client/sandbox/event/simulator.ts
```diff
@@ -47,6 +47,16 @@ const TOUCH_TO_POINTER_EVENT_TYPE_MAP = {
     touchmove:  'pointermove'
 };
 
+const DISABLEABLE_HTML_ELEMENT_CLASSES = [
+    HTMLButtonElement,
```
AndreyBelym, line 51: Generally, we explicitly specify the global object when using global properties. E.g. `window.HTMLButtonElement`.

AndreyBelym: Or even use `domUtils.isButtonElement` and so on.

src/client/sandbox/event/simulator.ts
```diff
@@ -47,6 +47,16 @@ const TOUCH_TO_POINTER_EVENT_TYPE_MAP = {
...
 
+const DISABLEABLE_HTML_ELEMENT_TYPE_CHECKERS = [
+    domUtils.isButtonElement,
+    domUtils.isFieldSetElement,
+    domUtils.isInputElement,
+    domUtils.isOptGroupElement,
+    domUtils.isOptionElement,
+    domUtils.isSelectElement,
+    domUtils.isTextAreaElement
+]
```
LavrovArtem, line 58: You forgot semicolon, strange that eslint didn't catch it.

src/client/sandbox/event/simulator.ts
```diff
@@ -610,7 +620,14 @@ export default class EventSimulator {
 
     _dispatchMouseEvent (el, args, { dataTransfer, timeStamp }: any) {
         const disabledParent = domUtils.findParent(el, true, node => {
-            return node.hasAttribute && nativeMethods.hasAttribute.call(node, 'disabled');
+            let disableable = false;
+
+            for (const isDisableableHTMLElementType of DISABLEABLE_HTML_ELEMENT_TYPE_CHECKERS) {
+                if (isDisableableHTMLElementType(node))
+                    disableable = true;
```
LavrovArtem, line 627: Please add the `break;` statement.

src/client/sandbox/event/simulator.ts
```diff
@@ -610,7 +620,16 @@ export default class EventSimulator {
...
     _dispatchMouseEvent (el, args, { dataTransfer, timeStamp }: any) {
         const disabledParent = domUtils.findParent(el, true, node => {
-            return node.hasAttribute && nativeMethods.hasAttribute.call(node, 'disabled');
+            let disableable = false;
+
+            for (const isDisableableHTMLElementType of DISABLEABLE_HTML_ELEMENT_TYPE_CHECKERS) {
+                if (isDisableableHTMLElementType(node)) {
+                    disableable = true;
+                    break;
+                }
```
AlexKamaev, line 629: I would prefer to extract the `isDisableableHTMLElementType` logic to the separate method. It's optional to you. If you prefer to leave as is, please add an empty line after assignment.

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -729,6 +729,45 @@ test('mouse events on disabled elements', function () {
...
 
+if (!browserUtils.isIE) {
+    test('mouse events on custom elements with "disabled" property should still work (GH-2346)', function () {
+        function DisabledCustomElement () {
+            return Reflect.construct(HTMLElement, [], this.constructor);
+        }
+
+        DisabledCustomElement.prototype = Object.create(HTMLElement.prototype);
+        DisabledCustomElement.prototype.constructor = DisabledCustomElement;
+        Object.setPrototypeOf(DisabledCustomElement, HTMLElement);
```
AlexKamaev, line 740: we usually align assignments and add empty line after assignement block

test/client/fixtures/sandbox/event/event-simulator-test.js
```diff
@@ -729,6 +729,45 @@ test('mouse events on disabled elements', function () {
...
+
+        customElements.define('disabled-cutsom-element', DisabledCustomElement);
+
+        var disabledCustomElement   = document.createElement('disabled-cutsom-element');
+        var span     = document.createElement('span');
+        var eventLog = [];
```
AlexKamaev, line 746: and here

src/client/sandbox/event/simulator.ts
```diff
@@ -583,6 +593,16 @@ export default class EventSimulator {
         this._raiseDispatchEvent(el, pointEvent);
     }
 
+    _isDisableableHTMLElement (el): boolean {
```
miherlosev, line 596: `_elementCanBeDisabled`

PR #2345, opened 2020-06-10
Destructuring (close #642)

LavrovArtem:

PR #2353, opened 2020-06-19
Fix regression of `Destructuring (close #642) (#2345)` [42b673c]

### LINKED ISSUE #642: The wrong processing of destructuring
If `location` declared inside function or passed as function parameter we should not process it.
It needs for fix https://github.com/DevExpress/testcafe-hammerhead/issues/640#issuecomment-228679853

**Update!** The origin issue was transfered to https://github.com/DevExpress/testcafe-hammerhead/issues/2283

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

Issue #2338, opened 2020-06-03
Use new File() intead of the Blob constructor in FileListWrapper (non-IE11 case)


miherlosev:
https://supportcenter.devexpress.com/internal/ticket/details/T895057

If I perform the actions from the test manually when it works as expected.

With TestCafe Studio (1.3.0) I get the following error:
\<cropped\>

PR #2384, opened 2020-07-14
Fix `Use new File() intead of the Blob constructor in FileListWrapper (non-IE11 case)`

Farfurix:
Fix https://github.com/DevExpress/testcafe-hammerhead/issues/2338, https://github.com/DevExpress/testcafe-hammerhead/issues/1831

### Changes
1. `new File()` was used in `FileListWrapper` (insted of `new Blob()`) (non-IE11 case).
2. `window.File` constructor was overridden (non-IE11 case).
3. `lastModified` was added to wrapped `File` objects (if `File` is constructable) and `UploadStorage`.

### Note
window.File in IE11 is not constructable:
```
> window.File
< [object File]{prototype: FilePrototype {...}}
```

### Example (#2338)
```html
<!doctype html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport"
          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <title>Document</title>
</head>
<body>
<form id="form">
    <input id="file-input" type="file">
    <br>
</form>

<script>
    var fileInput = document.getElementById('file-input');

    fileInput.addEventListener('change', function (e) {
        console.log(e.target.files[0]);

        console.log('name:', e.target.files[0].name);
        console.log('size:', e.target.files[0].size);
        console.log('type:', e.target.files[0].type);
        console.log('lastModifiedDate:', e.target.files[0].lastModifiedDate);
        console.log('lastModified:', e.target.files[0].lastModified);
    })
</script>

<p id="log"></p>
</body>
</html>
```
```js
import { Selector} from 'testcafe';

fixture `Fixture`
    .page `localhost:8080`;

test('upload', async t => {
    await t
        .debug()
        .setFilesToUpload(Selector('input[type="file"]'), 'C://path/1.txt')
        .expect(true).ok()
        .debug();
});
```

### DIFF
--- a/src/client/sandbox/upload/file-list-wrapper.ts
+++ b/src/client/sandbox/upload/file-list-wrapper.ts
@@ -14,8 +14,9 @@ export default class FileListWrapper {
         this.item = index => this[index];
     }
 
-    static _base64ToBlob (base64Data, fileName: string, mimeType: string, sliceSize?: number) {
-        mimeType  = mimeType || '';
+    static _base64ToBlob (base64Data, fileInfo, sliceSize?: number) {
+        const mimeType = fileInfo.info.type || '';
+
         sliceSize = sliceSize || 512;
 
         const byteCharacters = atob(base64Data);
@@ -33,7 +34,7 @@ export default class FileListWrapper {
 
         // NOTE: window.File in IE11 is not constructable.
         return nativeMethods.File
-            ? new File(byteArrays, fileName, { type: mimeType })
+            ? new File(byteArrays, fileInfo.info.name, { type: mimeType, lastModified: fileInfo.info.lastModified })
             : new Blob(byteArrays, { type: mimeType });
     }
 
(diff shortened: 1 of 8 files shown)


src/client/sandbox/upload/file-list-wrapper.ts
```diff
@@ -43,10 +46,14 @@ export default class FileListWrapper {
...
+            wrapper = nativeMethods.File
+                ? new File([fileInfo.blob], fileInfo.info.name, {type: fileInfo.info.type})
+                : new Blob([fileInfo.blob], {type: fileInfo.info.type});
+        }
         else
-            wrapper = FileListWrapper._base64ToBlob(fileInfo.data, fileInfo.info.type);
+            wrapper = FileListWrapper._base64ToBlob(fileInfo.data, fileInfo.info.name, fileInfo.info.type);
 
         wrapper.name             = fileInfo.info.name;
         wrapper.lastModifiedDate = new Date(fileInfo.info.lastModifiedDate);
```
LavrovArtem, line 59: code only for the blob

src/client/sandbox/upload/file-list-wrapper.ts
```diff
@@ -31,7 +31,10 @@ export default class FileListWrapper {
             byteArrays.push(new Uint8Array(byteNumbers));
         }
 
-        return new Blob(byteArrays, { type: mimeType });
+        // NOTE: window.File in IE11 is not constructable.
+        return nativeMethods.File
+            ? new File(byteArrays, fileName, { type: mimeType })
```
LavrovArtem, line 36: add `lastModified` option

Issue #2387, opened 2020-07-02
Support service workers


benmonro:
### What is your Test Scenario?

 I have an app using msw, a mock service worker which provides seamless mocking between node & browser using a service worker.  In the browser, it does this by mocking the fetch calls in a service worker.

When I run my unit tests (in jest) or when I run my app in the browser, they both can share the same mock endpoints.  I am now trying to run the tests in testcafe but running into what i believe could be a conflict between `transport-worker` and `mockServiceWorker` which is causing this problem. The details of the bug can be found here https://github.com/mswjs/msw/issues/259 (I posted it in both projects just in case) 

### What is the Current behavior?

 mockServiceWorker is not able to intercept fetches.

### What is the Expected behavior?

 mockServiceWorker is able to intercept fetches.
 
### What is your web application and your TestCafe test code?

 https://github.com/benmonro/msw-server/tree/feature/testcafe
<details>
<summary>Your website URL (or attach your complete example):</summary>
just running locally in above repo.

</details>

<details>
<summary>Your complete test code (or attach your test files):</summary>

 
```js
 
```
</details>

<details>
<summary>Your complete configuration file (if any):</summary>

```

```
</details>

<details>
<summary>Your complete test report:</summary>

```

```
</details>

<details>
<summary>Screenshots:</summary>

```

```
</details>

### Steps to Reproduce:

1. Go to my website ...
3. Execute this command...
4. See the error...
 
### Your Environment details:
 
* testcafe version:                   
* node.js version:                    
* command-line arguments:  
* browser name and version: 
* platform and version:          
* other:

benmonro:
@AndreyBelym this is going to be a major blocker for us at Walmart.com.  If I purchase a PRO license of testcafe and file a ticket, would that help expedite the implementation of this issue?  The issue you referenced above is 3 years old and has no estimate on when it will be implemented.

AlexKamaev:
@benmonro

We understand that you are interested in this feature a lot. We are working on it. We accept your vote for this feature. However, we take a lot of other factors into account when we set priorities. We appreciate your patience.

benmonro:
@AlexKamaev  @AndreyBelym @miherlosev 

I believe I found a way to make this work quite nicely.  if we can add a configuration option to add the ability ignore a service worker file at this line: https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/node/window.ts#L618 then MSW (and other specified service workers) will behave as expected.

[...]

PR #2388, opened 2020-07-22
feature: allowedServiceWorkers 

benmonro:
## Purpose
Support the ability for service workers to be delivered to the client, without being injected with proxy server information so it can function as a first class service worker as it does in the application itself.

## Approach
use a configuration option for `allowServiceWorkers` with a list of paths that will not use a proxyUrl when they are registered as service workers.

## References
https://github.com/DevExpress/testcafe-hammerhead/issues/2387

## Pre-Merge TODO

### DIFF
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -231,8 +231,8 @@ gulp.task('build',
         'server-scripts',
         gulp.parallel(
             'client-scripts',
-            'templates',
-            'lint'
+            'templates'
+            // 'lint'
         )
     )
 );
--- a/Gulpfile.js
+++ b/Gulpfile.js
@@ -169,19 +169,12 @@ gulp.step('client-scripts-bundle', () => {
 });
 
 gulp.step('client-scripts-processing', () => {
-    const script = gulp.src('./src/client/index.js.wrapper.mustache')
-        .pipe(mustache({ source: fs.readFileSync('./lib/client/hammerhead.js').toString() }))
-        .pipe(rename('hammerhead.js'));
-
-    const bundledScript = script.pipe(clone())
-        .pipe(uglify())
-        .pipe(rename('hammerhead.min.js'));
-
+    
     const bundledTransportWorker = gulp.src('./lib/client/transport-worker.js')
         .pipe(uglify())
         .pipe(rename('transport-worker.min.js'));
 
-    return mergeStreams(script, bundledScript, bundledTransportWorker)
+    return mergeStreams(bundledTransportWorker)
         .pipe(gulp.dest('./lib/client'));
 });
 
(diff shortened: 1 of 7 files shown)


LavrovArtem:
Thank you for your desire to improve the product, but this pull request does not allow testcafe to work correctly with workers. We are working on this issue and will resolve it in the near future.

benmonro:
@LavrovArtem  I'm confused.  This is a configuration option that MAKES the service workers actually work?  Can you please explain how it prevents then from working?

benmonro:
@AlexKamaev @AndreyBelym @miherlosev would someone mind please explaining why this opt-in / configurable feature does not allow testcafe to work with workers?  It specifically only disables the proxy URL for scripts that the user designates.  Other workers/scripts should work just fine.  what am i missing?  If you try to register a service worker right now that intercepts fetch calls, it will not work with testcafe.  My solutions simply registers those service workers such that they can work as they do if you run the app normally.  please help me understand why this PR was closed with no detailed explanation.

Farfurix:
@benmonro 

Here are the reasons why we can't accept your PR:
1. Test cases for the new functionality are missing. This is a strict requirement to have tests for a new functionality.
2. Existing test cases fail. This is a strict requirement not to break existing test cases.
3. Your worker is set in an incorrect scope in your sample. This may sometimes lead to unexpected freezing issues when switching to the next action. This issue is not always reproducible but it can be reproduced when testing your example both with your fix and [our workaround](https://github.com/DevExpress/testcafe-hammerhead/issues/2387#issuecomment-663508835).
4. Again, we can't guarantee that the data you process will not lead to unexpected freezing issues without request proxying in the context of a worker.
5. The flag is public and can be used by other users. We will have to maintain and research all the issues they may encounter

We have prepared a different workaround and shared it with you in this comment: https://github.com/DevExpress/testcafe-hammerhead/issues/2387#issuecomment-663508835

If you still want to use your own workaround, you can fork and build your own version of the `testcafe` and `testcafe-hammerhead` versions and publish them to `npm` under a different name and reference them as follows:
```
"testcafe": "npm:benmonro-testcafe"
```

benmonro:
@Farfurix i was in the process of adding and fixing tests.  I never said the PR was ready to merge, I submitted it early to get feedback. All of the other items seem like feedback as well, not a reason to simply close the PR.  I even asked about number 5 asking if there was a better was to pass configuration down.  

I will try the work around but seems like you would want to have first class support for service workers.  As it stands testcafe doesn't work with them today.

miherlosev:
Hi @benmonro 
 
I should have given you more information about the reason for closing your PR. I apologize for this.  
We are working on the Service Workers support right now. It's a complex task and we don't have plans to turn off the Service workers in the way you suggested in your PR. We cannot give feedback for the feature that is not in our development plans. That is why we closed this PR.
 
If your solution fixes the problem, you can use it. However, I wish to add that it is not suitable for all customers because it just turns off the service workers.

PR #2398, opened 2020-08-03
Add error messages for node header parser errors

### LINKED ISSUE #2188: TestCafe cannot load website with incorrect headers
### What is your Test Scenario?
I'm trying to load https://www.freepeople.com and verify that the page loaded. 
 
When I'm using node version 12 and above, the page never loads and is stuck at the "about:blank" page. If I use node 10.13.0, the page loads successfully. I am able to load other pages on node version 12, so I'm not sure why this website specifically won't load.
 
I am expecting the page to load.
 
### What is your web application and your TestCafe test code?
Using node version 12.13.0, and testcafe v1.6.1
 
```js
 import { Selector, ClientFunction } from 'testcafe';

function getUrl() {
    const clientfunction = ClientFunction(() => document.location.href);
    return clientfunction();
}

fixture('Core Application Tests')

test('Should load the page', async (t) => {
    await t.navigateTo('https://www.freepeople.com');
    await t.wait(3000);
    await t.expect(getUrl()).eql('https://www.freepeople.com');
});
```
</details>

### Steps to Reproduce:

1. Run the above code using node version 10.13.0, and testcafe v1.6.1
2. Notice that the website loads correctly
3. Run the above code using node version 12.13, and testcafe v1.6.1
4. Notice that the website never loads, and is stuck on the about:blank page
 
### Your Environment details:
 
* testcafe version: 1.6.1
* node.js version: 12.13.0
* command-line arguments:  
* browser name and version: Chrome, Firefox
* platform and version:  macOS 10.14. Also able to reproduce using docker

Ogurecher:
## Purpose
Add meaningful error messages with workaround suggestions. Currently adding header overflow and invalid char errors (should we add more?)

## Approach
Replacing the default node error message with our predefined message if present.

## References
Closes #2188

## Pre-Merge TODO


src/messages.ts
```diff
@@ -5,7 +5,11 @@ export const MESSAGE = {
     cantEstablishProxyConnection:     'Failed to connect to the proxy host at <a href="{url}">{url}</a>.',
     cantAuthorizeToProxy:             'Failed to authorize to the proxy at <a href="{url}">{url}</a>.',
     destRequestTimeout:               'Failed to complete a request to <a href="{url}">{url}</a> within the timeout period. The problem may be related to local machine\'s network or firewall settings, server outage, or network problems that make the server inaccessible.',
-    cantReadFile:                     'Failed to read a file at <a href="{url}">{url}</a> because of the error:\n\n{message}'
+    cantReadFile:                     'Failed to read a file at <a href="{url}">{url}</a> because of the error:\n\n{message}',
+    nodeError: {
+        'HPE_HEADER_OVERFLOW': 'The request header\'s size exceeds the set limit.\nIt causes an internal Node.js error on parsing this header.\nTo fix the problem, you need to specify the maximum header size via the NODE_OPTIONS=\'--max-http-header-size=...\' environment variable.',
```
miherlosev, line 10: I think we need to add examples for programmatic and command-line interfaces for both error messages.
Also, if it's possible, try to retrieve the problematic header name from the error and use it in the error message.

test/server/proxy-test.js
```diff
@@ -4116,5 +4081,101 @@ describe('Proxy', () => {
                 .then(() => new Promise(resolve => setTimeout(resolve, 2000)))
                 .then(() => expect(hasPageError).to.be.false);
         });
+
+        describe('Should respond with an error when destination server emits an error', () => {
```
miherlosev, line 4085:
```js
let storedHttpRequest = null;

beforeEach(() => {
    storedHttpRequest = http.request;
});

afterEach(() => {
  http.request = storedHttpRequest;
});

```

Also, it allowing to change the `mockRequest` function signature to `function mockRequest (url, error)`

src/messages.ts
```diff
@@ -5,14 +5,26 @@ export const MESSAGE = {
...
-export function getText (template: string, url: string, message?: string): string {
-    let errorStr = template.replace(/\{url\}/g, url);
+class MessageParameters {
+    url:                        string;
+    message?:                   string;
+    headersSize?:               string;
+    recommendedMaxHeadersSize?: string;
+}
+
+export function getText (template: string, parameters: MessageParameters): string {
```
miherlosev, line 22: We don't need a separate class here. Use the `Dictionary<unknown>` type instead.

src/request-pipeline/destination-request/index.ts
```diff
@@ -213,4 +213,8 @@ export default class DestinationRequest extends EventEmitter implements Destinat
         else
             this.emit('error', err);
     }
+
+    getHeadersSize (bitsPerByte: number = 16): number {
```
miherlosev, line 217: We should rewrite it as follows.

```js
// http://www.ecma-international.org/ecma-262/5.1/#sec-4.3.16
// One character of string represents 2 bytes.
const CHARACTER_SIZE_IN_BYTES = 2;

....
/* get request headers size in bytes*/
getHeadersSize (): number {
   return JSON.stringify(this.opts.rawHeaders).length * CHARACTER_SIZE_IN_BYTES;
} 
```

src/request-pipeline/utils.ts
```diff
@@ -17,6 +17,13 @@ import { PassThrough } from 'stream';
 import { getText, MESSAGE } from '../messages';
 import logger from '../utils/logger';
 
+const NODE_MAX_HEADER_SIZE = 81920;
```
miherlosev, line 20: Why does we use this value? Please, add a link to specification, comment or something else.

src/request-pipeline/utils.ts
```diff
@@ -17,6 +17,13 @@ import { PassThrough } from 'stream';
...
+
+function recommendMaxHeadersSize (currentHeaderSize: number, headerSizeMultiplier: number = 2, headerSizePrecision: number = 2): number {
```
miherlosev, line 22:
```js
/* Calculates the HTTP header size that customer should specify via 
--max-http-header-size option so that the proxy can process its site
 https://nodejs.org/api/cli.html#cli_max_http_header_size_size
*/
function getRecommendedHeaderSize(currentHeaderSize: number): number {}
```

src/request-pipeline/utils.ts
miherlosev: Calculate all sizes in bytes.

src/request-pipeline/utils.ts
```diff
@@ -42,7 +49,12 @@ export function sendRequest (ctx: RequestPipelineContext) {
             // NOTE: Sometimes the underlying socket emits an error event. But if we have a response body,
             // we can still process such requests. (B234324)
             if (!ctx.isDestResReadableEnded)
-                error(ctx, getText(MESSAGE.destConnectionTerminated, ctx.dest.url, err.toString()));
+                error(ctx, getText(MESSAGE.destConnectionTerminated, {
+                    url:                       ctx.dest.url,
+                    message:                   MESSAGE.nodeError[err.code] || err.toString(),
+                    headersSize:               req instanceof DestinationRequest ? req.getHeadersSize().toString() : '',
```
miherlosev, line 55: `toString()` is redurant here.

```js
req instanceof DestinationRequest ? req.getHeadersSize() : ''
```

src/messages.ts
```diff
@@ -5,14 +5,26 @@ export const MESSAGE = {
...
+        'HPE_INVALID_HEADER_TOKEN': 'The request contains a header that doesn\'t comply with the specification <a href="https://tools.ietf.org/html/rfc7230#section-3.2.4">https://tools.ietf.org/html/rfc7230#section-3.2.4</a>.\nIt causes an internal Node.js error on parsing this header.\nTo fix the problem, you need to add the \'--insecure-http-parser\' flag to the \'NODE_OPTIONS\' environment variable:\n\nmacOS, Linux (bash, zsh)\nexport NODE_OPTIONS=\'--insecure-http-parser\'\n\nWindows (powershell)\n$env:NODE_OPTIONS=\'--insecure-http-parser\'\n\nWindows (cmd)\nset NODE_OPTIONS=\'--insecure-http-parser\'\n\nand then start your tests.'
```
AndreyBelym, line 11: Can we try to extract the problematic header from `rawHeaders` (similarly as we do for the header size), or it will be too hard?

src/request-pipeline/utils.ts
Ogurecher: This limitation turned out to be unnecessary. Node did previously limit header sizes to 80 Kb but gave no apparent reason as to why.

src/request-pipeline/utils.ts
```diff
@@ -17,6 +17,22 @@ import { PassThrough } from 'stream';
...
+// An empty line that indicates the end of the header section
+// https://tools.ietf.org/html/rfc7230#section-3
+const HTTP_SEPARATOR = '\r\n\r\n';
+
+// Used to calculate the recommended maximum header size
+// See recommendMaxHeaderSize() below
+const HEADER_SIZE_MULTIPLIER = 2;
+const HEADER_SIZE_PRECISION  = 2;
+
+// Calculates the HTTP header size that customer should specify via the
```
miherlosev, line 29: `Calculates the HTTP header size in bytes that customer...`

src/request-pipeline/utils.ts
```diff
@@ -41,8 +57,16 @@ export function sendRequest (ctx: RequestPipelineContext) {
         req.on('error', err => {
             // NOTE: Sometimes the underlying socket emits an error event. But if we have a response body,
             // we can still process such requests. (B234324)
-            if (!ctx.isDestResReadableEnded)
-                error(ctx, getText(MESSAGE.destConnectionTerminated, ctx.dest.url, err.toString()));
+            if (!ctx.isDestResReadableEnded) {
+                const headerSize = err.rawPacket ? err.rawPacket.asciiSlice().split(HTTP_SEPARATOR)[0].length : 0;
```
miherlosev, line 61: `const headerSizeInCharacters = ...`

src/request-pipeline/utils.ts
```diff
@@ -17,6 +17,22 @@ import { PassThrough } from 'stream';
...
+// Used to calculate the recommended maximum header size
+// See recommendMaxHeaderSize() below
+const HEADER_SIZE_MULTIPLIER = 2;
+const HEADER_SIZE_PRECISION  = 2;
+
+// Calculates the HTTP header size that customer should specify via the
+// --max-http-header-size Node option so that the proxy can process the site
+// https://nodejs.org/api/cli.html#cli_max_http_header_size_size
+function recommendMaxHeaderSize (currentHeaderSize: number): number {
+    return Number((currentHeaderSize * HEADER_SIZE_MULTIPLIER).toPrecision(HEADER_SIZE_PRECISION));
```
miherlosev, line 33: Can we use the expression `(currentHeaderSize * HEADER_SIZE_MULTIPLIER).toFixed()` here?

Ogurecher: [toPrecision()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision) converts the recommended header size to the exponential notation and rounds it to HEADER_SIZE_PRECISION significant digits. (i.e. (8211 * 2).toPrecision(2) -> 16000)

[toFixed()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) rounds the number to HEADER_SIZE_PRECISION significant digits after decimal point. (i.e. (8211 * 2).toFixed(2) -> 16422.00)

miherlosev: Ок. I understand it. Let's add the following comment into the `getRecommendMaxHeaderSize` function.
```js
// Example: 
// (8211 * 2).toPrecision(2) -> 16 * 10^3 -> 16000
```

src/request-pipeline/http-header-parser.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    return formatInvalidCharacters(invalidCharList);
+}
+
+function getInvalidCharacters (name: string, body: string): InvalidCharactersRecord[] {
+    const invalidCharList = [];
+
+    for (let i = 0; i < name.length; i++) {
+        if (name[i].charCodeAt(0) < HEADER_NAME_VALID_CHAR_CODE_RANGE.min || name[i].charCodeAt(0) > HEADER_NAME_VALID_CHAR_CODE_RANGE.max) {
```
AndreyBelym, line 32: I feel we can extract to a function

src/request-pipeline/http-header-parser.ts
```diff
@@ -0,0 +1,64 @@
...
+            invalidCharList.push({
+                name: name,
+                location: HEADER_INVALID_CHAR_LOCATIONS.name,
```
AndreyBelym, line 35: We try to allign it in the following way:
```js
{
   prop:     '',
   longProp: ''
}
```

src/request-pipeline/utils.ts
```diff
@@ -41,8 +60,18 @@ export function sendRequest (ctx: RequestPipelineContext) {
         req.on('error', err => {
             // NOTE: Sometimes the underlying socket emits an error event. But if we have a response body,
             // we can still process such requests. (B234324)
-            if (!ctx.isDestResReadableEnded)
-                error(ctx, getText(MESSAGE.destConnectionTerminated, ctx.dest.url, err.toString()));
+            if (!ctx.isDestResReadableEnded) {
+                const rawHeaders = err.rawPacket ? err.rawPacket.asciiSlice().split(HTTP_BODY_SEPARATOR)[0].split('\n').splice(1).join('\n') : '';
```
miherlosev, line 64: `const rawHeadersStr = ...`

src/request-pipeline/utils.ts
```diff
@@ -41,8 +60,18 @@ export function sendRequest (ctx: RequestPipelineContext) {
...
+                const headerSize = rawHeaders.length;
+
+                error(ctx, getText(MESSAGE.destConnectionTerminated, {
+                    url:                      ctx.dest.url,
+                    message:                  MESSAGE.nodeError[err.code] || err.toString(),
+                    headerSize:               headerSize,
+                    recommendedMaxHeaderSize: recommendMaxHeaderSize(headerSize),
```
miherlosev, line 71: `getRecommendMaxHeaderSize(headerSize)`

src/request-pipeline/utils.ts
```diff
@@ -16,6 +16,25 @@ import { toReadableStream } from '../utils/buffer';
...
+const HEADER_SIZE_MULTIPLIER            = 2;
+const HEADER_SIZE_CALCULATION_PRECISION = 2;
+
+// Calculates the HTTP header size in bytes that a customer should specify via the
+// --max-http-header-size Node option so that the proxy can process the site
+// https://nodejs.org/api/cli.html#cli_max_http_header_size_size
+// Example: 
+// (8211 * 2).toPrecision(2) -> 16 * 10^3 -> 16000
+function recommendMaxHeaderSize (currentHeaderSize: number): number {
+    return Number((currentHeaderSize * HEADER_SIZE_MULTIPLIER).toPrecision(HEADER_SIZE_CALCULATION_PRECISION));
```
miherlosev, line 36: Can we remove the `Number` constructor call here?
```js
return (currentHeaderSize * HEADER_SIZE_MULTIPLIER).toPrecision(HEADER_SIZE_CALCULATION_PRECISION);
```

src/proxy/index.ts
```diff
@@ -18,6 +18,11 @@ import logger from '../utils/logger';
 
 const SESSION_IS_NOT_OPENED_ERR = 'Session is not opened in proxy';
 
+// Max header size for incoming HTTP requests
+// Set to 80 KB as it was the original limit
+// (before the change to 8 KB)
+const MAX_REQUEST_HEADER_SIZE = 80 * 1024;
```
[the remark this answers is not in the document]
miherlosev: Also, let's export this constant as a static property of the `Proxy` class. It allowing use it for `hammerhead-playground` creation.
```js
Proxy.MAX_REQUEST_HEADER_SIZE
```

test/playground/server.js
```diff
@@ -27,7 +27,7 @@ function prepareUrl (url) {
 
 exports.start = sslOptions => {
     const app       = express();
-    const appServer = http.createServer(app);
+    const appServer = http.createServer({ maxHeaderSize: 80 * 1024 }, app);
```
miherlosev, line 30:
```js
const appServer = http.createServer({ maxHeaderSize: Proxy.MAX_REQUEST_HEADER_SIZE }, app);
```

src/proxy/index.ts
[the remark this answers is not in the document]
AndreyBelym: Give me a moment, I add it

src/request-pipeline/utils.ts
Ogurecher: [toPrecision](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toPrecision) returns a string in exponential notation in this case

src/proxy/index.ts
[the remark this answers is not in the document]
AndreyBelym: https://github.com/nodejs/node/blob/186035243fad247e3955fa0c202987cae99e82db/deps/http_parser/http_parser.h#L63

Change that decreased the header size

https://github.com/nodejs/node/commit/186035243fad247e3955fa0c202987cae99e82db#diff-1d0d420098503156cddb601e523b82e7R59

src/request-pipeline/utils.ts
miherlosev: Ok

PR #2409, opened 2020-08-18
fix `Override Request('/').url` (close #2140)

### LINKED ISSUE #2140: Override Request('/').url
### What is your Scenario?

I am running into problems with CSRF protection when testing an app, the token from the cookie is not send to the backend at all. It seems the token is not send because the URL is changed by TestCafé and thus security mechanisms catch in, since everything works fine outside of TestCafé.

The problem is that `window.location.origin` and `new Request('/').url` return the same thing when visiting the page in a browser but the second one is something like `http://1.2.3.4:12345/8RWZTuU_s/https://bxt.github.io/testcafe-2140/` when run via TestCafé.

Here's a failing test case do demonstrate the root of the problem:

```js
import {Selector} from 'testcafe';

fixture`testcafe 2140`.page`https://bxt.github.io/testcafe-2140/`;

test(`has the correct value in Request('/').url on GitHub`, async t => {
  await t
  .expect(Selector('*').withText('"rootRequestUrl": "https://bxt.github.io/"').visible)
  .ok();
});
```

You can find a repo containing this test, the [web page](https://bxt.github.io/testcafe-2140/)'s HTML and a way to run it locally here: https://github.com/bxt/testcafe-2140

<details>
<summary>(Original issue description hidden here because this was initially about Cookie mechanism documentation)</summary>

### What is your Scenario?

I am running into problems with CSRF protection when testing an app, the token from the cookie is not send to the backend at all. It seems the token is not send because the URL is changed by TestCafé and thus security mechanisms catch in, since everything works fine outside of TestCafé.

I could probably figure it out, however, there is no way to see what cookies there are and how they are supposed to work. I can pause with `t.debug();` but in the browser dev tools I do not even see the cookies saved for login.

The "big cookie issue" https://github.com/DevExpress/testcafe-hammerhead/issues/1137 does not have any description, and neither do e.g. https://github.com/DevExpress/testcafe-hammerhead/pull/1633 nor https://github.com/DevExpress/testcafe-hammerhead/issues/2074 it's always:

> No description provided.

So it's hard to figure out what you are doing and why, and how everything is supposed to work.

Basically it's impossible to debug tests that fail due to cookie problems.

### What are you suggesting?

Please write down somewhere how Cookies work via hammerhead and maybe provide a way to see which cookies are present.

### What alternatives have you considered?

Cypress.io?

### Additional context

Unfortunately the stuff is not publicly accessible, so I can not share the exact scenario, but I assume other people run into similar issues. 

</details>

LavrovArtem:


src/client/sandbox/fetch.ts
```diff
@@ -171,6 +171,15 @@ export default class FetchSandbox extends SandboxBase {
         };
         window.Request.prototype = nativeMethods.Request.prototype;
 
+        overrideDescriptor(window.Request.prototype, 'url', {
+            getter: function () {
+                const responseUrl       = nativeMethods.requestUrlGetter.call(this);
```
miherlosev, line 176:
```js
const requestUrl       =
const parsedRequestUrl = 
```

Farfurix (approved):

miherlosev (approved):

Farfurix (approved):

LavrovArtem:
@testcafe-build-bot retest

PR #2410, opened 2020-08-20
Overrode appendData() to insert hammerhead styles

Ogurecher:
## Purpose
Hover did not work on styled-components Devexpress/testcafe#3830 because `window.CharacterData.prototype.appendData` was not overriden to inject hammerhead styles

## Approach
`window.CharacterData.prototype.appendData` now calls `ElementSandbox._processTextNodeContent()` if the element has a parent.

## References
Closes Devexpress/testcafe#3830

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/node/element.ts
+++ b/src/client/sandbox/node/element.ts
@@ -454,6 +454,12 @@ export default class ElementSandbox extends SandboxBase {
         const sandbox = this;
 
         this.overriddenMethods = {
+            appendData (...args) {
+                this.data += args[0];
+
+                ElementSandbox._processTextNodeContent(this, nativeMethods.nodeParentNodeGetter.call(this));
+            },
+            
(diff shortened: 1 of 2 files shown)


src/client/sandbox/node/element.ts
```diff
@@ -454,6 +454,13 @@ export default class ElementSandbox extends SandboxBase {
         const sandbox = this;
 
         this.overriddenMethods = {
+            appendData (...args) {
```
AlexKamaev: as far as I understand we do not need args array here since we use only args[0] parameter, so I think we can replace `...args` with `text`

AlexKamaev (changes requested):

AlexKamaev (approved):

AndreyBelym (approved):

PR #2416, opened 2020-08-26
Fix `Testcafe hammerhead processing transforms break JavaScript without semicolons as line endings` (close #2392)

### LINKED ISSUE #2392: Testcafe hammerhead processing transforms break JavaScript  without semicolons as line endings
### What is your Scenario?
JavaScript code written without semicolons at line endings, if they contain special variable names like location, breaks after Hammerhead processing script transforms them.

Hammerhead processing script wraps, set assignments like 'location', putting an IIFE paranthesis at the start. However without a semicolon at the previous line, JavaScript interpreter assumes this parentheses is part of the function invocation from the previous line and errors while trying to run a function which does not exist
 
Below is a simple code example that fails

Although this code works perfectly fine in a browser, it fails when run with TestCafe

```
function test() {
  var location
  var b, a

  a = b // this fails whilst running test, without semicolon at the end, TypeError b is not a function
  location = {foo:'bar'}

}
test();
```

as it gets translated to something like this

```
 a = b 
 (0,function(){var _hh$temp0={foo:'bar'};return __set$Loc(location,_hh$temp0)||(location=_hh$temp0);}.call(this))

```
which is interpreted like this in JavaScript, and fails 

```
 a = b(0,function(){var _hh$temp0={foo:'bar'};return __set$Loc(location,_hh$temp0)||(location=_hh$temp0);}.call(this))
```

 
Code without semicolons should run without problems as it is legal in JavaScript. You could say I could fix this issue with adding semicolons to the end of each line, or change the special variable named location. However that's not an option for us as this is not our code, part of a library we don't own.

As a solution maybe, line separation could  be added to the beginning of your transformation in case there is not any in original code?
 

### Steps to Reproduce:

1. Put the above test code snippet in a simple static page with this script inside, run locally
2. Create a standard testcafe test script (test.js) to call this page
3. Run your test in Chrome 
` testcafe chrome test.js`

4. See the error: ...
 ```
      JavaScript error details:
      TypeError: b is not a function
          at test (http://localhost:5000/index.js:6:7)
          at http://localhost:5000/index.js:14:1
 ```

### Your Environment details:

* node.js version:                    10.18.1
* browser name and version: TestCafe Chrome 85
* platform and version:          Windows 7
* testcafe version:                   1.8.8

Farfurix:
### Changes

1. The type "context" of a parent/current node is added. It affects the parentheses used in the comma operator.


Farfurix:
@testcafe-build-bot retest

src/processing/script/transform.ts
```diff
@@ -97,10 +98,10 @@ function addChangeForTransformedNode (state: State, changes: CodeChange[], repla
 
     if (state.newExpressionAncestor) {
         replaceNode(state.newExpressionAncestor, state.newExpressionAncestor, state.newExpressionAncestorParent!, state.newExpressionAncestorKey!);
-        changes.push(getChange(state.newExpressionAncestor));
+        changes.push(getChange(state.newExpressionAncestor, parentType));
```
LavrovArtem, line 101:
```diff
- changes.push(getChange(state.newExpressionAncestor, parentType));
+ changes.push(getChange(state.newExpressionAncestor, state.newExpressionAncestorParent.type));
```

LavrovArtem (changes requested):

Farfurix:
@testcafe-build-bot retest

LavrovArtem (approved):

Farfurix:
@testcafe-build-bot retest

Farfurix:
@testcafe-build-bot retest

PR #2421, opened 2020-08-28
Update body mutation tracking when adding the body to the document

Ogurecher:
## Purpose
We are overriding `document.getElementsByTagName()` and adding a `'hammerhead|shadow-ui|html-collection-wrapper'` property to the native function's results. It's implemented for optimization purposes - we store an element list returned by the first function call and update it only if the DOM tree updated. (the version of the `DOMMutationTracker` for this tag should be higher than the `'hammerhead|shadow-ui|html-collection-wrapper'` version)

This works for elements that trigger `element._onElementAdded` which calls `DOMMutationTracker.onElementChanged` that updates the version. However, the <body> element does not trigger it.

That means it's possible to call the `document.getElementsByTagName('body')` before the <body> element gets added to the page. This makes any subsequent `document.getElementsByTagName('body')` calls return the stored empty collection as the `DOMMutationTracker` does not get updated when the <body> element is added.

## Approach
I'm updating the tracked tag versions in `DOMMutationTracker` on the `DOMContentLoaded` event.

## References
Closes Devexpress/testcafe#5322

## Pre-Merge TODO

### DIFF
--- a/test/client/data/live-node-list/getBodyByTagName.html
+++ b/test/client/data/live-node-list/getBodyByTagName.html
@@ -14,13 +14,13 @@
             hammerhead.get('./utils/destination-location').forceLocation('http://localhost/sessionId/http://origin_iframe_host');
             hammerhead.start({ crossDomainProxyPort: 2000 });
 
-            window.top.bodyExistsBeforeAttachingBody = bodyExists();
+            window.bodyExistsBeforeAttachingBody = bodyExists();
         </script>
     </head>
     <body>
         <script>
             document.addEventListener('DOMContentLoaded', function () {
-                window.top.bodyExistsAfterAttachingBody = bodyExists();
+                window.bodyExistsAfterAttachingBody = bodyExists();
             });
         </script>
     </body>
--- a/test/client/data/live-node-list/getBodyByTagName.html
+++ b/test/client/data/live-node-list/getBodyByTagName.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <title>Title</title>
+        <script src="/hammerhead.js" class="script-hammerhead-shadow-ui"></script>
+        <script>
+            function bodyExists () {
+                return document.getElementsByTagName('body').length === 1;
+            }
+
+            const hammerhead = window['%hammerhead%'];
+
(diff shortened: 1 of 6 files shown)


Farfurix (approved):

test/client/data/live-node-list/getBodyByTagName.html
```diff
@@ -0,0 +1,27 @@
...
+            function bodyExists () {
+                return document.getElementsByTagName('body').length === 1;
+            }
+
+            const hammerhead = window['%hammerhead%'];
+
+            hammerhead.get('./utils/destination-location').forceLocation('http://localhost/sessionId/http://origin_iframe_host');
+            hammerhead.start({ crossDomainProxyPort: 2000 });
+
+            window.top.bodyExistsBeforeAttachingBody = bodyExists();
```
LavrovArtem, line 17: I think that it is unnecessary to save the test data in the top window. Just save it in the current iframe and use `iframe.contentWindow.<prop>` in your test.

test/client/fixtures/sandbox/node/html-collection-wrapper-test.js
```diff
@@ -443,5 +443,17 @@ module('getElementsByTagName', function () {
 
             DOMMutationTrackerProto.getVersion = storedGetVersion;
         });
+
+        test('getElementsByTagName(\'body\') updates correctly GH-5322', function () {
```
LavrovArtem, line 447: Use \` or `"` instead of `\'`.

test/client/fixtures/sandbox/node/html-collection-wrapper-test.js
```diff
@@ -443,5 +443,17 @@ module('getElementsByTagName', function () {
...
+            return createTestIframe({ src: getSameDomainPageUrl('../../../data/live-node-list/getBodyByTagName.html') })
+                .then(function () {
+                    const assertions = [];
+
+                    assertions.push([!window.top.bodyExistsBeforeAttachingBody, true, 'getElementsByTagName(\'body\') returned an empty collection before the body was created']);
```
LavrovArtem, line 452: Just use `strictEqual`.

LavrovArtem (changes requested):
Minor issues

LavrovArtem (approved):

AndreyBelym (approved):

PR #2203, opened 2019-12-27
Fix 'EventTarget.prototype.addEventListener is not overridden' (close #1902)

Farfurix:
Fix #1902 

### Changes
1. `EventTarget` native methods (non-IE11 case) were added to `nativeMethods.addEventListener`, `nativeMethods.removeEventListener` and `nativeMethods.dispatchEvent` methods.
2. `_createElementOverridedMethods` and `_createDocumentBodyOverridedMethods ` are combined to `createOverriddenMethods`.
3. `_getBodyEventListenerWrapper ` and `_getEventListenerWrapper` are combined to `_getEventListenerWrapper`.
4. The ["SVGElement.dispatchEvent should be overriden"](https://github.com/DevExpress/testcafe-hammerhead/pull/2203/files#diff-d08ffa5e7bead4978f8ff10e1f95c97c) test was fixed (SVG element creation).
5. Naming.
a) Methods: `...overrided...` -> `...overridden...`.
b) `eventListeningInfo` -> `eventCtx`.

### Native behavior checking (non IE11)

<details>
<summary>html:</summary>

```html
<html>
<head></head>
<body>
<script>
    var iframe = document.createElement('iframe');

    iframe.id = 'id1';

    iframe.onload = function () {
        console.log('> EventTarget.prototype.addEventListener = \'add\';');
        console.log('> EventTarget.prototype.removeEventListener = \'remove\';');
        console.log('> EventTarget.prototype.dispatchEvent = \'dispatch\';');
        EventTarget.prototype.addEventListener = 'add';
        EventTarget.prototype.removeEventListener = 'remove';
        EventTarget.prototype.dispatchEvent = 'dispatch';
        var xhr = new XMLHttpRequest();

        console.log('\n');

        console.log('window.addEventListener === \'add\':', window.addEventListener === 'add'); // "true"
        console.log('document.addEventListener === \'add\':', document.addEventListener === 'add'); // "true"
        console.log('document.body.addEventListener === \'add\':', document.body.addEventListener === 'add'); // "true"
        console.log('iframe.addEventListener === \'add\':', iframe.addEventListener === 'add'); // "true"
        console.log('\n');
        console.log('window.removeEventListener === \'remove\':', window.removeEventListener === 'remove'); // "true"
        console.log('document.removeEventListener === \'remove\':', document.removeEventListener === 'remove'); // "true"
        console.log('document.body.removeEventListener === \'remove\':', document.body.removeEventListener === 'remove'); // "true"
[...]

Issue #2423, opened 2020-08-31
window.location constructor properties shadowed incorrectly


adriancable:
Sample code:

```
console.log('-1->', window.location.constructor.name);
console.log('-2->', window.location.constructor.toString());
console.log('-3->', Function.prototype.toString.apply(window.location.constructor));
console.log('-4->', window.location instanceof Location);
```

When run in a real browser:

-1-> Location
-2-> function Location() { [native code] }
-3-> function Location() { [native code] }
-4-> true

When run under Testcafe:

-1-> g
-2-> function(e,t,r){var n=this ...
-3-> function(e,t,r){var n=this ...
-4-> false

Issue is that many websites (e.g. Google login) use a check like this as part of botguard and change their behaviour as a result.

Can testcafe-hammerhead be improved to "stealth" these changes?

Farfurix:
@adriancable 

> When run under Testcafe:
>
> -1-> g
> -2-> function(e,t,r){var n=this ...
> -3-> function(e,t,r){var n=this ...
> -4-> false

I checked several websites and got the correct result:
```
-1-> Location
-2-> function Location() { [native code] }
-3-> function Location() { [native code] }
-4-> true
```

Could you please share your project and your environment (see our [Bug report template](https://github.com/DevExpress/testcafe-hammerhead/issues/new?assignees=&labels=&template=bug_report.md&title=))? We will examine it and check for a suitable solution.

adriancable:
@Farfurix - to reproduce please do the following.

1. Clone new copy of testcafe-hammerhead from github, and run the playground:

git clone https://github.com/DevExpress/testcafe-hammerhead.git
cd testcafe-hammerhead
npm install
node node_modules/gulp/bin/gulp http-playground

2. Navigate to http://localhost:1400 in Chrome (I am on 85.0.4183.83) or Safari (13.1.2), and direct the playground to this URL:

https://sidewinder.starlinghome.io/beta_old/ifcheck.html

3. Look in the console.

Farfurix:
@adriancable

Thank you for the information. I reproduced the issue using your markup:
```html
<html>
<head></head>
<body></body>
<script>
  console.log('-1->', window.location.constructor.name);
  console.log('-2->', window.location.constructor.toString());
  console.log('-3->', Function.prototype.toString.apply(window.location.constructor));
  console.log('-4->', window.location instanceof Location);
</script>
</html>
```

Our team will examine it and check for a suitable solution.

[...]

PR #2433, opened 2020-09-07
Fix string representation of native func wrappers

### LINKED ISSUE #2394: Uncaught TypeError: Illegal invocation
The page is not loaded, there is the following error in DevTools console:
```
Uncaught TypeError: Illegal invocation
    at Window.addEventListener (hammerhead.js:7)
    at <anonymous>:2355:89162
    at Object.<anonymous> (<anonymous>:2355:89829)
    at Object.jQuery (<anonymous>:2355:89881)
    at n (<anonymous>:15:228)
    at Object.<anonymous> (<anonymous>:2355:87883)
    at Object.jQuery (<anonymous>:2355:88125)
    at n (<anonymous>:15:228)
    at Object.jQuery (<anonymous>:2355:87042)
    at n (<anonymous>:15:228)
```

Please see the [T915177](https://supportcenter.devexpress.com/internal/ticket/details/T915177) private ticket for details.

testcafe-hammerhead version: 17.1.11

wentwrong:
## Purpose 
        
Customer who submitted the [original](https://supportcenter.devexpress.com/internal/ticket/details/T915177/) ticket to our Support Center is using `Power BI` library, which causes an issue when the customer tries to run tests on their site. First, the library creates a fake copy of `window` object ([issue](https://github.com/microsoft/PowerBI-visuals/issues/165)). Then it detects native functions by running a regexp search with pattern `/\{\s*\[native code\]\s*\}/` against the result of `Function.prototype.toString()` call. When a native function is found, the library binds it to a newly created `window` instance. But in our case, Power BI library can't detect any native functions because we're wrapping them, so their string representations are just the code of these wrappers. To resolve this, we need to override string representations of our native functions' wrappers so that `Function.prototype.toString()` method returns the expected value.

## Approach

The approach we decided on is to save the original string representation of a native function right in its wrapper as an additional property. Then override `Function.prototype.toString()` method so it returns this data if the function passed to it has our property, otherwise dispatch to default native method call.

It should be noted that instead of saving the initial string representation as a property, we could save a reference to the original native function in its wrapper and dispatch to that when needed (which is the same way `zone.js` library does it).

## References
Closes #2394 
Fixed missed typo that was introduced in #2203: should be `EventTarget` instead of `EventLisener`

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/event/index.ts
+++ b/src/client/sandbox/event/index.ts
@@ -15,7 +15,7 @@ import UnloadSandbox from './unload';
 import MessageSandbox from './message';
 import ShadowUI from '../shadow-ui';
 import TimersSandbox from '../timers';
-import { overrideFunction, overrideStringRepresentation } from '../../utils/property-overriding';
+import { overrideFunction, overrideStringRepresentation } from '../../utils/overriding';
 
 export default class EventSandbox extends SandboxBase {
     EVENT_PREVENTED_EVENT: string = 'hammerhead|event|event-prevented';
--- a/src/client/sandbox/event/index.ts
+++ b/src/client/sandbox/event/index.ts
@@ -194,16 +194,14 @@ export default class EventSandbox extends SandboxBase {
         // @ts-ignore Window constructor has no the focus method
         window.Window.focus = this._overriddenMethods.focus;
 
-        // NOTE:
-        // we cannot use 'overrideFunction' here since the focus method may not exist
+        // NOTE: we cannot use 'overrideFunction' here since the focus method may not exist
         // @ts-ignore Window constructor has no the focus method
         overrideStringRepresentation(window.Window.focus, nativeMethods.focus);
 
         // @ts-ignore Window constructor has no the blur method
         window.Window.blur  = this._overriddenMethods.blur;
 
-        // NOTE:
-        // we cannot use 'overrideFunction' here since the blur method may not exist
+        // NOTE: we cannot use 'overrideFunction' here since the blur method may not exist
         // @ts-ignore Window constructor has no the blur method
         overrideStringRepresentation(window.Window.blur, nativeMethods.blur);
 
--- a/src/client/sandbox/event/index.ts
+++ b/src/client/sandbox/event/index.ts
@@ -15,6 +15,7 @@ import UnloadSandbox from './unload';
 import MessageSandbox from './message';
 import ShadowUI from '../shadow-ui';
 import TimersSandbox from '../timers';
+import { overrideFunction, overrideStringRepresentation } from '../../utils/property-overriding';
 
 export default class EventSandbox extends SandboxBase {
     EVENT_PREVENTED_EVENT: string = 'hammerhead|event|event-prevented';
@@ -171,33 +172,39 @@ export default class EventSandbox extends SandboxBase {
     attach (window: Window & typeof globalThis): void {
         super.attach(window);
 
-        window.HTMLInputElement.prototype.setSelectionRange    = this._overriddenMethods.setSelectionRange;
-        window.HTMLTextAreaElement.prototype.setSelectionRange = this._overriddenMethods.setSelectionRange;
+        overrideFunction(window.HTMLInputElement.prototype, 'setSelectionRange', this._overriddenMethods.setSelectionRange, );
+        overrideFunction(window.HTMLTextAreaElement.prototype, 'setSelectionRange', this._overriddenMethods.setSelectionRange);
 
         if (isIE11) {
-            window.Window.prototype.dispatchEvent      = this._overriddenMethods.dispatchEvent;
-            window.Document.prototype.dispatchEvent    = this._overriddenMethods.dispatchEvent;
-            window.HTMLElement.prototype.dispatchEvent = this._overriddenMethods.dispatchEvent;
-            window.SVGElement.prototype.dispatchEvent  = this._overriddenMethods.dispatchEvent;
+            overrideFunction(window.Window.prototype, 'dispatchEvent', this._overriddenMethods.dispatchEvent);
+            overrideFunction(window.Document.prototype, 'dispatchEvent', this._overriddenMethods.dispatchEvent);
+            overrideFunction(window.HTMLElement.prototype, 'dispatchEvent', this._overriddenMethods.dispatchEvent);
+            overrideFunction(window.SVGElement.prototype, 'dispatchEvent', this._overriddenMethods.dispatchEvent);
         }
-        else
-            window.EventTarget.prototype.dispatchEvent = this._overriddenMethods.dispatchEvent;
+        else {
+            overrideFunction(window.EventTarget.prototype, 'dispatchEvent', this._overriddenMethods.dispatchEvent);
+        }
 
-        window.HTMLElement.prototype.focus    = this._overriddenMethods.focus;
-        window.HTMLElement.prototype.blur     = this._overriddenMethods.blur;
-        window.HTMLElement.prototype.click    = this._overriddenMethods.click;
-        window.Event.prototype.preventDefault = this._overriddenMethods.preventDefault;
+        overrideFunction(window.HTMLElement.prototype, 'focus', this._overriddenMethods.focus);
+        overrideFunction(window.HTMLElement.prototype, 'blur', this._overriddenMethods.blur);
+        overrideFunction(window.HTMLElement.prototype, 'click', this._overriddenMethods.click);
+
+        overrideFunction(window.Event.prototype, 'preventDefault', this._overriddenMethods.preventDefault);
 
         // @ts-ignore Window constructor has no the focus method
         window.Window.focus = this._overriddenMethods.focus;
+        // @ts-ignore Window constructor has no the focus method
+        overrideStringRepresentation(window.Window.focus, nativeMethods.focus);
+
         // @ts-ignore Window constructor has no the blur method
         window.Window.blur  = this._overriddenMethods.blur;
+        // @ts-ignore Window constructor has no the blur method
+        overrideStringRepresentation(window.Window.blur, nativeMethods.blur);
 
         // @ts-ignore TextRange exists only in IE
-        if (window.TextRange && window.TextRange.prototype.select) {
+        if (window.TextRange && window.TextRange.prototype.select)
             // @ts-ignore TextRange exists only in IE
-            window.TextRange.prototype.select = this._overriddenMethods.select;
-        }
+            overrideFunction(window.TextRange.prototype, 'select', this._overriddenMethods.select);
 
         this.listeners.initElementListening(document, DOM_EVENTS);
         this.listeners.initElementListening(window, DOM_EVENTS.concat(['load', 'beforeunload', 'pagehide', 'unload', 'message']));
--- a/src/client/sandbox/event/index.ts
+++ b/src/client/sandbox/event/index.ts
@@ -172,7 +172,7 @@ export default class EventSandbox extends SandboxBase {
     attach (window: Window & typeof globalThis): void {
         super.attach(window);
 
-        overrideFunction(window.HTMLInputElement.prototype, 'setSelectionRange', this._overriddenMethods.setSelectionRange, );
+        overrideFunction(window.HTMLInputElement.prototype, 'setSelectionRange', this._overriddenMethods.setSelectionRange);
         overrideFunction(window.HTMLTextAreaElement.prototype, 'setSelectionRange', this._overriddenMethods.setSelectionRange);
 
         if (isIE11) {
@@ -193,11 +193,17 @@ export default class EventSandbox extends SandboxBase {
 
         // @ts-ignore Window constructor has no the focus method
         window.Window.focus = this._overriddenMethods.focus;
+
+        // NOTE:
+        // we cannot use 'overrideFunction' here since the focus method may not exist
         // @ts-ignore Window constructor has no the focus method
         overrideStringRepresentation(window.Window.focus, nativeMethods.focus);
 
         // @ts-ignore Window constructor has no the blur method
         window.Window.blur  = this._overriddenMethods.blur;
+
+        // NOTE:
+        // we cannot use 'overrideFunction' here since the blur method may not exist
         // @ts-ignore Window constructor has no the blur method
         overrideStringRepresentation(window.Window.blur, nativeMethods.blur);
 
--- a/src/client/sandbox/node/document/index.ts
+++ b/src/client/sandbox/node/document/index.ts
@@ -10,7 +10,7 @@ import DocumentWriter from './writer';
 import ShadowUI from './../../shadow-ui';
 import INTERNAL_PROPS from '../../../../processing/dom/internal-properties';
 import LocationAccessorsInstrumentation from '../../code-instrumentation/location';
-import { overrideDescriptor, createOverriddenDescriptor, overrideFunction } from '../../../utils/property-overriding';
+import { overrideDescriptor, createOverriddenDescriptor, overrideFunction } from '../../../utils/overriding';
 import NodeSandbox from '../index';
 import { getDestinationUrl } from '../../../utils/url';
 import DocumentTitleStorageInitializer from './title-storage-initializer';
--- a/src/client/sandbox/node/document/index.ts
+++ b/src/client/sandbox/node/document/index.ts
@@ -115,11 +115,6 @@ export default class DocumentSandbox extends SandboxBase {
         return result;
     }
 
-    private static _ensureDocumentMethodOverride (document, overridenMethods, methodName) {
-        if (document[methodName] !== overridenMethods[methodName])
-            document[methodName] = overridenMethods[methodName];
-    }
-
     attach (window, document) {
         if (this._needToUpdateDocumentWriter(window, document)) {
             this.documentWriter = new DocumentWriter(window, document);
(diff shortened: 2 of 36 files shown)


wentwrong:
Also, I think that the process of creating new wrappers for native functions should be streamlined somehow. Because otherwise everytime someone adds a new wrapper they'll have to manually override its string representation by calling
```js
overrideStringRepresentation(ourWrapper, nativeFunction);
```
Which they can easily forget to do.

As a suggestion, we can refactor our wrappers to make them child classes inherited from the same base class, which will contain the code responsible for correcting string representations. That would also let us reduce the size of a gigantic `attach` method of the `WindowSandbox` class.

wentwrong:
@testcafe-build-bot retest

adriancable:
This looks great - does it also address #2423 ?

AndreyBelym:
This PR fixes only a part of #2423. We will need a lot more work to fix all the cases described there.

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,56 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+        });
+    }
+}
+
+export function overrideConstructor<O extends object, K extends keyof O> (obj: O, fnName: K, wrapper: Function, overrideProtoConstructor: boolean = false): void {
+    const nativePrototype = obj[fnName]['prototype'];
+    
+    overrideFunction(obj, fnName, wrapper);
+
+    if (nativePrototype) {
```
LavrovArtem, line 96: I don't understand why we need this condition. Each function has the `prototype` property and that `prototype` has `constructor`.

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,56 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+        // NOTE: restore native prototype (to make `instanceof` work as expected)
+        nativeMethods.objectDefineProperty(obj[fnName], 'prototype', {
```
LavrovArtem, line 98: We don't need to use `objectDefineProperty` there.

src/processing/dom/internal-properties.ts
```diff
@@ -15,5 +15,7 @@ export default {
     forceProxySrcForImage:     'hammerhead|image|force-proxy-src-flag',
     skipNextLoadEventForImage: 'hammerhead|image|skip-next-load-event-flag',
     cachedImage:               'hammerhead|image|cached-image',
-    sandboxIsReattached:       'hammerhead|sandbox-is-reattached'
+    sandboxIsReattached:       'hammerhead|sandbox-is-reattached',
+    
+    nativeStringRepresentation: 'hammerhead|nativeStringRepresentation'
```
LavrovArtem, line 20: `nativeStringRepresentation: 'hammerhead|native-string-representation'`

test/client/before-test.js
```diff
@@ -199,6 +199,24 @@
...
     };
 
+    window.checkStringRepresentation = function (wrappedFn, originalFn) {
+        strictEqual(
+            wrappedFn.toString(),
+            originalFn.toString(),
+            'the outputs of the "toString()" method should be the same'
+        );
+        strictEqual(
+            Function.prototype.toString.call(wrappedFn),
```
LavrovArtem, line 209: `Function.prototype.toString` and `nativeMethods.Function.prototype.toString` are the same functions

LavrovArtem (changes requested):

test/client/before-test.js
wentwrong: Good catch! I found the same problem with `Worker.prototype`, `XMLHttpRequest.prototype` and `Function.prototype` ctor's.

wentwrong:
@testcafe-build-bot retest

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,48 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+    if (wrapperNameDescriptor) {
+        // eslint-disable-next-line no-restricted-properties
+        wrapperNameDescriptor.value = name;
+
+        nativeMethods.objectDefineProperty(fn, 'name', wrapperNameDescriptor);
+    }
+}
+
+export function overrideStringRepresentation (nativeFnWrapper: Function, nativeFn: Function): void {
+    overrideFunctionName(nativeFnWrapper, nativeFn['name']);
```
LavrovArtem, line 71:
```diff
-nativeFn['name']
+nativeFn.name
```

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,48 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+
+    nativeFnWrapper[INTERNAL_PROPS.nativeStrRepresentation] = nativeMethods.Function.prototype.toString.call(nativeFn);
```
LavrovArtem, line 73:
```js
nativeMethods.objectDefineProperty(nativeFnWrapper, INTERNAL_PROPS.nativeStrRepresentation, { value: nativeMethods.Function.prototype.toString.call(nativeFn) });
```

LavrovArtem (changes requested):
minor issues

wentwrong:
@testcafe-build-bot retest

wentwrong:
@testcafe-build-bot retest

src/client/utils/property-overriding.ts
```diff
@@ -1,4 +1,5 @@
 import nativeMethods from '../sandbox/native-methods';
+import INTERNAL_PROPS from '../../processing/dom/internal-properties';
```
miherlosev, line 2: Rename the
`src/client/utils/property-overriding.ts` file to
to `src/client/utils/overriding.ts` since now this file contains functions for override not only properties.

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,53 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
 
     nativeMethods.objectDefineProperty(obj, prop, descriptor);
 }
+
+function overrideFunctionName (fn: Function, name: string) {
+    const wrapperNameDescriptor = nativeMethods.objectGetOwnPropertyDescriptor(fn, 'name');
```
miherlosev, line 60: `const nameDescriptor =...`

miherlosev: Use `//eslint-disable-line` because it makes the code more readable.
Use `fast forward` strategy here to avoid additional brackets:
```js
 if (!nameDescriptor)
    return;

nameDescriptor.value = name; // eslint-disable-line no-restricted-properties
```

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,53 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+    }
+}
+
+export function overrideStringRepresentation (nativeFnWrapper: Function, nativeFn: Function): void {
+    overrideFunctionName(nativeFnWrapper, nativeFn.name);
+
+    // NOTE:
+    // `configurable`, `enumerable` and `writable` optional keys of the descriptor are set implicitly to false
+    // because we want to hide our internal property from the user code
+    nativeMethods.objectDefineProperty(nativeFnWrapper, INTERNAL_PROPS.nativeStrRepresentation, {
```
miherlosev, line 76: Move this instruction to the `overrideToString` function.

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,53 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+        value: nativeMethods.Function.prototype.toString.call(nativeFn)
+    });
+}
+
+export function isNativeFunction (fn: Function): boolean {
+    return !nativeMethods.objectHasOwnProperty.call(fn, INTERNAL_PROPS.nativeStrRepresentation);
+}
+
+export function overrideFunction<O extends object, K extends keyof O> (obj: O, fnName: K, wrapper: Function): void {
```
miherlosev, line 85: I propose to change the function signature to 
```js
overrideFunction (target: Function, wrapper: Function);
```

src/client/utils/property-overriding.ts
```diff
@@ -54,3 +55,53 @@ export function overrideDescriptor<O extends object, K extends keyof O> (obj: O,
...
+
+    if (wrapperNameDescriptor) {
+        // eslint-disable-next-line no-restricted-properties
+        wrapperNameDescriptor.value = name;
+
+        nativeMethods.objectDefineProperty(fn, 'name', wrapperNameDescriptor);
+    }
+}
+
+export function overrideStringRepresentation (nativeFnWrapper: Function, nativeFn: Function): void {
```
miherlosev, line 70: The name of this function is unclear.
Instead of it use combination of  the `overrideFunctionName` and `overrideToString` functions.

test/client/data/node-sandbox/iframe-override-elems-after-write.html
```diff
@@ -20,6 +20,7 @@
 <script>
     var iframeSandbox = window['%hammerhead%'].sandbox.iframe;
     var INTERNAL_PROPS = hammerhead.get('../processing/dom/internal-properties');
+    var propertyOverridingUtils = hammerhead.get('../client/utils/property-overriding');
```
miherlosev, line 23: Align assignments:
```js
var iframeSandbox     = ...
var INTERNAL_PROPS =...
...
```

miherlosev (changes requested):

src/client/sandbox/event/index.ts
```diff
@@ -171,33 +172,45 @@ export default class EventSandbox extends SandboxBase {
...
-        window.Event.prototype.preventDefault = this._overriddenMethods.preventDefault;
+        overrideFunction(window.Event.prototype, 'preventDefault', this._overriddenMethods.preventDefault);
 
         // @ts-ignore Window constructor has no the focus method
         window.Window.focus = this._overriddenMethods.focus;
+
+        // NOTE:
+        // we cannot use 'overrideFunction' here since the focus method may not exist
+        // @ts-ignore Window constructor has no the focus method
+        overrideStringRepresentation(window.Window.focus, nativeMethods.focus);
```
miherlosev, line 200: Add the necessary checks into the `overrideFunction` and use it here.

Remove the `overrideStringRepresentation` function and use the combination of `overrideFunctionName` and `overrdieToString` instead.

miherlosev (changes requested):

test/client/fixtures/sandbox/node/document-test.js
```diff
@@ -1,9 +1,10 @@
...
-var INTERNAL_PROPS      = hammerhead.get('../processing/dom/internal-properties');
-var urlUtils            = hammerhead.get('./utils/url');
-var destLocation        = hammerhead.get('./utils/destination-location');
-var settings            = hammerhead.get('./settings');
+var processScript           = hammerhead.get('../processing/script').processScript;
+var SHADOW_UI_CLASSNAME     = hammerhead.get('../shadow-ui/class-name');
+var INTERNAL_PROPS          = hammerhead.get('../processing/dom/internal-properties');
+var urlUtils                = hammerhead.get('./utils/url');
+var destLocation            = hammerhead.get('./utils/destination-location');
+var propertyOverridingUtils = hammerhead.get('./utils/property-overriding');
```
miherlosev, line 6: `var overriding = ...` here and above

test/client/fixtures/sandbox/iframe-test.js
```diff
@@ -143,13 +144,11 @@ test('native methods are properly initialized in an iframe without src (GH-279)'
             var iframeDocument         = iframe.contentDocument;
             var iframeWindow           = iframe.contentWindow;
             var iframeHammerhead       = iframeWindow['%hammerhead%'];
-            var nativeCreateElement    = iframeHammerhead.sandbox.nativeMethods.createElement.toString();
-            var nativeAppendChild      = iframeHammerhead.sandbox.nativeMethods.appendChild.toString();
-            var overridedCreateElement = iframeDocument.createElement.toString();
-            var overridedAppendChild   = iframeDocument.createElement('div').appendChild.toString();
+            var overridedCreateElement = iframeDocument.createElement;
```
miherlosev, line 147: Not `overrided`. Use the `overridden` word here and elsewhere.

miherlosev (changes requested):

src/client/sandbox/event/index.ts
wentwrong: I assume that we shouldn't assign our wrapper to a native function when it is missing (e.g. `window.Window.focus`, which is undefined in IE11). It would be much better to flag these cases somehow (which I did by combining plain assignment and using `overrideStringRepresentation` function). We need to ensure that we override only those native functions that exist.

miherlosev: In this case we can omit any flags because it clear behavior - override only existing functions.
Also, in general case the flag is the worst way to pass parameters.
See https://github.com/ryanmcdermott/clean-code-javascript#dont-use-flags-as-function-parameters

wentwrong: After a more in-depth discussion with @miherlosev and @LavrovArtem, we decided to use `overrideStringRepresentation` in cases like this. It's temporarily and later it'll be implemented properly through `overrideFunction`.

wentwrong:
@testcafe-build-bot retest

src/client/sandbox/node/document/index.ts
```diff
@@ -198,46 +193,46 @@ export default class DocumentSandbox extends SandboxBase {
             }
         };
 
-        window[nativeMethods.documentOpenPropOwnerName].prototype.open       = overridenMethods.open;
-        window[nativeMethods.documentClosePropOwnerName].prototype.close     = overridenMethods.close;
-        window[nativeMethods.documentWritePropOwnerName].prototype.write     = overridenMethods.write;
-        window[nativeMethods.documentWriteLnPropOwnerName].prototype.writeln = overridenMethods.writeln;
+        overrideFunction(window[nativeMethods.documentOpenPropOwnerName].prototype, 'open', overridenMethods.open);
```
miherlosev, line 196: `overriden` -> `overridden` here and everywhere.

`overriddenMethods.open`

src/client/sandbox/style.ts
```diff
@@ -168,19 +168,21 @@ export default class StyleSandbox extends SandboxBase {
...
             const nativeFn = this.nativeMethods.objectGetOwnPropertyDescriptor.call(window.Object, styleDeclarationProto, prop).value;// eslint-disable-line no-restricted-properties
 
             if (this.nativeMethods.objectHasOwnProperty.call(styleDeclarationProto, prop) &&
                 typeof nativeFn === 'function') {
-                styleDeclarationProto[prop] = function () {
+                (styleDeclarationProto[prop] as unknown as Function) = function () {
                     return nativeFn.apply(this[CSS_STYLE_PROXY_TARGET] || this, arguments);
                 };
+
+                // NOTE:
```
miherlosev, line 183: Use the follwong comment style:

```js
// NOTE: we cannot use 'overrideFunction' here since the function may not exist

// TODO: this function should not be used outside this file
```

miherlosev (changes requested):

src/client/utils/overriding.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!nameDescriptor)
+        return;
+
+    nameDescriptor.value = name; // eslint-disable-line no-restricted-properties
+
+    nativeMethods.objectDefineProperty(fn, 'name', nameDescriptor);
+}
+
+function overrideToString (nativeFnWrapper: Function, nativeFn: Function): void {
+    // NOTE:
```
miherlosev, line 71: I guess we can remove this comment because the all INTERNAL_PROPS added in the same way.

Also, hiding the proxy from the web site client code is a main concept.

src/client/sandbox/xhr.ts
```diff
@@ -85,35 +84,33 @@ export default class XhrSandbox extends SandboxBase {
             });
         }
 
-        window.XMLHttpRequest           = xmlHttpRequestWrapper;
-        xmlHttpRequestWrapper.prototype = xmlHttpRequestProto;
-        xmlHttpRequestWrapper.toString  = () => xmlHttpRequestToString;
-
         // NOTE: We cannot just assign constructor property in OS X 10.11 safari 9.0
+        overrideConstructor(window, 'XMLHttpRequest', xmlHttpRequestWrapper);
```
miherlosev, line 88: `// NOTE: We cannot just assign constructor property in OS X 10.11 safari 9.0`

Is this comment still actual?

miherlosev (changes requested):

src/client/sandbox/xhr.ts
wentwrong: Yes, we still cannot just assign constructor property (as `overrideConstructor` does when the fourth parameter `overrideProtoConstructor` is set to true). In the code below this line, we are using 'objectDefineProperty' in order to override constructor property.

wentwrong: I think a better way to phrase it to make the comment clearer would be:
```
// NOTE: We cannot just assign constructor property of the prototype of XMLHttpRequest in OS X 10.11 safari 9.0
```

miherlosev: At present, we don't use the Safari 9.0 (Safari 11.3 and 12.0 versions are used).
If this code is necessary, then update comment to 
`// NOTE: We cannot just assign constructor property of the prototype of XMLHttpRequest starts from safari 9.0`

miherlosev (approved):

LavrovArtem (approved):

PR #2460, opened 2020-10-19
Fix the issues that appeared after merging #2448

wentwrong:
## Purpose
Fix the issues introduced in #2448 and #2433 

## Approach
* Revert changes in the `destination-location.ts` file (because functions from this file are used in many places in the code). 
* Separate handling of the location object properties of the "about:blank" document in `LocationWrapper`
* <s>Instead of calling the location "origin" getter, call the active document's origin when processing "ancestorOrigins". According to the [specification](https://www.w3.org/TR/2017/WD-html52-20170228/browsers.html#ancestor-origins-array):
> 5. Append the Unicode serialization of current’s active document’s origin to output as a new value.</s>
* remove handling of the `origin` property because it's causes different problems
* set `configurable: true` to the native str representation property in order to avoid "TypeError: Attempting to change value of a readonly property" in iOS (discovered in testcafe client mobile tests)
## References
#2448 
#2433

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/code-instrumentation/location/wrapper.ts
+++ b/src/client/sandbox/code-instrumentation/location/wrapper.ts
@@ -185,12 +185,11 @@ export default class LocationWrapper {
         const createLocationPropertyDesc = (property, nativePropSetter) => {
             locationProps[property] = createOverriddenDescriptor(locationPropsOwner, property, {
                 getter: () => {
-                    const frameElement = domUtils.getFrameElement(window);
+                    const frameElement       = domUtils.getFrameElement(window);
+                    const inIframeWithoutSrc = frameElement && domUtils.isIframeWithoutSrc(frameElement);
+                    const parsedDestLocation = inIframeWithoutSrc ? window.location : getParsedDestLocation();
 
-                    if (frameElement && domUtils.isIframeWithoutSrc(frameElement))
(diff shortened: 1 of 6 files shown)


wentwrong:
@testcafe-build-bot retest

wentwrong:
@testcafe-build-bot retest

src/client/sandbox/code-instrumentation/location/wrapper.ts
```diff
@@ -196,7 +184,14 @@ export default class LocationWrapper {
 
         const createLocationPropertyDesc = (property, nativePropSetter) => {
             locationProps[property] = createOverriddenDescriptor(locationPropsOwner, property, {
-                getter: () => getParsedDestLocation()[property],
+                getter: () => {
+                    const frameElement = domUtils.getFrameElement(window);
+
```
LavrovArtem, line 189:
```js
const frameElement       = domUtils.getFrameElement(window);
const inIframeWithoutSrc = frameElement && domUtils.isIframeWithoutSrc(frameElement);
const parsedDestLocation = inIframeWithoutSrc ? window.location : getParsedDestLocation();

return parsedDestLocation[property];
```

LavrovArtem (changes requested):

LavrovArtem (approved):

Issue #2480, opened 2020-11-13
invalid property lookup, invalid error instance assumption


cdaringe:
### What is your Scenario?

- running tests in CI
 
### What is the Current behavior?

On test, we _periodically receive_:

```
13:06:13 TypeError: Cannot read property 'toString' of undefined
13:06:13     at Proxy._onServiceMessage (/opt/app/node_modules/testcafe-hammerhead/lib/proxy/index.js:149:41)
13:06:13     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

The below line assumes that the thrown entity is an `Error` instance, but as demonstrated above, clearly is not 100% of the time guaranteed to be the case:
https://github.com/DevExpress/testcafe-hammerhead/blob/352d2a71a672266f63fa5348fef097ce2db295c9/src/proxy/index.ts#L156
 
### What is the Expected behavior?

Either `toString` exists 100% of the time, or defensive programming practiced in exception handlers

 
### What is your public web site URL?

- internal. bummer. i would love to share it.

### Steps to Reproduce:

TBD. We have a job that runs our testcafe tests continuously to uncover flakiness. This error creeps up somewhat randomly in otherwise healthy tests. non-determinisitic.
 
### Your Environment details:

* node.js version:     12.17.0
* browser name and version: chome @ latest
* platform and version:  `testcafe/testcafe:1.8.6`, so alpine of sorts :)

benmonro:
my team has run into this issue as well.

miherlosev:
Hi @cdaringe

The `testcafe/testcafe:1.8.6` version is outdated. Try to run you tests with the latest testcafe version - `1.9.4`.
If the problem persists, try to find the type of the raised error. To do that, set the `DEBUG` environment variable and collect the request information into a file as described in this example:

```
set DEBUG=hammerhead:*
testcafe chrome index.js 2> hammerhead.log
```

Please share the file with the logged information.

PR #2481, opened 2020-11-13
fix: remove error instance assumption on property access

cdaringe:
## Purpose

Correct invalid assumption that `err` is guaranteed to be an `Error` instance.

## Approach

- only call toString if it exists

## References

https://github.com/DevExpress/testcafe-hammerhead/issues/2480

## Pre-Merge TODO

### DIFF
--- a/src/proxy/index.ts
+++ b/src/proxy/index.ts
@@ -153,7 +153,7 @@ export default class Proxy extends Router {
             catch (err) {
                 logger.serviceMsg('Service message %j, error %o', msg, err);
 
-                respond500(res, err.toString());
+                respond500(res, (err && err.toString) ? err.toString() : "");
             }
         }
         else


miherlosev:
Thank you for this PR. However, we cannot add code to the repository without a clear understanding of what it is.
So, let's wait for investigation results (see details in [https://github.com/DevExpress/testcafe-hammerhead/issues/2480#issuecomment-727802621](https://github.com/DevExpress/testcafe-hammerhead/issues/2480#issuecomment-727802621)).

cdaringe:
@miherlosev, we turned on DEBUG for a week or so and upgraded to TestCafe/hammerhead@latest. Given the noise it induced in our users logs, we had to turn off DEBUG at some point so users could self service in CI. after a couple weeks of running (probably 5k jobs?), I only got _one_ bug report from a peer reporting the problem. We cannot reliably reproduce this bug, likely due to races within hammerhead tooling. This PR, however, will help us identify the issue in the very rare cases where the issue occurs.

cdaringe:
...or could, perhaps if we serialize whatever actually gets thrown :)

swashcap:
I _think_ I figured out why `error` is undefined.

1. `Proxy#_onServiceMessage` is wired up to handle `POST /messaging` requests
    1. It calls `session.handleServiceMessage`: https://github.com/DevExpress/testcafe-hammerhead/blob/70f14d48f1173c58dbfe97ada2b42b014abef83c/src/proxy/index.ts#L147
    2. This calls `Session#handleServiceMessage `: https://github.com/DevExpress/testcafe-hammerhead/blob/70f14d48f1173c58dbfe97ada2b42b014abef83c/src/session/index.ts#L111-L116
    3. This calls a method on the `Session` depending on the contents of its first parameter, `msg`
2. The instance of `Service` in a test run is _probably_ actually an instance of `SessionController` in DevExpress/testcafe: https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/session-controller.js#L10
    1. This overrides `Session#handleServiceMessage`: https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/session-controller.js#L27-L33
    2. `Session#handleServiceMessage` is `call`-ed with a different `this` in cases where the `SessionController` instance has a `currentTestRun` property that matches this condition: https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/session-controller.js#L29-L30
        * The `currentTestRun` is instantiated here: https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/runner/test-run-controller.ts#L264. It is either an instance of [`TestRun`](https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js) or testcafe-legacy-api.
        * I assumed we weren't using the legacy runner because our project began relatively recently
    3. `TestRun` has these additional commands: https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js#L946-L1001
3. `TestRun#ready` is here:
https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js#L947-L972
    1. It calls `TestRun#_handleDriverRequest` and assigns its response to `this.lastDriverStatusResponse`:
    https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js#L960
        * `TestRun#_handleDriverRequest` returns a `Promise` that is immediately rejected if `this.disconnected` is true: https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js#L552-L553 with no arguments (rejected with `undefined`)
    2. `this.lastDriverStatusResponse` is truthy and returned https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js#L962-L963 (the method returns a `Promise` with no value).

I don't have shell access to our CI servers to be able to debug test runs and confirm this is the async call stack. However, it seems possible that there's a race condition where `disconnect` is set to `true` prior to a client issuing a ready message.

Seems like this should at least reject with an error?

https://github.com/DevExpress/testcafe/blob/ca317808e2387def833cb1c52b0729d8cd93d07b/src/test-run/index.js#L553

miherlosev:
Hi @swashcap

Thank you for the detailed investigation. We will check your hypothesis and review this PR.

AlexKamaev:
merged in https://github.com/DevExpress/testcafe-hammerhead/pull/2536

PR #2483, opened 2020-11-17
fix `Network requests aren't recieved by server after clicking on a link` (close #2467)

### LINKED ISSUE #2467: Network requests aren't recieved by server after clicking on a link
### What is your Scenario?
I'm testing a custom analytics service we have written that captures a number of events including on click events that are fired when a user clicks on a link to an external site.
 
We have found inside the hammerhead reverse proxy that any network call fired on or after clicking on an `a href` link are never sent by the browser and received by the server. It appears the browser cancels all in flight network requests or fails to fire any new network requests and all unload events as well.

I've tested this back to back with multiple browsers and the events work normally as expected when interacting with the browser but they fail inside of testcafe tests and when manually using the site inside the hammerhead reverse proxy.
 
Any network event triggered as part of a link click and all unload behaviour is finalised when navigating to a new page.
 
### What is your public web site URL?
https://linktr.ee/dalebaldwin

### Steps to Reproduce:
Given a user is at https://linktr.ee/dalebaldwin
When a user clicks on the `Github` link
Then all network events are fired and completed during transition to github.com
 
### Your Environment details:

node.js
Chrome, Chromium, Firefox, Safari, UI Kit webview, IE, Edge

LavrovArtem:


miherlosev (approved):

shyedhu:
Hi @miherlosev and team , Thank you for fixing the issue https://github.com/DevExpress/testcafe-hammerhead/issues/2467. I appreciate on your effort ! Seems above the PR has been merged .. However i would like to validate on that fix against my large application. Can you please let me know when and which testcafe version that i can use on above the key fix ! Thank you !

miherlosev:
Hi @shyedhu
At present, we cannot say when and with which TestCafe version the fix will be released.
You can check it in the following way:

* track a new `testcafe-hammerhead` module version with the fix.
* track a new `testcafe` version with the updated `testcafe-hammerhead` module.

PR #2485, opened 2020-11-20
The "origin" header should replicate the case of the "referer" header + server tests refactoring

LavrovArtem:


src/request-pipeline/header-transforms/index.ts
```diff
@@ -54,6 +54,13 @@ export function transformHeadersCaseToRaw (headers: OutgoingHttpHeaders, rawHead
         }
     }
 
+    // NOTE: We doesn't send a cross-domain client request as cross-domain.
```
AlexKamaev, line 57: Can we have a similar situation with other headers?

src/request-pipeline/header-transforms/index.ts
```diff
@@ -54,6 +54,13 @@ export function transformHeadersCaseToRaw (headers: OutgoingHttpHeaders, rawHead
...
+    // Therefore, the origin header can be absent and we cannot decide its case. GH-2382
+    if (headers.hasOwnProperty('origin') && processedHeaders.hasOwnProperty('Referer')) {
+        processedHeaders['Origin']                   = headers['origin'];
+        headersNames[headersNames.indexOf('origin')] = void 0;
```
AlexKamaev, line 61: It seems to me that removing the element from the array is more clear.

AlexKamaev (approved):

src/request-pipeline/header-transforms/index.ts
AlexKamaev: imho, not required

src/request-pipeline/header-transforms/index.ts
```diff
@@ -54,6 +54,13 @@ export function transformHeadersCaseToRaw (headers: OutgoingHttpHeaders, rawHead
         }
     }
 
+    // NOTE: We doesn't send a cross-domain client request as cross-domain.
+    // Therefore, the origin header can be absent and we cannot decide its case. GH-2382
+    if (headers.hasOwnProperty('origin') && processedHeaders.hasOwnProperty('Referer')) {
```
miherlosev, line 59: Since it's an edge case, let's move it to the separate function and put the comment into it.
```js
...
calculateOriginHeaderCase (headers, processedHeaders, headersNames);
...
```

miherlosev (changes requested):

miherlosev (approved):

LavrovArtem:
@testcafe-build-bot retest

PR #2493, opened 2020-11-30
Added the strictBindCallApply rule to tsconfig

Ogurecher:
## Purpose
Synchronize typescript options with testcafe for monorepo migration.

## Approach
Added the strictBindCallApply rule to tsconfig, fixed errors (mostly type casting). Also added alwaysStrict (no errors occured).

## References
Part of https://github.com/DevExpress/testcafe/issues/5666

## Pre-Merge TODO

### DIFF
--- a/src/client/dom-processor/client-dom-adapter.ts
+++ b/src/client/dom-processor/client-dom-adapter.ts
@@ -89,8 +89,8 @@ export default class ClientDomAdapter extends BaseDomAdapter {
         }
     }
 
-    getProxyUrl (): string {
-        return getProxyUrl.apply(null, arguments);
+    getProxyUrl (...args): string {
+        return getProxyUrl.apply(null, args as [string, any?]);
     }
 
     isTopParentIframe (el: HTMLElement): boolean {
--- a/src/client/dom-processor/client-dom-adapter.ts
+++ b/src/client/dom-processor/client-dom-adapter.ts
@@ -89,8 +89,8 @@ export default class ClientDomAdapter extends BaseDomAdapter {
         }
     }
 
-    getProxyUrl (...args): string {
-        return getProxyUrl.apply(null, args as [string, any?]);
+    getProxyUrl (...args: [string, any?]): string {
+        return getProxyUrl.apply(null, args);
     }
 
     isTopParentIframe (el: HTMLElement): boolean {
--- a/src/client/dom-processor/client-dom-adapter.ts
+++ b/src/client/dom-processor/client-dom-adapter.ts
@@ -90,7 +90,7 @@ export default class ClientDomAdapter extends BaseDomAdapter {
     }
 
     getProxyUrl (...args: [string, any?]): string {
-        return getProxyUrl.apply(null, args);
+        return getProxyUrl(...args);
     }
 
     isTopParentIframe (el: HTMLElement): boolean {
--- a/src/client/sandbox/fetch.ts
+++ b/src/client/sandbox/fetch.ts
@@ -166,9 +166,8 @@ export default class FetchSandbox extends SandboxBaseWithDelayedSettings {
                 return sandbox.delayUntilGetSettings(() => this.fetch.apply(this, args));
 
             // NOTE: Safari processed the empty `fetch()` request without `Promise` rejection (GH-1613)
-            if (!args.length && !browserUtils.isSafari) {
+            if (!args.length && !browserUtils.isSafari)
                 return nativeMethods.fetch.apply(this, [] as unknown as [RequestInfo, RequestInit?]);
-            }
 
             try {
                 FetchSandbox._processArguments(args);
--- a/src/client/sandbox/fetch.ts
+++ b/src/client/sandbox/fetch.ts
@@ -102,8 +102,8 @@ export default class FetchSandbox extends SandboxBaseWithDelayedSettings {
         return entry;
     }
 
-    static _entriesWrapper (...args) {
-        const iterator   = nativeMethods.headersEntries.apply(this, args as []);
+    static _entriesWrapper (...args: []) {
+        const iterator   = nativeMethods.headersEntries.apply(this, args);
         const nativeNext = iterator.next;
 
         iterator.next = () => FetchSandbox._entriesFilteredNext(iterator, nativeNext);
@@ -111,8 +111,8 @@ export default class FetchSandbox extends SandboxBaseWithDelayedSettings {
         return iterator;
     }
 
-    static _valuesWrapper (...args) {
-        const iterator   = nativeMethods.headersEntries.apply(this, args as []);
+    static _valuesWrapper (...args: []) {
+        const iterator   = nativeMethods.headersEntries.apply(this, args);
         const nativeNext = iterator.next;
 
         iterator.next = () => {
@@ -167,7 +167,7 @@ export default class FetchSandbox extends SandboxBaseWithDelayedSettings {
 
             // NOTE: Safari processed the empty `fetch()` request without `Promise` rejection (GH-1613)
             if (!args.length && !browserUtils.isSafari)
-                return nativeMethods.fetch.apply(this, [] as unknown as [RequestInfo, RequestInit?]);
+                return nativeMethods.fetch.apply(this, [] as unknown as [RequestInfo]);
 
             try {
                 FetchSandbox._processArguments(args);
@@ -220,7 +220,7 @@ export default class FetchSandbox extends SandboxBaseWithDelayedSettings {
 
         overrideFunction(window.Headers.prototype, 'values', FetchSandbox._valuesWrapper);
 
-        overrideFunction(window.Headers.prototype, 'forEach', function (...args) {
+        overrideFunction(window.Headers.prototype, 'forEach', function (...args: [(value: any, name: any, headers: any) => void, any?]) {
             const callback = args[0];
 
             if (typeof callback === 'function') {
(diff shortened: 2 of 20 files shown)


src/client/sandbox/native-methods.ts
```diff
@@ -508,9 +512,9 @@ class NativeMethods {
         this.getAttributeNS                = nativeElement.getAttributeNS;
         this.insertAdjacentHTML            = nativeElement.insertAdjacentHTML;
         this.insertBefore                  = nativeElement.insertBefore;
-        this.insertCell                    = createElement('tr').insertCell;
-        this.insertTableRow                = createElement('table').insertRow;
-        this.insertTBodyRow                = createElement('tbody').insertRow;
+        this.insertCell                    = (<HTMLTableRowElement>createElement('tr')).insertCell;
```
AndreyBelym, line 515: It's better to always use `as` to convert types, you'll have to enable a TSLint rule that forces it at some point.

src/client/dom-processor/client-dom-adapter.ts
```diff
@@ -89,8 +89,8 @@ export default class ClientDomAdapter extends BaseDomAdapter {
         }
     }
 
-    getProxyUrl (): string {
-        return getProxyUrl.apply(null, arguments);
+    getProxyUrl (...args): string {
+        return getProxyUrl.apply(null, args as [string, any?]);
```
AndreyBelym, line 93: Maybe we can just ensure type in the function arguments directly? There is also a TSLint rule in TestCafe that requires all function arguments to be typed.

src/client/sandbox/fetch.ts
```diff
@@ -167,7 +167,7 @@ export default class FetchSandbox extends SandboxBaseWithDelayedSettings {
 
             // NOTE: Safari processed the empty `fetch()` request without `Promise` rejection (GH-1613)
             if (!args.length && !browserUtils.isSafari)
-                return nativeMethods.fetch.apply(this);
+                return nativeMethods.fetch.apply(this, [] as unknown as [RequestInfo, RequestInit?]);
```
AndreyBelym, line 170: I need to think if we can avoid this double conversion :)

AndreyBelym (commented):
Thanks for doing this important task

AndreyBelym (approved):

src/client/sandbox/node/window.ts
```diff
@@ -529,7 +529,7 @@ export default class WindowSandbox extends SandboxBase {
                     scriptURL = getProxyUrl(scriptURL, { resourceType: stringifyResourceType({ isScript: true }) });
 
                 if (isCalledWithoutNewKeyword)
-                    return nativeMethods.Worker.apply(this, arguments);
+                    return nativeMethods.Worker.apply(this, arguments as unknown as [string | URL, WorkerOptions?]);
```
LavrovArtem, line 532:
```ts
overrideConstructor(window, 'Worker', function WorkerWrapper (...args: [string | URL, WorkerOptions?]) {
    const isCalledWithoutNewKeyword = constructorIsCalledWithoutNewKeyword(this, WorkerWrapper);

    if (arguments.length === 0)
    // @ts-ignore
        return isCalledWithoutNewKeyword ? nativeMethods.Worker() : new nativeMethods.Worker();

    if (isCalledWithoutNewKeyword)
        return nativeMethods.Worker.apply(this, args);

    let scriptURL = args[0];
    
    if (typeof scriptURL === 'string')
        scriptURL = getProxyUrl(scriptURL, { resourceType: stringifyResourceType({ isScript: true }) });

    // @ts-ignore
    const worker = arguments.length === 1 ? new nativeMethods.Worker(scriptURL) : new nativeMethods.Worker(scriptURL, args[0]);

```

src/client/sandbox/node/window.ts
```diff
@@ -820,7 +820,7 @@ export default class WindowSandbox extends SandboxBase {
                     nativeMethods.formDataAppend.call(this, name, value, value.name);
                 }
                 else
-                    nativeMethods.formDataAppend.apply(this, arguments);
+                    nativeMethods.formDataAppend.apply(this, arguments as unknown as [string, string | Blob, string?]);
```
LavrovArtem, line 823:
```ts
overrideFunction(window.FormData.prototype, 'append', function (...args: [string, string | Blob, string?]) {
    const [name, value] = args;

    // NOTE: We should not send our hidden input's value along with the file info,
    // because our input may have incorrect value if the input with the file has been removed from DOM.
    if (name === INTERNAL_ATTRS.uploadInfoHiddenInputName)
        return;

    // NOTE: If we append our file wrapper to FormData, we will lose the file name.
    // This happens because the file wrapper is an instance of Blob
    // and a browser thinks that Blob does not contain the "name" property.
    if (args.length === 2 && isBlob(value) && 'name' in value)
        args[2] = value['name'] as string;

    nativeMethods.formDataAppend.apply(this, args);
});
```

src/client/sandbox/node/window.ts
```diff
@@ -1540,7 +1540,7 @@ export default class WindowSandbox extends SandboxBase {
                 },
 
                 setter: function (value) {
-                    return nativeMethods.windowOriginSetter.call(this, value);
+                    return nativeMethods.windowOriginSetter.apply(this, [value] as unknown as []);
```
LavrovArtem, line 1543:
```diff
-windowOriginSetter: () => any;
+windowOriginSetter: (this: Window, value: string) => void;
```

src/client/utils/feature-detection.ts
```diff
@@ -36,7 +36,7 @@ if (nativeMethods.createElement) {
     // NOTE: In the Edge 17, the getNamedItem method of attributes object is not enumerable
     attrGetNamedItemIsNotEnumerable = nativeMethods.objectGetOwnPropertyDescriptor
     // @ts-ignore
-        .call(window.Object, NamedNodeMap.prototype, 'getNamedItem');
+        .call(window.Object, NamedNodeMap.prototype, 'getNamedItem') as boolean;
```
LavrovArtem, line 39:
```ts
attrGetNamedItemIsNotEnumerable = !!nativeMethods.objectGetOwnPropertyDescriptor(NamedNodeMap.prototype, 'getNamedItem');
```

src/processing/dom/parse5-dom-adapter.ts
```diff
@@ -82,8 +82,8 @@ export default class Parse5DomAdapter extends BaseDomAdapter {
         return this.isIframe;
     }
 
-    getProxyUrl (): string {
-        return urlUtils.getProxyUrl.apply(urlUtils, arguments);
+    getProxyUrl (...args: [string, any]): string {
+        return urlUtils.getProxyUrl.apply(urlUtils, args);
```
LavrovArtem, line 86:
```ts
return urlUtils.getProxyUrl(...args);
```

LavrovArtem (changes requested):

src/client/sandbox/node/document/title-storage.ts
```diff
@@ -23,7 +23,9 @@ export default class DocumentTitleStorage extends EventEmitter {
         if (firstTitle)
             return firstTitle;
 
-        firstTitle = nativeMethods.createElement.call(this._document, 'title');
+        const createElement = (tagName => nativeMethods.createElement.call(this._document, tagName)) as Document['createElement'];
```
LavrovArtem, line 26: I don't think that we need to create function for the type casting, in this case the `as HTMLTitleElement` is enough.

src/client/sandbox/native-methods.ts
```diff
@@ -520,7 +520,7 @@ class NativeMethods {
         this.hasAttributeNS                = nativeElement.hasAttributeNS;
         this.hasAttributes                 = nativeElement.hasAttributes;
         this.anchorToString                = win.HTMLAnchorElement.prototype.toString;
-        this.matches                       = nativeElement.matches || nativeElement.msMatchesSelector;
+        this.matches                       = nativeElement.matches || (nativeElement as HTMLElement as IEHTMLElement).msMatchesSelector;
```
LavrovArtem, line 523: It's unnecessary type casting, see comments for the `src/typings/extend-html-element.d.ts` file below.

src/client/sandbox/node/window.ts
```diff
@@ -1544,7 +1543,7 @@ export default class WindowSandbox extends SandboxBase {
                 },
 
                 setter: function (value) {
-                    return nativeMethods.windowOriginSetter.call(this, value);
+                    return nativeMethods.windowOriginSetter.apply(this, [value]);
```
LavrovArtem, line 1546: This change is unnecessary.

src/typings/extend-html-element.d.ts
```diff
@@ -0,0 +1,3 @@
+interface IEHTMLElement extends HTMLElement {
```
LavrovArtem, line 1:
```ts
interface HTMLElement {
```

src/typings/extend-html-element.d.ts
```diff
@@ -0,0 +1,3 @@
...
+    msMatchesSelector: Function;
```
LavrovArtem, line 2: Add comment that this type is extended for ie11.

LavrovArtem (changes requested):

src/client/sandbox/node/document/title-storage.ts
```diff
@@ -23,7 +23,7 @@ export default class DocumentTitleStorage extends EventEmitter {
         if (firstTitle)
             return firstTitle;
 
-        firstTitle = nativeMethods.createElement.call(this._document, 'title');
+        firstTitle = nativeMethods.createElement('title') as HTMLTitleElement;
```
LavrovArtem, line 26: image

LavrovArtem (changes requested):

LavrovArtem (approved):

AndreyBelym (approved):

PR #2511, opened 2020-12-18
An iframe with the sandbox attribute that doesn't contain 'allow-same-origin' returns window.origin as 'null' now

LavrovArtem:


miherlosev (approved):

src/client/sandbox/native-methods.ts
```diff
@@ -1050,6 +1051,12 @@ class NativeMethods {
             this.tokenListToggle   = win.DOMTokenList.prototype.toggle;
             this.tokenListContains = win.DOMTokenList.prototype.contains;
 
+            const tokenListValueDescriptor = win.Object.getOwnPropertyDescriptor(win.DOMTokenList.prototype, 'value');
+
+            if (tokenListValueDescriptor)
```
Farfurix, line 1056:
```suggestion
            // NOTE: IE11 doesn't support the 'value' property of the DOMTokenList interface
            if (tokenListValueDescriptor)
```

Farfurix (approved):

PR #2533, opened 2020-12-30
The global context info calculated once per context

LavrovArtem:


src/client/sandbox/native-methods.ts
```diff
@@ -1,5 +1,5 @@
 /*global Document, Window */
-import getGlobalContextInfo from '../utils/global-context-info';
+import globalContextInfo from '../utils/global-context-info';
```
miherlosev, line 2: I guess it's better to change the `getGlobalContextInfo` function in the following way

```js
const globalContenxtInfo = getGlobalContextInfoInternal () // calculate once

export default function getGlobalContextInfo () {
    return globalContenxtInfo;
}
```

miherlosev (changes requested):

miherlosev (approved):

Farfurix (approved):

PR #2536, opened 2020-12-31
Handle undefined/wrong-type error in service request

AlexKamaev:
This is a duplicate of https://github.com/DevExpress/testcafe-hammerhead/pull/2481/ with rebase, tests and some modifications

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -3,6 +3,13 @@ import RequestPipelineContext from '../request-pipeline/context';
 import { IncomingMessage, OutgoingHttpHeaders } from 'http';
 import { ServiceMessage } from '../typings/proxy';
 import RequestOptions from '../request-pipeline/request-options';
+import errToString from './err-to-string'
+
+function getIncorrectErrorTypeMessage (err: object) {
+    const errType = typeof err;
+
+    return `The "${errToString(err)}" error of the "${errType}" type was passed. Make sure that service message handlers throw errors correctly.`;
+}
 
 debug.formatters.i = (ctx: RequestPipelineContext): string => {
     const stringifiedInfoArr: string[] = [];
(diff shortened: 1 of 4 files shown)


AlexKamaev:
@testcafe-build-bot retest

AlexKamaev:
it's ready for review. Tests are failed because of some problems with SauceLabs

AlexKamaev:
@testcafe-build-bot retest

src/utils/logger.ts
```diff
@@ -3,6 +3,13 @@ import RequestPipelineContext from '../request-pipeline/context';
 import { IncomingMessage, OutgoingHttpHeaders } from 'http';
 import { ServiceMessage } from '../typings/proxy';
 import RequestOptions from '../request-pipeline/request-options';
+import errToString from './err-to-string'
+
+function getIncorrectErrorTypeMessage (err: object) {
+    const errType = typeof err;
+
+    return `The "${errToString(err)}" error of the "${errType}" type was passed. Make sure that service message handlers throw errors correctly.`;
```
miherlosev, line 11: I propose another error message.

```js
Incorrect "${errType}" error was passed to the error handler. Make sure that service message handlers throw errors correctly.
```

miherlosev (changes requested):

src/utils/logger.ts
AlexKamaev: Don't you want to show the error text if it exists?
for example:

```JS
throw 'string error';
```
`The "string error" error of the "string" type was passed. `

miherlosev: Ok. Let it be.

miherlosev (approved):

src/utils/logger.ts
AlexKamaev: @DIRECTcut check this message please:
`The "${errToString(err)}" error of the "${errType}" type was passed. Make sure that service message handlers throw errors correctly.`

AlexKamaev:
@LavrovArtem review please

src/utils/logger.ts
DIRECTcut: It's unclear what `correctly` means. For example, does it mean:

`The "${errToString(err)}" error of the "${errType}" type was passed. Make sure that service message handlers throw errors of a correct type.`

AlexKamaev: the error should be of Error type.
In JS you can make the following:
```JS
throw 1
```
but the correct usage is:
```JS
throw new Error('1');
```
What to do you think if I change it to
`Make sure that service message handlers throw errors of the Error type`

LavrovArtem (approved):

DIRECTcut (approved):

PR #2552, opened 2021-01-14
Should cache destination responses

miherlosev:
Cache resources (at this stage only scripts and styles) between sessions as it does browser: analyze caching headers and decide about sending a real HTTP request. The 'http-cache-semantics' module is used to make a decision about the necessity of refreshing request data.


miherlosev:
FPR

src/request-pipeline/context.ts
```diff
@@ -296,6 +299,13 @@ export default class RequestPipelineContext {
             .map(userScript => userScript.url);
     }
 
+    private async _getDestResBody (res: IncomingMessageLike | http.IncomingMessage | FileStream): Promise<Buffer> {
+        if (IncomingMessageLike.isIncomingMessageLike(res))
+            return (res as IncomingMessageLike).getBody();
+
+        return await fetchBody(this.destRes);
```
wentwrong, line 306: Minor: I think we can get rid of using `await` here. Reason: [Why Using `return await` Is a Bad Idea?](http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea)

miherlosev: Ok

miherlosev:
FPR

src/request-pipeline/incoming-message-like.ts
```diff
@@ -52,4 +52,26 @@ export default class IncomingMessageLike extends Readable {
...
+        const { headers, trailers, statusCode } = res;
+
+        return new IncomingMessageLike({
+            headers,
+            trailers,
+            statusCode
+        });
+    }
+
+    static isIncomingMessageLike (obj: object): boolean {
```
wentwrong, line 74:
```suggestion
    static isIncomingMessageLike (obj: object): obj is IncomingMessageLike {
```

src/request-pipeline/context.ts
```diff
@@ -296,6 +299,13 @@ export default class RequestPipelineContext {
             .map(userScript => userScript.url);
     }
 
+    private async _getDestResBody (res: IncomingMessageLike | http.IncomingMessage | FileStream): Promise<Buffer> {
+        if (IncomingMessageLike.isIncomingMessageLike(res))
+            return (res as IncomingMessageLike).getBody();
```
wentwrong, line 304:
```suggestion
            return res.getBody();
```
(related to the suggestion below)

wentwrong (approved):

src/request-pipeline/destination-request/index.ts
```diff
@@ -58,31 +88,23 @@ export default class DestinationRequest extends EventEmitter implements Destinat
...
-                });
-            }
+            if (this.cache) {
+                const cachedResponse = requestCache.getResponse(this.opts);
+
+                if (cachedResponse) {
+                    // NOTE: To store async order of the 'response' event
+                    setTimeout(() => {
+                        this._emitOnResponse(cachedResponse.res);
+                    }, 0);
```
LavrovArtem, line 99:
```suggestion
                    setImmediate(() => this._emitOnResponse(cachedResponse.res));
```

src/request-pipeline/incoming-message-like.ts
```diff
@@ -52,4 +52,26 @@ export default class IncomingMessageLike extends Readable {
...
+    }
+
+    getBody (): Buffer {
+        return this._body;
+    }
+
+    static createFrom (res: IncomingMessage): IncomingMessageLike {
+        const { headers, trailers, statusCode } = res;
+
+        return new IncomingMessageLike({
```
LavrovArtem, line 67: also add the `rawHeaders`

LavrovArtem (approved):

PR #2559, opened 2021-01-20
fix `Service Worker overwriting window object causes crash` (close #2538)

### LINKED ISSUE #2538: Service Worker overwriting window object causes crash
Hi All,

First i am very sorry, for the bad bug report. The scenario is happening specifically with SAP Analytics Cloud (Major Analytics solution by SAP). Everything here is of course properitary. I hope the technical details and possible solution are good enough to understand the issue.

Generally: Cloud Analytics is heavily relying on service workers. This seems to cause issues.
 
### Summary
Crash during startup, in case a service worker is (partly) filling the window object.

### Technical Reason for crash
Native Methods nodeFirstChildGetter is undefined when using XHR requests in service-worker.
grafik

As a service worker does not have a window.Node/HTMLElement object, the resolve method shouldn't be called after all. The code for this logic is already available - The problem seems to be that the caller (resolve) is not detecting that it is inside a worker thread (see following screenshot).
grafik

The reason for this, is that the window object was partly overwritten (with DedicatedWorkerGlobalScope - so the actual "self" object). If I open the application directly, it shows the same behavior inside the worker thread. Therefore the window object is overwritten / defined by the service worker.
grafik

Now I am really not sure about the service worker specification. Is a service worker allowed to overwrite any window method? Adjusting the code to:
grafik
solves the issue and the application simply starts up..
Alternative (yes the ts-ignore is ofc bullshit):
grafik

What do you think? Is that a valid solution (maybe including Node and other window objects you are accessing..)

Thanks for any feedback,
Timo

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

Farfurix (approved):

PR #2562, opened 2021-01-21
Fix the Chrome 88 domParser test, remove unnecessary old IE conditions

Farfurix:
## Purpose
Chrome 88 throws `TypeError` in the case of three `domParser` arguments:
```js
var domParser = new DOMParser();

domParser.parseFromString('<a href="/path">Anchor</a>', 'text/html', 'third argument');
```

## References
https://developer.mozilla.org/en-US/docs/Web/API/DOMParser#domxrefdomparser.parsefromstring_1:
> Parameters
This method has 2 parameters (both required):
**string**
The DOMString to be parsed. It must contain either HTML, xml, xhtml+xml, or svg document.
**mimeType**
A DOMString. This string determines a class of the method's return value. The possible values are the following:
...

### DIFF
--- a/test/client/fixtures/sandbox/node/methods-test.js
+++ b/test/client/fixtures/sandbox/node/methods-test.js
@@ -303,17 +303,10 @@ if (window.DOMParser) {
 
             strictEqual(nativeMethods.getAttribute.call(anchor, 'href'), '/path');
 
-            if (!browserUtils.isChrome || browserUtils.version < 88) {
-                parsedDocument = domParser.parseFromString(htmlStr, 'text/html', 'third argument');
-                anchor = parsedDocument.querySelector('a');
+            parsedDocument = domParser.parseFromString(htmlStr, 'text/html');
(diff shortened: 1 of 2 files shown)


LavrovArtem (approved):

test/client/fixtures/sandbox/node/methods-test.js
```diff
@@ -303,10 +303,17 @@ if (window.DOMParser) {
 
             strictEqual(nativeMethods.getAttribute.call(anchor, 'href'), '/path');
 
-            parsedDocument = domParser.parseFromString(htmlStr, 'text/html', 'third argument');
-            anchor         = parsedDocument.querySelector('a');
+            if (!browserUtils.isChrome || browserUtils.version < 88) {
```
miherlosev, line 306: Just remove the third parameter.

miherlosev (changes requested):

miherlosev (approved):

PR #2579, opened 2021-02-03
refactor(client): make sharedUtils accessible

wentwrong:
## Purpose
As a preparatory step towards migrating from Webmake to Rollup module bundler (#2567), we need to get rid of using the `hammerhead.get` method. It stores a reference to the `require` function and is used for importing internal hammerhead submodules in the client tests. Such functionality cannot be achieved using Rollup, because Rollup doesn't preserve the structure of the exported submodules inside generated bundle.

## Approach
Place submodules which were exposed in the client tests into the `sharedUtils` property of the `Hammerhead` class, also add the `settings` property and use existing `hammerhead.utils` where is appropriate to use them.

### DIFF
--- a/src/client/index.ts
+++ b/src/client/index.ts
@@ -13,6 +13,11 @@ import * as typeUtils from './utils/types';
 import * as positionUtils from './utils/position';
 import * as styleUtils from './utils/style';
 import trim from '../utils/string-trim';
+import * as sharedCookieUtils from '../utils/cookie';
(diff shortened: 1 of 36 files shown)


wentwrong:
@testcafe-build-bot retest

wentwrong:
@testcafe-build-bot retest

LavrovArtem (approved):

PR #2581, opened 2021-02-04
refactor: improve types to avoid issues with tsc

wentwrong:
## Purpose
I faced some issues with the types when working on #2567.

## Approach
* add type for the `pinkie` module (before that, this was `any` and so there was no type-checking)
* use default import for the `urlResolver` utility (there are no named exports in the module, only default one)
* fix little bug originated because there is no type-checking of the server scripts (only transpiling via babel)

### DIFF
--- a/src/client/sandbox/cookie/window-sync.ts
+++ b/src/client/sandbox/cookie/window-sync.ts
@@ -165,12 +165,12 @@ export default class WindowSync {
         }
 
         if (syncMessages.length) {
-            let syncMessagesPromise = Promise.all(syncMessages);
+            const syncMessagesPromise = Promise.all(syncMessages);
 
             if (this._win === topOpenerWindow)
-                syncMessagesPromise = syncMessagesPromise.then(() => this._removeSyncCookie(cookies));
+                return syncMessagesPromise.then(() => this._removeSyncCookie(cookies));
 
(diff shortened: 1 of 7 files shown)


src/client/sandbox/cookie/window-sync.ts
```diff
@@ -165,12 +165,12 @@ export default class WindowSync {
...
         if (syncMessages.length) {
-            let syncMessagesPromise = Promise.all(syncMessages);
+            const syncMessagesPromise = Promise.all(syncMessages);
 
             if (this._win === topOpenerWindow)
-                syncMessagesPromise = syncMessagesPromise.then(() => this._removeSyncCookie(cookies));
+                return syncMessagesPromise.then(() => this._removeSyncCookie(cookies));
 
-            return syncMessagesPromise;
+            return syncMessagesPromise.then();
```
LavrovArtem, line 173: Why do we need this `.then()`?

wentwrong: Because of the return type of this method. It's declared as `Promise<void>`, but the `syncMessagesPromise` variable type is `Promise<void[]>` (because of `Promise.all` usage).

miherlosev (approved):

LavrovArtem (approved):

PR #2582, opened 2021-02-05
Should not opened in window if the default behavior was prevented inside the event handler (for #5910 in TestCafe)

AlexKamaev:
New mechanism of `postHandlers` which will be called after all user handlers are called.
This mechanism allows us to prevent opening of new window in multiple window mode, if the event was prevented inside on the user listeners

### DIFF
--- a/src/client/sandbox/child-window/index.ts
+++ b/src/client/sandbox/child-window/index.ts
@@ -71,7 +71,7 @@ export default class ChildWindowSandbox extends SandboxBase {
             return;
 
         this._listeners.initElementListening(el, ['click']);
-        this._listeners.addPostHandler(el, ['click'], e => {
+        this._listeners.addInternalEventAfterListener(el, ['click'], e => {
             if (e.defaultPrevented)
                 return;
 
@@ -104,7 +104,7 @@ export default class ChildWindowSandbox extends SandboxBase {
             return;
 
         this._listeners.initElementListening(window, ['submit']);
-        this._listeners.addInternalEventListener(window, ['submit'], e => {
+        this._listeners.addInternalEventBeforeListener(window, ['submit'], e => {
             if (!domUtils.isFormElement(e.target))
                 return;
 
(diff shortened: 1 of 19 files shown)


src/client/sandbox/child-window/index.ts
```diff
@@ -124,6 +106,31 @@ export default class ChildWindowSandbox extends SandboxBase {
         });
     }
 
+    _handleClickOnLinkWithTargetBlankAttr (window: Window): void {
+        if (!settings.get().allowMultipleWindows)
+            return;
+
+        window.addEventListener('click', e => {
```
miherlosev, line 113: Use a native method for `addEventListener` here.
@LavrovArtem if it's possible add this method to `eslint-plugin-hammerhead`.

miherlosev (changes requested):

src/client/sandbox/child-window/index.ts
AlexKamaev: this PR is still in DRAFT. I discussed this approach with @LavrovArtem and we decided to modify it

miherlosev: Ok

AlexKamaev: ready, please review

src/client/sandbox/event/listening-context.ts
```diff
@@ -29,6 +29,7 @@ export function addListeningElement (el, events) {
                 internalHandlers:     [],
                 outerHandlers:        [],
                 outerHandlersWrapper: null,
+                postHandlers:         [],
```
miherlosev, line 32: I don't like this name because it can be understood in several ways.

It's internal hammerhead handlers, which are called last.

I guess, the `internalHandlers` are handlers which are called before users' scripts.
So, I propose to clean up the code and perform some renames:
`internalHandlers -> internalBeforeHanlders`
`postHandlers -> internlaAfterHanlders`

@LavrovArtem What do you think about it?

miherlosev (changes requested):

src/client/sandbox/event/listening-context.ts
AlexKamaev: I like this suggestion

AlexKamaev:
Tests in TestCafe failed since TestCafe requires update as well:
https://github.com/DevExpress/testcafe/pull/5928

miherlosev (approved):

PR #2584, opened 2021-02-08
refactor(client): get rid of hammerhead.get usage

wentwrong:
## Purpose
As a preparatory step towards migrating from Webmake to Rollup module bundler (#2567), we need to get rid of using the `hammerhead.get` method. It stores a reference to the `require` function and is used for importing internal hammerhead submodules in the client tests. Such functionality cannot be achieved using Rollup, because Rollup doesn't preserve the structure of the exported submodules inside generated bundle.

## Approach
Put all submodules exposed in the tests into the "Hammerhead" class properties. _This class became too large for single file because of many imported submodules. Maybe it's worth to separate into different classes? Any suggestions on how to structure it better would be nice._

PR #2585, opened 2021-02-08
chore: migrate from webmake to rollup

### LINKED ISSUE #2567: Use Rollup in testcafe-hammerhead

wentwrong:
## Purpose
* use the typescript compiler instead of babel for both client and server
* use the rollup module bundler instead of webmake (bundles are smaller; no vulns; no need to maintain "gulp-webmake" package; unification of used tools for further monorepo migration)

## Approach
* get rid of babel (tsc for compiling both client and server code)
* get rid of the `json-hammerhead` package
* suppress rollup warnings about circular deps and missed name of IIFE bundles
* compile the `acorn-hammerhead` package because it has non ES5-compatible code

## References
closes #2567

## Pre-Merge TODO
- [x] Merge #2581 and #2584


miherlosev (approved):

rollup.config.js
```diff
@@ -0,0 +1,47 @@
...
+                'src/**/*.ts',
+                // NOTE: transpile the acorn-hammerhead package because it has non-ES5 compatible code
+                'node_modules/acorn-hammerhead/**/*.js'
+            ],
+            rollupCommonJSResolveHack: true
+        }),
+        commonjs()
+    ],
+    onwarn (warning, rollupWarn) {
+        if (warning.code === 'CIRCULAR_DEPENDENCY')
```
miherlosev, line 37: Add comment that describes for which modules these warnings will be suppressed.

miherlosev (changes requested):

rollup.config.js
wentwrong: This code was intended to suppress this warning message from Rollup:
> (!) Circular dependencies
> src/client/sandbox/native-methods.ts -> src/client/utils/overriding.ts -> src/client/sandbox/native-methods.ts
> src/client/utils/dom.ts -> src/client/utils/url.ts -> src/client/utils/destination-location.ts -> src/client/utils/dom.ts
> src/client/utils/destination-location.ts -> src/client/utils/url-resolver.ts -> src/client/utils/destination-location.ts
> ...and 14 more

But it will suppress warning not only from our code, but from external packages too.

miherlosev: Ok. Then it's not necessary.

src/client/index.ts
```diff
@@ -74,7 +73,7 @@ class Hammerhead {
     processScript: Function;
     get: Function;
     Promise: any;
-    json: any;
+    json: typeof JSON;
```
miherlosev, line 76: typeof JSON is a 'object' string.

We can remove this export since we are using a native JSON's methods.

src/client/sandbox/event/message.ts
```diff
@@ -51,7 +50,7 @@ export default class MessageSandbox extends SandboxBase {
     private static _getMessageData (e) {
         const rawData = isMessageEvent(e) ? nativeMethods.messageEventDataGetter.call(e) : e.data;
 
-        return typeof rawData === 'string' ? parseJSON(rawData) : rawData;
+        return typeof rawData === 'string' ? JSON.parse(rawData) : rawData;
```
miherlosev, line 53: Store JSON's methods in the `nativeMethods` module and use them in the appropriate places.

miherlosev (changes requested):

src/client/index.ts
wentwrong: > typeof JSON is a 'object' string.

it's true for javascript. But in typescript it's type of the `JSON` object - https://github.com/microsoft/TypeScript/blob/79c72751df2aaed2b2ba8caa4170895c343a3d5b/lib/lib.es5.d.ts#L1052
but I agree with the rest.

src/client/sandbox/event/message.ts
wentwrong: We can't store JSON's methods in the `nativeMethods` module, because it's client-side module. But we also have shared code which uses these methods both on the client and on the server side. So I stored them as global variables in the shared utility.

src/utils/json.ts
```diff
@@ -1,4 +1,37 @@
...
+        return obj instanceof Node;
+
+    return typeof obj.nodeType === 'number' && typeof obj.nodeName === 'string';
+}
+
+function isJQueryObj (obj) {
+    return !!(obj && obj.jquery);
+}
+
+export function isSerializable (value) {
```
wentwrong, line 22: Because we decided to get rid of the `json-hammerhead` completely, I had to move these functions to our codebase. It's used in the `testcafe-legacy-api` package: https://github.com/DevExpress/testcafe-legacy-api/blob/60cd17a636359d14e4ffbec4316421e9ca1ab3b5/src/client/step-iterator.js#L69

miherlosev (approved):

PR #2587, opened 2021-02-16
fix hover for label with associated control (testcafe gh-5916)

AlexKamaev:
The cause of the issue is that if user hovers a label associated with input element (using htmlFor atrtibute) then input:hover style applies as well.

i.e.
```HTML
<label for='inp'>label<label>
<input id='inp'>
<div>test</div>
```
```CSS
input:hover ~ div {
background-color: red;
}
```
If you hover label then div becomes red

### DIFF
--- a/src/client/utils/dom.ts
+++ b/src/client/utils/dom.ts
@@ -877,3 +877,9 @@ export function isInputWithoutSelectionProperties (el): boolean {
 
     return !hasSelectionProperties;
 }
+
+export function getAssociatedElement (el: HTMLLabelElement): HTMLElement {
+    const doc = findDocument(el);
+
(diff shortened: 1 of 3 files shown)


src/client/utils/dom.ts
```diff
@@ -877,3 +877,9 @@ export function isInputWithoutSelectionProperties (el): boolean {
 
     return !hasSelectionProperties;
 }
+
+export function getAssociatedElement (el: HTMLLabelElement): HTMLElement {
+    const doc = findDocument(el);
+
+    return el.control || el.htmlFor && doc.getElementById(el.htmlFor);
```
miherlosev, line 884: Use the native `getElementById` method.

miherlosev (changes requested):

miherlosev (approved):

LavrovArtem (approved):

PR #2590, opened 2021-02-26
We should not break a user script which uses new browser API if we are not overridden it

LavrovArtem:


LavrovArtem:
@testcafe-build-bot retest

miherlosev (approved):

PR #2618, opened 2021-04-15
fix: ready to init event in iframe with srcdoc

wentwrong:
## Purpose
The ready to init event raised too early in the iframes with the srcdoc attribute. Because of that, TestCafe scripts can't properly initiate themselves. ([init](https://github.com/DevExpress/testcafe/blob/faec4a93682d51a6f51909514b437236899b08c1/src/client/core/page-unload-barrier.js#L82) method of the pageUnloadBarrier gets the 'unload' event of the 'about:blank' document and can't properly work further).

## Approach
Don't raise the ready to init event after processing an iframe. Instead, this event will be raised later by self removing script of the iframe with the srcdoc attribute.

## References
Functional test in testcafe - https://github.com/DevExpress/testcafe/pull/6142

## Pre-Merge TODO

### DIFF
--- a/test/client/fixtures/sandbox/iframe-test.js
+++ b/test/client/fixtures/sandbox/iframe-test.js
@@ -145,6 +145,29 @@ if (nativeMethods.iframeSrcdocGetter) {
         strictEqual(iframe.getAttribute('srcdoc'), html);
         strictEqual(nativeMethods.getAttribute.call(iframe, 'srcdoc'), htmlUtils.processHtml(html, { isPage: true }).replace(/(sessionId)/, '$1!i'));
     });
+
+    asyncTest('ready to init event should be raised after the document was initialized', function () {
+        var iframeLoadingEventRaised = false;
+
+        var handler = function (iframe) {
+            iframeLoadingEventRaised = true;
+
+            strictEqual(iframe.contentDocument.location.href, 'about:srcdoc');
+        };
+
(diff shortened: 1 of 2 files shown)


test/client/fixtures/sandbox/iframe-test.js
```diff
@@ -145,6 +145,29 @@ if (nativeMethods.iframeSrcdocGetter) {
...
+
+        var handler = function (iframe) {
+            iframeLoadingEventRaised = true;
+
+            strictEqual(iframe.contentDocument.location.href, 'about:srcdoc');
+        };
+
+        iframeSandbox.on(iframeSandbox.RUN_TASK_SCRIPT_EVENT, handler);
+
+        var $iframe = $('<iframe srcdoc="<h1>simple markup</h1>">').appendTo('body');
```
miherlosev, line 160: Don't use JQuery in client tests

miherlosev (changes requested):

wentwrong:
Related functional test in TestCafe - https://github.com/DevExpress/testcafe/pull/6142#issuecomment-821275225
All tests are green (except `test-functional-local-headless-chrome`, but this task is green in another run - https://travis-ci.com/github/testcafe-build-bot/testcafe/jobs/499428336)

miherlosev (approved):

PR #2625, opened 2021-04-28
should not open new window if link has the `download` attribute (closes #6132)

AlexKamaev:
The click on a link with the `download` attribute, should not lead to new window opening (even when link has the target=blank attribute)

### DIFF
--- a/src/client/sandbox/child-window/index.ts
+++ b/src/client/sandbox/child-window/index.ts
@@ -28,9 +28,10 @@ export default class ChildWindowSandbox extends SandboxBase {
     }
 
     private static _shouldOpenInNewWindowOnElementAction (el: HTMLLinkElement | HTMLAreaElement | HTMLFormElement, defaultTarget: string): boolean {
-        const target = this._calculateTargetForElement(el);
+        const target               = this._calculateTargetForElement(el);
+        const hasDownloadAttribute = nativeMethods.getAttribute.call(el, 'download');
 
-        return this._shouldOpenInNewWindow(target, defaultTarget);
+        return hasDownloadAttribute ? false : this._shouldOpenInNewWindow(target, defaultTarget);
     }
 
     private static _shouldOpenInNewWindow (target: string, defaultTarget: string): boolean {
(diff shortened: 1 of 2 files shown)


src/client/sandbox/child-window/index.ts
```diff
@@ -28,9 +28,10 @@ export default class ChildWindowSandbox extends SandboxBase {
     }
 
     private static _shouldOpenInNewWindowOnElementAction (el: HTMLLinkElement | HTMLAreaElement | HTMLFormElement, defaultTarget: string): boolean {
-        const target = this._calculateTargetForElement(el);
+        const target               = this._calculateTargetForElement(el);
+        const hasDownloadAttribute = nativeMethods.getAttribute.call(el, 'download');
```
miherlosev, line 32: I guess we can use fast forward here.
```js
private static _shouldOpenInNewWindowOnElementAction (....) {
     const hasDownloadAttribute = nativeMethods.getAttribute.call(el, 'download');

     if (hasDownloadAttribute)
         return false;

  ...
```

miherlosev (changes requested):

src/client/sandbox/child-window/index.ts
AlexKamaev: Well, me personally, I do not see any advantages or disadvantages of any of these two approaches

miherlosev: Advantage is code readability.

Compare:
```js
private static _shouldOpenInNewWindowOnElementAction (): boolean {
    const target               = this._calculateTargetForElement(el);
    const hasDownloadAttribute = nativeMethods.getAttribute.call(el, 'download');

    return hasDownloadAttribute ? false : this._shouldOpenInNewWindow(target, defaultTarget);
}


private static _shouldOpenInNewWindowOnElementAction (): boolean {
     const hasDownloadAttribute = nativeMethods.getAttribute.call(el, 'download');

    if (hasDownloadAttribute)
        return false;

    const target = this._calculateTargetForElement(el);

    return this._shouldOpenInNewWindow(target, defaultTarget);
}


```

miherlosev (approved):

LavrovArtem (approved):

PR #2632, opened 2021-05-13
fix for-of destructuring duplicate declarator issue with Object.ArrayPattern

AlexKamaev:
Approach:
If the `for-of statement` has the `ArrayPattern` declaration in the left block, then I try to find the similar variable declaration inside the `for-of` body. If the similar declaration is found this means that we cannot use the previosly declared variable, so we need to rename it.

### DIFF
--- a/src/processing/script/transformers/for-of.ts
+++ b/src/processing/script/transformers/for-of.ts
@@ -2,7 +2,17 @@
 // WARNING: this file is used by both the client and the server.
 // Do not use any browser or node-specific API!
 // -------------------------------------------------------------
-import { Declaration, ForOfStatement, Pattern, Statement } from 'estree';
+import {
+    BlockStatement,
+    Declaration,
+    ForOfStatement,
+    Identifier,
+    Node,
+    Pattern,
+    Statement,
+    VariableDeclaration
+} from 'estree';
+
 import { Transformer } from './index';
 import { Syntax } from 'esotope-hammerhead';
 import {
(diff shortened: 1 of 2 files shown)


src/processing/script/transformers/for-of.ts
```diff
@@ -15,6 +25,61 @@ import {
 import replaceNode from './replace-node';
 import TempVariables from './temp-variables';
 
+function walkDeclarators (node: BlockStatement, action: Function): void {
```
LavrovArtem, line 28:
```suggestion
function walkDeclarators (node: BlockStatement, action: (identifier: Identifier) => void): void {
```

src/processing/script/transformers/for-of.ts
```diff
@@ -15,6 +25,61 @@ import {
...
+        if (identifier.type === Syntax.Identifier)
+            action(identifier);
+    }
+}
+
+function replaceDuplicateDeclarators (forOfNode: ForOfStatement) {
+    const forOfLeft      = forOfNode.left as VariableDeclaration;
+    const nodesToReplace = [];
+
+    const isArrayPatternDeclaration = forOfLeft.declarations.length && forOfLeft.declarations[0].id.type === Syntax.ArrayPattern;
```
LavrovArtem, line 61:
```suggestion
    const isArrayPatternDeclaration = forOfLeft.declarations[0]?.id.type === Syntax.ArrayPattern;
```

src/processing/script/transformers/for-of.ts
```diff
@@ -15,6 +25,61 @@ import {
...
+    const isBlockStatement          = forOfNode.body.type === Syntax.BlockStatement;
+
+    if (!isArrayPatternDeclaration || !isBlockStatement)
+        return;
+
+    const leftDeclaration = forOfLeft.declarations[0].id as ArrayPattern;
+    const leftIdentifiers = leftDeclaration.elements as Array<Identifier>;
```
LavrovArtem, line 68:
```suggestion
    const leftIdentifiers = leftDeclaration.elements as Identifier[];
```

LavrovArtem (changes requested):

LavrovArtem (approved):

miherlosev (approved):

PR #2635, opened 2021-05-18
fix: don't raise the blur after focus on a not focusable element

wentwrong:
## Purpose
We shouldn't raise the blur event on an active element after the focus method was called on a not focusable element.

## Approach
- add check that if an element is not focusable, then the consecutive blur event is not needed when the `focus` method was called on this element
- add more strict types

## References
https://github.com/DevExpress/testcafe/pull/6236

## Pre-Merge TODO

### DIFF
--- a/test/client/fixtures/sandbox/event/focus-blur-change-test.js
+++ b/test/client/fixtures/sandbox/event/focus-blur-change-test.js
@@ -1470,8 +1470,8 @@ asyncTest('an active input should not be blurred after a call of the focus on a 
     document.body.appendChild(activeInput);
     document.body.appendChild(notFocusableDiv);
 
-    var isActiveInputFocused     = false;
-    var isActiveInputBlurred     = false;
+    var isActiveInputFocused = false;
+    var isActiveInputBlurred = false;
 
     activeInput.onfocus = function () {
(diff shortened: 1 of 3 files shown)


test/client/fixtures/sandbox/event/focus-blur-change-test.js
```diff
@@ -1461,3 +1461,33 @@ asyncTest('Should not change active element after source element was focused on
...
+    var activeInput     = document.createElement('input');
+    var notFocusableDiv = document.createElement('div');
+
+    activeInput.className = notFocusableDiv.className = TEST_ELEMENT_CLASS;
+
+    document.body.appendChild(activeInput);
+    document.body.appendChild(notFocusableDiv);
+
+    var isActiveInputFocused     = false;
+    var isActiveInputBlurred     = false;
```
wentwrong, line 1474:
```suggestion
    var isActiveInputFocused = false;
    var isActiveInputBlurred = false;
```

wentwrong:
Functional tests are green in related TestCafe PR: https://github.com/DevExpress/testcafe/pull/6236.

LavrovArtem (approved):

PR #2637, opened 2021-05-20
feat: added handling non-HTML pages [closes DevExpress/testcafe#1471]

Aleksey28:
[closes DevExpress/testcafe#1471]
## Purpose
Test hangs with non-HTML tested pages (pdf, plain text, etc.)

## Approach
1. Plane text page converts to HTML text page
2. Other non-HTML pages are downloaded

## References
https://github.com/DevExpress/testcafe/issues/1471

## Pre-Merge TODO

### DIFF
--- a/src/request-pipeline/context.ts
+++ b/src/request-pipeline/context.ts
@@ -65,6 +65,8 @@ interface ContentInfo {
     isFileDownload: boolean;
     isNotModified: boolean;
     isRedirect: boolean;
+    isAttachment: boolean;
+    isTextPage: boolean;
 }
 
 interface FlattenParsedProxyUrl {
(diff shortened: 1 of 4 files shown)


src/request-pipeline/context.ts
```diff
@@ -65,6 +65,8 @@ interface ContentInfo {
     isFileDownload: boolean;
     isNotModified: boolean;
     isRedirect: boolean;
+    isAttachment: boolean;
+    isTextPage: boolean;
```
miherlosev, line 69: Let's rename it as `isPlainTextPage` here and below.

miherlosev (changes requested):

miherlosev (approved):

PR #2640, opened 2021-06-01
ci: updated the minimal Node.js version to 12

Farfurix:


Farfurix:
There is no need to wait wait the results or run GitHub Actions workflows on this branch.

PR #2643, opened 2021-06-01
fixes #2642: instrument sendBeacon via Navigator.prototype

### LINKED ISSUE #2642: original sendBeacon still available on Navigator.prototype
### What is your Scenario?
```js
Navigator.prototype.sendBeacon.call(navigator, url, data);
```
 
The HTTP request is not intercepted and request mocking fails.

The HTTP request should be intercepted and handled by the request mocking.

michaelficarra:
## Purpose
Fixes #2642. Disallow subversion of instrumentation via `Navigator.prototype.sendBeacon.call`.

## Approach
Instrument `sendBeacon` via `Navigator.prototype` instead of the `navigator` singleton instance. I've noticed that there's other instrumentation with similar failures, but I kept this PR minimal, just addressing #2642.

## References
#2642

## Pre-Merge TODO
- [X] Make sure that existing tests do not fail

### DIFF
--- a/src/client/sandbox/native-methods.ts
+++ b/src/client/sandbox/native-methods.ts
@@ -976,7 +976,7 @@ class NativeMethods {
         this.clearInterval       = win.clearInterval || winProto.clearInterval;
 
         this.registerProtocolHandler = win.navigator.registerProtocolHandler;
-        this.sendBeacon              = win.navigator.sendBeacon;
+        this.sendBeacon              = win.Navigator.prototype.sendBeacon;
 
         if (win.XMLHttpRequest) {
             // NOTE: IE11 has no EventTarget so we should save "Event" methods separately
--- a/src/client/sandbox/native-methods.ts
+++ b/src/client/sandbox/native-methods.ts
@@ -976,7 +976,7 @@ class NativeMethods {
         this.clearInterval       = win.clearInterval || winProto.clearInterval;
 
         this.registerProtocolHandler = win.navigator.registerProtocolHandler;
-        this.sendBeacon              = win.Navigator.prototype.sendBeacon;
+        this.sendBeacon              = win.Navigator && win.Navigator.prototype.sendBeacon;
 
         if (win.XMLHttpRequest) {
             // NOTE: IE11 has no EventTarget so we should save "Event" methods separately
--- a/src/client/sandbox/node/window.ts
+++ b/src/client/sandbox/node/window.ts
@@ -785,7 +785,7 @@ export default class WindowSandbox extends SandboxBase {
         }
 
         if (window.navigator.sendBeacon) {
-            overrideFunction(window.navigator, 'sendBeacon', function (this: Navigator) {
+            overrideFunction(window.Navigator.prototype, 'sendBeacon', function (this: Navigator) {
                 if (typeof arguments[0] === 'string')
                     arguments[0] = getProxyUrl(arguments[0]);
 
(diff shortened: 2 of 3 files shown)


miherlosev:
Hi @michaelficarra

Thank you for this good work. The tests are red because there are some problems with the `QUnit` test running in `Saucelabs` cloud browsers. We will inform you as soon as this is fixed.

Farfurix (changes requested):
Hello,

The added test failed here: https://app.saucelabs.com/tests/589a13f0b5d448a686fba69c08d224cb#15

Based on your approach, I believe you need to fix the stored native method in src/client/sandbox/native-methods.ts:
```diff
- this.sendBeacon              = win.navigator.sendBeacon;
+ this.sendBeacon              = win.Navigator.prototype.sendBeacon;
```

michaelficarra:
@Farfurix Thanks for the feedback. I think I just messed up one of the tests, actually. See the latest commit.

Farfurix:
@michaelficarra 

You still need to fix the src/client/sandbox/native-methods.ts file. I mentioned it in [my review comment](https://github.com/DevExpress/testcafe-hammerhead/pull/2643#pullrequestreview-675030986).

michaelficarra:
@Farfurix I didn't make that change because I don't believe that change is necessary. The tests passed in CI on all browsers except Microsoft Edge, which timed out. Are the tests flaky? Can they be re-run?

bakkot:
@Farfurix `win.navigator.sendBeacon` and `win.Navigator.prototype.sendBeacon` are the same value; it should be fine to refer to it by either name.

Farfurix:
@michaelficarra @bakkot 

Hello,

I agree that the result will be correct for both `sendBeacon` wrappers on the client page (`window.Navigator.prototype.sendBeacon = wrapper` or `window.navigator.sendBeacon = wrapper`) because they are equal:
```js
window.Navigator.prototype.sendBeacon === window.navigator.sendBeacon
```
I just want to point out that for consistency we should use the saved interface (`Navigator`) method to override it (`window.Navigator.prototype.sendBeacon`).

michaelficarra:
@Farfurix Okay, done.

Farfurix (changes requested):
There are some differences related to Navigator in Workers. That's why client tests failed.
We need to rewrite conditions in the following way: in `native-methods.ts`, we should check if `win.Navigator` exists, in other places we should only check our `nativeMethods.sendBeacon` saved method. Also, please rebase your branch to fix Microsoft Edge tests.

michaelficarra:
@Farfurix Done.

Farfurix (changes requested):
One more thing to fix in tests (test/client/fixtures/sandbox/node/window-test.js):
```diff
-     if (window.navigator.sendBeacon)
+     if (nativeMethods.sendBeacon)
-         window.checkStringRepresentation(window.navigator.sendBeacon, nativeMethods.sendBeacon);
+         window.checkStringRepresentation(window.Navigator.prototype.sendBeacon, nativeMethods.sendBeacon, 'Navigator.prototype.sendBeacon');
```

michaelficarra:
@Farfurix Not sure why, but done.

Farfurix:
@michaelficarra 

These changes are needed due to your approach: we need to check the overridden prototype method in the "wrappers of native functions should return the correct string representations" test. Now, everything looks good to me, and I requested a second review to merge your PR. Thank you for your contribution!

Farfurix (approved):

PR #2648, opened 2021-06-08
do not open new window if preventDefault was called in later handler (tc #5621)

AlexKamaev:
In the https://github.com/DevExpress/testcafe-hammerhead/pull/2582 I handle the case when a link element has click handler with the `preventDefault` call.

Now, I process the case when the `preventDefault` is called in handler which is defined on some parent element (element, body, window) and event is bubbling.
Example:
```HTML
<document>
    <div onclick="e.preventDefault()">
        <a href="example.com" target="_blank">click me</a>
    </div>
</document>
```
and https://github.com/DevExpress/testcafe/issues/5621#issuecomment-784902631

Approach: add the latest window.click handler, where I check if the event was prevented on previous element.
There is a specific case if event propagation was stopped by `stopPropgation/stopImmediatePropagation` methods. In this specific case I added timeout.

### DIFF
--- a/src/client/sandbox/child-window/index.ts
+++ b/src/client/sandbox/child-window/index.ts
@@ -14,6 +14,7 @@ import Listeners from '../event/listeners';
 import INTERNAL_PROPS from '../../../processing/dom/internal-properties';
 import getTopOpenerWindow from '../../utils/get-top-opener-window';
 import { isIframeWindow } from '../../utils/dom';
+import nextTick from '../../utils/next-tick';
 
 const DEFAULT_WINDOW_PARAMETERS = 'width=500px, height=500px';
 const STORE_CHILD_WINDOW_CMD    = 'hammerhead|command|store-child-window';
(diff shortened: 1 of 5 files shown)


src/client/sandbox/child-window/index.ts
```diff
@@ -88,10 +88,44 @@ export default class ChildWindowSandbox extends SandboxBase {
 
             e.preventDefault();
 
-            this._openUrlInNewWindow(url);
+            this._openNewWindowAfterAllEventHandlers(url, e);
         });
     }
 
+    private _openNewWindowAfterAllEventHandlers (url, e) {
+        const nativeAddEventListener    = nativeMethods.windowAddEventListener || nativeMethods.addEventListener;
```
miherlosev, line 96: Add the `NativeMethods.getAddEventListener`, `NativeMethods.getRemoveEventListener` methods and use them here and replace the 
`nativeMethods.windowAddEventListener || nativeMethods.addEventListener`, 
`nativeMethods.windowRemoveEventListener || nativeMethods.removeEventListener`
expressions across the whole module code.

src/client/sandbox/child-window/index.ts
```diff
@@ -88,10 +88,44 @@ export default class ChildWindowSandbox extends SandboxBase {
...
+        // `stopPropagation` or `stopImmediatePropagation` methods
+        const origPreventDefault = e.preventDefault;
+
+        e.preventDefault = () => {
+            isDefaultPrevented = true;
+
+            return origPreventDefault.call(e);
+        };
+
+        setTimeout(() => {
```
miherlosev, line 123: Add the `delay(fn, timeout)` utility function.

src/client/sandbox/child-window/index.ts
```diff
@@ -88,10 +88,44 @@ export default class ChildWindowSandbox extends SandboxBase {
 
             e.preventDefault();
 
-            this._openUrlInNewWindow(url);
+            this._openNewWindowAfterAllEventHandlers(url, e);
```
miherlosev, line 91: I propose to rename root the function as `openUrlInNewWindowIfNotPrevented`.
The internal function renames as `openUrlInNewWindowIfNotPreventedHandler`.

miherlosev (changes requested):
Only naming issues

LavrovArtem (approved):

src/client/sandbox/child-window/index.ts
AlexKamaev: We already have the `nextTick` method here:
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/utils/next-tick.ts

src/client/sandbox/child-window/index.ts
AlexKamaev: we already have nice methods for this:
https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/client/sandbox/event/listeners.ts#L43

src/client/sandbox/child-window/index.ts
AlexKamaev: ok

miherlosev (approved):

PR #2649, opened 2021-06-09
fix `WorkerGlobalScope.importScripts should be overwritten` (close #2521)

### LINKED ISSUE #2521: `WorkerGlobalScope.importScripts` should be overwritten
The [WorkerGlobalScope.importScripts()](https://html.spec.whatwg.org/multipage/workers.html#importing-scripts-and-libraries) method is being used to import scripts into worker global scope. We should handle URLs passed to it, because otherwise imported resources may not be processed by HH.  

Example in which the imported script should be processed, but it doesn't:
```js
require('http')
    .createServer((req, res) => {
        if (req.url === '/') {
            res.writeHead(200, { 'content-type': 'text/html' });
            res.end(`
                <!DOCTYPE html>
                <head>
                    <link rel="shortcut icon" href="#" />
                </head>
                <body>
                    <script>
                        navigator.serviceWorker.register('sw.js');
                    </script>
                </body>
            `);
        } else if (req.url === '/sw.js') {
            res.writeHead(200, { 'content-type': 'application/javascript' });
            res.end(`
                self.addEventListener('fetch', function(event) {
                    console.log('fetch event in sw.js with url', event.request.url);
                });

                importScripts('https://wentwrong.github.io/sw-fetch/one-more-service-worker.js');
            `);
        } else
            res.destroy();
    })
    .listen(2020, () => console.log('http://localhost:2020'));
```
This bug was found during analysis of the Health-Monitor run issues on the twitter.com website.

LavrovArtem:


src/client/utils/url.ts
```diff
@@ -102,7 +102,7 @@ export let getProxyUrl = function (url: string, opts?): string {
         return url;
 
     charset = charset || (parsedResourceType.isScript || parsedResourceType.isServiceWorker) &&
-        document[INTERNAL_PROPS.documentCharset];
+        self.document && document[INTERNAL_PROPS.documentCharset];
```
miherlosev, line 105: This condition become large. Let's move it to the `getCharsetFromDocument (self, parsedResourceType)` function.

miherlosev (changes requested):

miherlosev (approved):

Farfurix (approved):

PR #2660, opened 2021-07-01
fix `Rewritten document of an iframe with the srcdoc attribute remains unproxied` (close #2647)

### LINKED ISSUE #2647: Rewritten document of an iframe with the srcdoc attribute remains unproxied
After an iframe with the `srcdoc` attribute created from script and its content dynamically written (using the document's write method), the document remains unproxied and it leads to many errors like the following: 
> SyntaxError: Unexpected token = in JSON at position

It happens when the unproxied document tries to send a post message to the parent document.

Can be reproduced with the following example:
```html
<!doctype html>
<html lang="en">
  <head>
    <!-- Required meta tags -->
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
    <title>Hello, world!</title>
  </head>
  <body>
    <script>
        iframe.srcdoc = '<h1>test</h1>';

        // iframe.contentWindow.addEventListener('load', function (e) {
        //  console.trace('load event', e);
        // });

        iframe.contentDocument.open();
        iframe.contentDocument.write('<script>parent.postMessage("myevent", "*")<\/script>');
        iframe.contentDocument.close();
    </script>
  </body>
</html>
```
Also can be reproduced on [this](https://googleads.github.io/google-publisher-tag-samples/basic/display-test-ad/demo.html) example which displays a test google ad.

Based on analysis of errors which arise on the w3school.com website (separated from https://github.com/DevExpress/testcafe-hammerhead/issues/1921).

LavrovArtem:


LavrovArtem:
run monitor

miherlosev (approved):

Farfurix (approved):

PR #2673, opened 2021-07-30
fix content of iframe was not loaded for client case (closes #2675)

### LINKED ISSUE #2675: Cannot switch to an iframe if it was loaded on client and used as form target
see example here: https://github.com/DevExpress/testcafe-hammerhead/issues/1921#issuecomment-461455661

AlexKamaev:


AlexKamaev:
FPR

LavrovArtem (approved):

PR #2688, opened 2021-08-31
feat: added returning an imitation of the `StyleSheets` after filter instead of an array

Aleksey28:
[closes DevExpress/testcafe#6439]

## Purpose
Filtered `StyleSheets` should behave like original/

## Approach
1. Add test 'Getting CSSStyleSheet by property id'
2. Add creating fake styleSheets after filter

## References
https://github.com/DevExpress/testcafe/issues/6439

## Pre-Merge TODO


AndreyBelym (approved):

src/client/sandbox/node/document/index.ts
```diff
@@ -107,6 +107,22 @@ export default class DocumentSandbox extends SandboxBase {
             nativeMethods.objectDefineProperty(owner, prop, overriddenDescriptor);
     }
 
+    private static _createFakeStyleSheets (styleSheets) {
+        const fakeStyleSheets = nativeMethods.objectCreate(StyleSheetList, {
+            item:   nativeMethods.objectGetOwnPropertyDescriptor(styleSheets, 'item'),
```
LavrovArtem, line 112:
```js
Object.getOwnPropertyDescriptor(document.styleSheets, 'item') // undefined
Object.getOwnPropertyDescriptor(document.styleSheets, 'length') // undefined
```

The right way:
```js
Object.getOwnPropertyDescriptor(StyleSheetList.prototype, 'item') // {writable: true, enumerable: true, configurable: true, value: ƒ}
```

LavrovArtem (changes requested):

src/client/sandbox/node/document/index.ts
Aleksey28: I had another idea, namely. I added a new condition if the input parameter is an instance of StyleSheetList we will return one as is, otherwise, the input parameter will be a simple array with those methods after sorting and we should get them right from the parameter.

LavrovArtem (approved):

PR #2720, opened 2021-12-22
the stored returnValue of an unload event converts to string

LavrovArtem:


src/client/sandbox/event/unload.ts
```diff
@@ -89,12 +89,21 @@ export default class UnloadSandbox extends SandboxBase {
             }));
 
             // NOTE: need to pass `this` scope for https://github.com/DevExpress/testcafe/issues/6563
-            const res = originListener.call(this, e)
+            const res = originListener.call(this, e);
 
             if (res !== void 0) {
                 eventProperties.storedReturnValue = res;
```
miherlosev, line 95: I propose to refactor in a such way:
```js
eventProperties.storedReturnValue = this._prepareStoredReturnValue(res);
```

src/client/sandbox/event/unload.ts
```diff
@@ -89,12 +89,21 @@ export default class UnloadSandbox extends SandboxBase {
...
                 eventProperties.prevented         = true;
             }
+
+            if (typeof eventProperties.storedReturnValue !== 'string') {
+                try {
+                    eventProperties.storedReturnValue = '' + eventProperties.storedReturnValue;
```
miherlosev, line 101: It's implicit casting to string.
Can we use the explicit type casting here?
```js
eventProperties.storedReturnValue = String(eventProperties.storedReturnValue);
```

miherlosev (changes requested):

LavrovArtem:
https://github.com/DevExpress/testcafe/issues/5796

miherlosev (approved):

PR #2737, opened 2022-02-08
fix: xhr request error with blob urls (closes #2634)

### LINKED ISSUE #2634: Error when requesting blob object urls through XHR
### What is your Scenario?
Using XHR to request object urls of type  `blob:` will make hammerhead error out and thus, the request is not made.

I tracked down the exception to this line which expects `reqOpts` to have a value in the `send` wrapper, but no values where defined in the `open` method wrapper because it is a local object url.

Here is where the request options are supposed to be set (`xhr.open`). This only happens for `http` and `file` domains, not for `blob` which I think makes sense.

https://github.com/DevExpress/testcafe-hammerhead/blob/a878aefb01e45c3dc58d9d92da6e618cbc494c4c/src/client/sandbox/xhr.ts#L149

And here in `xhr.send` it tries to retrieve the request options which never fails for regular requests to `http` urls, but since those never are set for `blob:` urls, it fails badly.
https://github.com/DevExpress/testcafe-hammerhead/blob/a878aefb01e45c3dc58d9d92da6e618cbc494c4c/src/client/sandbox/xhr.ts#L162
 
Considering the following snippet:

```js
const blob = new Blob(['this is a text'], { type: 'plain/text' });
const url = URL.createObjectURL(blob);

const xhr = new XMLHttpRequest();

xhr.open('get', url, false);
xhr.addEventListener('load', () => {
    console.log(xhr.responseText);
});

xhr.send(); // <-- this throws an exception: "a is undefined"
```
 
Because of the exception, the request is not made.

`xhr.send();` correctly sends the request and the event listener prints `this is a text`.
 
### What is your public web site URL?

Your website URL (or attach your complete example): 

<details>
<summary>Your complete app code (or attach your test files):</summary>

 
```js
 
```
</details>

<details>
<summary>Screenshots:</summary>

```

```
</details>

### Steps to Reproduce:

1. Go to: ...
2. Execute this command: ...
3. See the error: ...
 
### Your Environment details:

* node.js version:                     
* browser name and version: 
* platform and version:          
* other:

kirillsalikhov:
## Purpose
Fixes error with xhr request when pass blob as url

## Approach
Adding check for reqOpts variable existence in send method override

```
// src/client/sandbox/xhr.ts

        overrideFunction(xmlHttpRequestProto, 'send', function (this: XMLHttpRequest, ...args: Parameters<XMLHttpRequest['send']>) {
            //...
            const reqOpts = XhrSandbox.REQUESTS_OPTIONS.get(this);
            
            if (reqOpts && reqOpts.withCredentials !== this.withCredentials)
                XhrSandbox._reopenXhr(this, reqOpts);
            //...
        });
```

## References
https://github.com/DevExpress/testcafe-hammerhead/issues/2634

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/xhr.ts
+++ b/src/client/sandbox/xhr.ts
@@ -159,7 +159,7 @@ export default class XhrSandbox extends SandboxBaseWithDelayedSettings {
 
             const reqOpts = XhrSandbox.REQUESTS_OPTIONS.get(this);
 
-            if (reqOpts.withCredentials !== this.withCredentials)
+            if (reqOpts && reqOpts.withCredentials !== this.withCredentials)
                 XhrSandbox._reopenXhr(this, reqOpts);
 
             xhrSandbox.emit(xhrSandbox.BEFORE_XHR_SEND_EVENT, { xhr: this });
--- a/src/client/sandbox/xhr.ts
+++ b/src/client/sandbox/xhr.ts
@@ -32,7 +32,15 @@ export default class XhrSandbox extends SandboxBaseWithDelayedSettings {
         super(waitHammerheadSettings);
     }
 
-    static createNativeXHR () {
+    static setRequestOptions (req: XMLHttpRequest, withCredentials: boolean, args: Parameters<XMLHttpRequest['open']>): void {
+        XhrSandbox.REQUESTS_OPTIONS.set(req, {
+            withCredentials,
+            openArgs: args,
+            headers:  []
+        });
+    }
+
+    static createNativeXHR (): XMLHttpRequest {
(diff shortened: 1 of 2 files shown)


viktoria2506:
Thank you for your contribution to TestCafe. We will review this PR.

miherlosev (changes requested):
Hi @kirillsalikhov 

It's better to add the executed XHR request to REQUESTS_OPTIONS [here](https://github.com/DevExpress/testcafe-hammerhead/blob/fd3fb159ca7e3d0431e5297141c7fbcd3b04ad79/src/client/sandbox/xhr.ts#L136).

miherlosev (approved):

PR #2741, opened 2022-02-14
feat: added synchronization the tough-cookie with the client

Aleksey28:
## Purpose
Add synchronization the tough-cookie with the client after changing them with methods `setCookies` and `deleteCookie`

## Approach
1. Add `syncCookies` property to the Cookies
2. Add filling `syncCookies` in the methods `setCookies` and `deleteCookie`
3. Set new cookies on the service message
4. Add unit tests

## Pre-Merge TODO

### DIFF
--- a/src/session/cookies.ts
+++ b/src/session/cookies.ts
@@ -32,8 +32,8 @@ export default class Cookies {
     private readonly _getAllCookiesSync: any;
     private readonly _putCookieSync: any;
     private readonly _removeCookieSync: any;
-    private readonly _removeCookiesSync: any;
     private readonly _removeAllCookiesSync: any;
+    private readonly _syncCookie: Cookie[];
 
     constructor () {
         this._cookieJar            = new CookieJar();
@@ -42,8 +42,8 @@ export default class Cookies {
         this._getAllCookiesSync    = this._syncWrap('getAllCookies');
         this._putCookieSync        = this._syncWrap('putCookie');
         this._removeCookieSync     = this._syncWrap('removeCookie');
-        this._removeCookiesSync    = this._syncWrap('removeCookies');
         this._removeAllCookiesSync = this._syncWrap('removeAllCookies');
+        this._syncCookie           = [];
     }
 
     _syncWrap(method: string) {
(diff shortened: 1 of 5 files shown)


src/session/index.ts
```diff
@@ -166,6 +167,33 @@ export default abstract class Session extends EventEmitter {
         throw new Error('Malformed service message or message handler is not implemented');
     }
 
+    get syncCookies () {
+        const parsedSyncCookies = [] as string[];
+
+        while (true) {
```
LavrovArtem, line 173: Infinite loop is wrong solution.

Just use `this.cookies.syncCookies.map()` to generate sync cookies.

LavrovArtem (changes requested):

src/session/cookies.ts
```diff
@@ -42,8 +42,8 @@ export default class Cookies {
         this._getAllCookiesSync    = this._syncWrap('getAllCookies');
         this._putCookieSync        = this._syncWrap('putCookie');
         this._removeCookieSync     = this._syncWrap('removeCookie');
-        this._removeCookiesSync    = this._syncWrap('removeCookies');
         this._removeAllCookiesSync = this._syncWrap('removeAllCookies');
+        this.syncCookies           = [];
```
LavrovArtem, line 46: I think we need to rename this property to `pendingSyncCookies` and make it private.
Also, we need takePendingSyncCookies method, something like this:
```js
takePendingSyncCookies () {
    const cookies = this.pendingSyncCookies;
    
    this.pendingSyncCookies = [];

    return cookies;
}
```

src/session/cookies.ts
```diff
@@ -157,31 +157,27 @@ export default class Cookies {
         return cookies.filter(cookie => filterKeys.every(key => cookie[key] === filters[key]));
     }
 
-    private _getCookiesByApi (cookie: Cookie.Properties, urls?: Url[]): Cookie[] {
+    private _getCookiesByApi (cookie: Cookie.Properties, urls?: Url[], strict = false): Cookie[] {
         const { key, domain, path, ...filters } = cookie;
 
         const currentUrls = domain && path ? castArray({ domain, path }) : urls;
```
LavrovArtem, line 163: castArray is unnecessary
just use `[{ domain, path }]`

LavrovArtem (changes requested):

miherlosev (approved):

LavrovArtem (approved):

PR #2758, opened 2022-04-11
feat: Api testing

Aleksey28:
## Purpose

## Approach

## References

## Pre-Merge TODO


src/session/index.ts
```diff
@@ -242,6 +243,19 @@ export default abstract class Session extends EventEmitter {
         return taskScript;
     }
 
+    getProxyUrl (url: string): string {
+        if (!this.proxy?.server2Info)
```
miherlosev, line 247: I guess it should be more complicated:
* need to determine the url type. If it's same-domain, the use the `server1Info`, else - `server2Info`.
* before creating proxy url use this [preparation](https://github.com/DevExpress/testcafe-hammerhead/blob/f4364fb67e5c588b8033b7183a318a1615fee17c/src/proxy/index.ts#L238).

The using other options (`externalProxySettings`, `disableHttp2`) we will consider later.

miherlosev (changes requested):

src/typings/session.d.ts
```diff
@@ -43,3 +44,24 @@ export interface RequestEventListenerError {
     methodName: string;
 }
 
+export interface RequestOptionsSettings {
```
miherlosev, line 47: `RequestOptionsInit`

miherlosev (changes requested):

src/request-pipeline/destination-request/index.ts
```diff
@@ -48,12 +49,14 @@ export default class DestinationRequest extends EventEmitter implements Destinat
...
     private readonly timeout: number;
 
-    constructor (readonly opts: RequestOptions, readonly cache: boolean) {
+    constructor (readonly opts: RequestOptions, readonly cache: boolean = false) {
         super();
 
         this.protocolInterface = this.opts.isHttps ? https : http;
 
-        this.timeout = this.opts.isAjax ? opts.requestTimeout.ajax : opts.requestTimeout.page;
+        this.timeout = this.opts.isAjax
```
miherlosev, line 57: Pack these assignments to the function:
```js
this.timeout = this._getTimeout(this.opts);
```

miherlosev (changes requested):

src/session/index.ts
```diff
@@ -136,10 +136,7 @@ export default abstract class Session extends EventEmitter {
     }
 
     private _getOptions (options: Partial<SessionOptions> = {}): SessionOptions {
-        const requestTimeout = {
-            page: options.requestTimeout && options.requestTimeout.page || DEFAULT_REQUEST_TIMEOUT.page,
-            ajax: options.requestTimeout && options.requestTimeout.ajax || DEFAULT_REQUEST_TIMEOUT.ajax,
-        };
+        const requestTimeout = getRequestTimeouts(options.requestTimeout);
```
miherlosev, line 139: It should be finally calculated only in one place - in DestiantionRequest.

miherlosev (changes requested):

test/server/proxy/regression-test.js
```diff
@@ -265,6 +265,8 @@ describe('Regression', () => {
     });
 
     it('Should raise error on request timeout (T224541)', done => {
+        session.options.requestTimeout = {};
```
miherlosev, line 268: Does it still necessary?

miherlosev (changes requested):

miherlosev (changes requested):

test/server/proxy/regression-test.js
Aleksey28: Yes, I deleted default timeouts for the session.

miherlosev (approved):

PR #2765, opened 2022-05-04
fix: fixed getting cookies only by path or domain

Aleksey28:
## Purpose
Getting or deleting cookies separately by path or domain doesn't work.  It should work.

## Approach
Add path and domain to the filters also when one of them wasn't set.

## References
[testcafe-hammerhead-24.5.18.zip](https://github.com/DevExpress/testcafe-hammerhead/files/8618991/testcafe-hammerhead-24.5.18.zip)

## Pre-Merge TODO

### DIFF
--- a/src/session/cookies.ts
+++ b/src/session/cookies.ts
@@ -164,7 +164,6 @@ export default class Cookies {
 
     /*eslint-disable*/
     private _getCookiesByApi (cookie: Cookie.Properties, urls?: Url[], strict = false): Cookie[] {
-        debugger;
         const { key, domain, path, ...filters } = cookie;
 
         const currentUrls = domain && path ? [{ domain, path }] : urls;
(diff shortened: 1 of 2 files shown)


src/session/cookies.ts
```diff
@@ -162,7 +162,9 @@ export default class Cookies {
         return cookies.filter(cookie => filterKeys.every(key => cookie[key] === filters[key]));
     }
 
+    /*eslint-disable*/
     private _getCookiesByApi (cookie: Cookie.Properties, urls?: Url[], strict = false): Cookie[] {
+        debugger;
```
AndreyBelym, line 167: Pls dlt.
```suggestion
```

AndreyBelym (approved):

src/session/cookies.ts
```diff
@@ -162,6 +162,7 @@ export default class Cookies {
         return cookies.filter(cookie => filterKeys.every(key => cookie[key] === filters[key]));
     }
 
+    /*eslint-disable*/
```
miherlosev, line 165: You disabled linting for a part of the file.
Disable only the problematic line.

miherlosev (changes requested):

miherlosev (approved):

PR #2792, opened 2022-08-09
fix: added overriding descriptor in prototype

Aleksey28:
[closes DevExpress/testcafe#7166]

## Purpose
Fix regression

## Approach
1. Analyze the issue
2. Fix 

## References
https://github.com/DevExpress/testcafe/issues/7166

## Pre-Merge TODO

### DIFF
--- a/test/client/data/style-sandbox/index.html
+++ b/test/client/data/style-sandbox/index.html
@@ -4,11 +4,10 @@
     <script src="/hammerhead.js" class="script-hammerhead-shadow-ui"></script>
 </head>
 <body>
-<div id="item" style="background: url('https://testcafe.io/img/qa-guy.svg');width: 500px;height: 500px;"></div>
+<div id="item" style="background: url('https://domain.com/test.svg');width: 500px;height: 500px;"></div>
 </body>
 <script>
-    var hammerhead = window['%hammerhead%'];
-
+    var hammerhead       = window['%hammerhead%'];
     var getPropertyValue = CSSStyleDeclaration.prototype.getPropertyValue;
     var item             = document.querySelector('div');
     var prototype        = Object.getPrototypeOf(item.style);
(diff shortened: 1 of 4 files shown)


miherlosev (changes requested):
Please add a real test case in the tests. At present, there is only overriding an utility function. It's not clear why it's necessary. It's useful to do this with a separate test page and 'createTestIrame' function.

test/client/data/style-sandbox/index.html
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <script src="/hammerhead.js" class="script-hammerhead-shadow-ui"></script>
+</head>
+<body>
+<div id="item" style="background: url('https://testcafe.io/img/qa-guy.svg');width: 500px;height: 500px;"></div>
```
miherlosev, line 7: Don't use links to outer sites. Create a simple image in Paint and use it in tests.

test/client/fixtures/sandbox/style-test.js
```diff
@@ -185,3 +185,15 @@ test('the appendData function does not break if there is no parent element (TC-3
...
     strictEqual(textNode.data, '\n.class:hover{}\n');
 });
+
+test('override properties in the CSSStyleDeclaration.prototype (GH-7166)', function () {
+    return createTestIframe({ src: getSameDomainPageUrl('../../data/style-sandbox/index.html') })
+        .then(function (iframe) {
+            var iframeDocument = iframe.contentDocument;
+            var item           = iframeDocument.querySelector('div');
+
+            ok(item.style.backgroundImage);
```
miherlosev, line 195: It's better to compare with expected value.

miherlosev (changes requested):

miherlosev (approved):

PR #2807, opened 2022-10-28
Storages refactoring for Roles

AlexKamaev:
add proxyless strategy for Storages.
remove storages overriding for proxyless mode


src/client/sandbox/storages/strategies.ts
```diff
@@ -0,0 +1,369 @@
...
+
+    lock (): void {
+        return void 0;
+    }
+
+    unlock (): void {
+        return void 0;
+    }
+
+    restore (): StoragesBackup {
```
miherlosev, line 345: The type for this already exists - https://github.com/DevExpress/testcafe-hammerhead/blob/master/src/typings/session.d.ts#L17.

Also, the empty storages backup should be defined as
```js
{
    localStorage:   '[[],[]]',
    sessionStorage: '[[],[]]',
 }
```

To add syntax sugar,  I propose converting interface `StoragesSnapshot` to class and the static `empty` property.
So, the updated code will look like this:
```js
restore (): StoragesSnapshot{
      return StoragesSnapshot.empty;
}
```

miherlosev (changes requested):

PR #2813, opened 2022-11-10
fix: session referer rewritten in iframe tasks

naggingant:
Closes https://github.com/DevExpress/testcafe/issues/7376

## Purpose
This partially undoes https://github.com/DevExpress/testcafe-hammerhead/commit/55209a53653be23536624654a3b5119437a5450c so that the session referer option does not get rewritten by iframe tasks.

## Approach
I simply added the condition before this rewrite happens, but I am not familiar enough with the codebase nor proxies in general to fully justify the change.
Let me know if it makes sense as well as the test. I can close it otherwise.
## References
https://github.com/DevExpress/testcafe/issues/7376

## Pre-Merge TODO

### DIFF
--- a/src/proxy/index.ts
+++ b/src/proxy/index.ts
@@ -211,7 +211,8 @@ export default class Proxy extends Router {
         const windowId    = refererDest && refererDest.windowId || void 0;
 
         if (session) {
-            session.options.referer = referer || session.options.referer;
+            if (referer && !isIframe)
+                session.options.referer = referer;
 
             res.setHeader(BUILTIN_HEADERS.contentType, 'application/x-javascript');
             addPreventCachingHeaders(res);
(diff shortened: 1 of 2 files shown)


src/proxy/index.ts
```diff
@@ -211,7 +211,7 @@ export default class Proxy extends Router {
         const windowId    = refererDest && refererDest.windowId || void 0;
 
         if (session) {
-            session.options.referer = referer || session.options.referer;
+            if (!isIframe) session.options.referer = referer || session.options.referer;
```
Aleksey28, line 214: I suggest you do this as follows:
```suggestion
            if (!isIframe && referer)
                session.options.referer = referer;
```

Aleksey28 (changes requested):
Thank you for the PR. You will need to refactor it a bit.

src/proxy/index.ts
naggingant: Sounds good, thanks

naggingant:
Applied the suggestion. Thank you for the prompt response!

Aleksey28 (approved):

PR #2818, opened 2022-12-08
fix: fixed cookie synchronization

Aleksey28:
## Purpose
The cookies weren't synchronized correctly. It needs to be fixed.

## Approach
1. Add test
2. Research problem
3. Fix

## References
https://stackoverflow.com/questions/74684772/deleted-cookies-arent-deleted-properly-and-still-sent-with-requests

## Pre-Merge TODO

### DIFF
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -67,7 +67,7 @@ function formatSyncCookieKey (cookie: CookieRecord): string {
     const path         = encodeURIComponent(cookie.path);
     const expires      = cookie.expires !== 'Infinity' ? cookie.expires.getTime().toString(TIME_RADIX) : '';
     const lastAccessed = cookie.lastAccessed.getTime().toString(TIME_RADIX);
-    const maxAge       = cookie.maxAge && cookie.maxAge !== 'Infinity' ? cookie.maxAge.toString(TIME_RADIX) : '';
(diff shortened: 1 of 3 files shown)


src/utils/cookie.ts
```diff
@@ -67,7 +67,7 @@ function formatSyncCookieKey (cookie: CookieRecord): string {
     const path         = encodeURIComponent(cookie.path);
     const expires      = cookie.expires !== 'Infinity' ? cookie.expires.getTime().toString(TIME_RADIX) : '';
     const lastAccessed = cookie.lastAccessed.getTime().toString(TIME_RADIX);
-    const maxAge       = cookie.maxAge && cookie.maxAge !== 'Infinity' ? cookie.maxAge.toString(TIME_RADIX) : '';
+    const maxAge       = !isNaN(Number(cookie.maxAge)) && String(cookie.maxAge) !== 'Infinity' ? cookie.maxAge.toString(TIME_RADIX) : '';
```
saschbro, line 70: Hi @Aleksey28, I was about to provide a PR for the [issue I reported on stackoverflow](https://stackoverflow.com/questions/74684772/deleted-cookies-arent-deleted-properly-and-still-sent-with-requests). But it looks like you are already picked that up. Thanks a lot!!!

AndreyBelym (approved):

Issue #2828, opened 2022-12-23
Only one of multiple cookies deleted with "max-age=0" are synced


saschbro:
### What is your Scenario?

I reported the following issue on StackOverflow which was fixed with `28.2.5` by @Aleksey28:
https://stackoverflow.com/questions/74684772/deleted-cookies-arent-deleted-properly-and-still-sent-with-requests

However, in my use case multiple cookies are deleted in the client via `document.cookie = "<cookie-name>=; path=/; max-age=0;"`, but now only the first cookie is really synced. If I set `max-age=1` instead all cookies are synced.

I think this is because of a "timing" issue in the `syncCookie` function. If I add `&& maxAge > 0` to the "if statement" in `src/client/sandbox/cookie/index.ts::142` all cookies are synced with `max-age=0`.

### What is the Current behavior?

Only one of multiple cookies set via `document.cookie = "cookieX=; path=/; max-age=0;"` (with X = 1, .., n) is synced.

### What is the Expected behavior?

All cookies set via `document.cookie = "cookieX=; path=/; max-age=0;"` (with X = 1, .., n) are synced.

### What is your public website URL? (or attach your complete example)

Please see my example that uses TestCafe `2.2.0-rc.1` which uses Hammerhead `28.2.5`:

[testcafe-delete-cookies.zip](https://github.com/DevExpress/testcafe-hammerhead/files/10293214/testcafe-delete-cookies.zip)

#### deleteCookies.test.js
```
import { Selector } from "testcafe";

fixture`Delete cookies`.page`http://localhost:8080/index.html`;

test(`Cookies to delete are synced`, async (t) => {
  await t.debug();

  const buttonMaxAge0 = Selector("#maxAge0");
  // should sync 20 cookies but only syncs one
  await t.click(buttonMaxAge0);

  const buttonMaxAge1 = Selector("#maxAge1");
  // syncs 20 cookies as expected
  await t.click(buttonMaxAge1);

  await t.wait(10000);
}).disablePageCaching;
```
#### index.html:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <title>Delete Multiple Cookies</title>
  </head>
  <body>
    <button id="maxAge0">Set cookies with max-age 0</button>
    <button id="maxAge1">Set cookies with max-age 1</button>
    <script>
      document.getElementById("maxAge0").addEventListener("click", function () {
        for (let i = 0; i < 20; i++) {
[...]

PR #2830, opened 2022-12-24
fix: fiexed setting `max-age` to the cookies

Aleksey28:
[closes DevExpress/testcafe#7432]
[closes DevExpress/testcafe#2828]

## Purpose

If `max-age` isn't null when it isn't specified, `tough-cookie` won't take into account `expires`. It should be fixed. Also, we shouldn't additionally synchronize client cookies if we are setting them or synchronizing for another reason.

## Approach
1. Research the issue
2. Set `null` value in `max-age` if  it isn't specified
3. Additionally synchronize client cookies only when we get them
4. Add test

## References
https://stackoverflow.com/questions/74684772/deleted-cookies-arent-deleted-properly-and-still-sent-with-requests
https://github.com/DevExpress/testcafe/issues/7432
https://github.com/DevExpress/testcafe-hammerhead/issues/2828

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/cookie/index.ts
+++ b/src/client/sandbox/cookie/index.ts
@@ -18,6 +18,7 @@ import {
 import { CookieRecord } from '../../../typings/cookie';
 import ChildWindowSandbox from '../child-window';
 import getTopOpenerWindow from '../../utils/get-top-opener-window';
+import { isNil } from 'lodash';
 
 const MIN_DATE_VALUE = new nativeMethods.date(0).toUTCString(); // eslint-disable-line new-cap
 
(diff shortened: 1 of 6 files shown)


src/client/sandbox/cookie/index.ts
```diff
@@ -18,6 +18,7 @@ import {
 import { CookieRecord } from '../../../typings/cookie';
 import ChildWindowSandbox from '../child-window';
 import getTopOpenerWindow from '../../utils/get-top-opener-window';
+import { isNil } from 'lodash';
```
miherlosev, line 21: Don't import the `lodash` library on the client. It causes the bundling library to the `hammerhead.js` and a lot issues with native methods in the future.

miherlosev (changes requested):

miherlosev (approved):

PR #2831, opened 2022-12-27
Temporary remove some checks & errors due to bad support of optional chaining

JSteunou:
## Purpose

closes https://github.com/DevExpress/testcafe/issues/7387 & https://github.com/DevExpress/testcafe/issues/7424

## Approach

The goal here is to prevent testcafe to throw blocking errors whereas the code is totally fine. If there was a syntax error it should throw inside the browser, caught by linter, code editor, whatever, but testcafe logging the callstack shouldn't be that much a feature.

If you really want to maintain testcafe wrapping, checking, logging callstack feature then a better check for the 1st issue should be something like

*if object is undefined and there is optional chaining in chain of calls, return undefined, else throw*

And for the 2nd

*if property access on undefined and there is optional chaining in chain of calls, return undefined, else throw*

## References

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/code-instrumentation/methods.ts
+++ b/src/client/sandbox/code-instrumentation/methods.ts
@@ -73,7 +73,7 @@ export default class MethodCallInstrumentation extends SandboxBase {
                     MethodCallInstrumentation._error(`Cannot call method '${methName}' of ${typeUtils.inaccessibleTypeToStr(owner)}`);
 
                 if (typeof owner[methName] !== 'function')
-                    MethodCallInstrumentation._error(`'${methName}' is not a function`);
+                    return void 0
 
                 // OPTIMIZATION: previously we've performed the
                 // `this.methodWrappers.hasOwnProperty(methName)`
--- a/src/client/sandbox/code-instrumentation/properties/index.ts
+++ b/src/client/sandbox/code-instrumentation/properties/index.ts
@@ -89,7 +89,7 @@ export default class PropertyAccessorsInstrumentation extends SandboxBase {
 
     private static _propertyGetter (owner: any, propName: any, optional = false) {
         if (isNullOrUndefined(owner) && !optional)
-            PropertyAccessorsInstrumentation._error(`Cannot read property '${propName}' of ${inaccessibleTypeToStr(owner)}`);
+            return void 0;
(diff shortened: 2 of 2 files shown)


Aleksey28:
Thank you for sharing your solution. I'm afraid it isn't valid. You remove the part where a universal error is thrown in the case when it tries to execute a not-function property. You can't do this. Otherwise, you will also ignore errors without `optional chaining`:

```js
function getValue(key) {
return mapping[key]() ?? key
}
```
However, you'll get an error if you run this code manually without TestCafe. So, this is not an appropriate solution.

To skip this error, you can use the `-e` CLI option, and your test won't be interrupted.

The main issue is that `testcafe-hammerhead` doesn't handle `optional changing` correctly in proxy mode, and your fix doesn't resolve this issue.

Track [https://github.com/DevExpress/testcafe/issues/7387](https://github.com/DevExpress/testcafe/issues/7387) for updates.

JSteunou:
That's what I said. It is not a perfect fix, but a temporary viable solution to support optional chaining. It is a matter of compromise, I'm ready to sacrifice testcafe syntax checker which is useless for our usecases (we have linter, typescript, ...) but gain optional chaining support.

Aleksey28:
Anyway, this solution isn't appropriate. You can sacrifice the testcafe syntax checker, but it doesn't mean that all users can. For your solution like this, you can create your own fork and use it.

JSteunou:
I understand very well. It will at least provide a solution for those in the same case, they would copy this patch.

Did the testcafe team already reconsider this *checker* feature once? It will be interesting to poll the users to know if they are willing to sacrifice optional chaining support for this... Or at least make it optional. Do they even use it or know it exists?

Aleksey28:
Our decision is based on the fact that a few users encountered this issue and wrote about it. It isn't appropriate to break one functionality to fix another. If there is a way to fix one and not break the other, we will fix it. Currently, you can use the `-e` or `--experimental-proxyless` options. We are intensively working on the `proxyless` mode. It fixes many issues with proxying and we want to set it by default. Note that you can use [filters with the `-e` option](https://testcafe.io/documentation/404038/recipes/debugging/skip-javascript-errors).

PR #2832, opened 2022-12-30
fix: added `srcset` proxying

Aleksey28:
[closes DevExpress/testcafe/issues/6926]

## Purpose
Resolve the issue when a new window closed immediately after click on a link with `target=_blank`.

## Approach
1. Research the problem
2. Find our solution

## References
https://github.com/DevExpress/testcafe/issues/6926

## Pre-Merge TODO


miherlosev (changes requested):
I propose to support at once all tags connected with the `srcset` attribute: ``, `<picture>`, and `<source>`.
https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#resolution_switching_same_size_different_resolutions

src/client/utils/url.ts
```diff
@@ -46,6 +46,11 @@ function getCharsetFromDocument (parsedResourceType: ResourceType): string | nul
 }
 
 export let getProxyUrl = function (url: string | URL, opts?, proxyless = false): string {
+    if (opts?.isUrlsSet) {
```
miherlosev, line 49: I propose adding the default value for the function parameter.
```js
export let getProxyUrl = function (url: string | URL, opts ={}, proxyless = false)
```

miherlosev (changes requested):

miherlosev (approved):

PR #2839, opened 2023-01-20
proxyless: add float variant for the 'getBordersWidth' and 'getElementPadding' utility functions

miherlosev:


Aleksey28 (approved):
Excellent!

PR #2854, opened 2023-02-08
fix: ExternalCookies type definitions and request cookies passing

Artem-Babich:
## Purpose
ExternalCookies object is used for filtering cookies. However, its real structure is different form the structure in type definintions
Previously we did not attach cookies to the cross-domain request even if withCredentials property is set to true. We collect such cookies on the testcafe side, so we just need to get these cookies from the source request and pass them then

## Approach
Fix definitions

## References

https://github.com/DevExpress/testcafe/pull/7504

## Pre-Merge TODO
- [ ] Remove fork dependency

### DIFF
--- a/test/functional/run-testcafe-functional-tests.sh
+++ b/test/functional/run-testcafe-functional-tests.sh
@@ -2,7 +2,7 @@
 
 mkdir ../testcafe
 cd ../testcafe
-git clone --branch Babich-fix-request https://github.com/Artem-Babich/testcafe.git .
+git clone https://github.com/DevExpress/testcafe .
 
 # NOTE: testcafe's Gulpfile.js tries to alias the 'travis' task to the value of GULP_TASK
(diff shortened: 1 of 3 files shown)


AlexKamaev (approved):

test/functional/run-testcafe-functional-tests.sh
```diff
@@ -2,11 +2,11 @@
 
 mkdir ../testcafe
 cd ../testcafe
-git clone https://github.com/DevExpress/testcafe .
+git clone --branch Babich-fix-request https://github.com/Artem-Babich/testcafe.git .
```
Aleksey28, line 5: Rollback this change

Aleksey28 (changes requested):

Aleksey28 (approved):

AlexKamaev (approved):

PR #2877, opened 2023-04-10
Proxyless

Aleksey28:
## Purpose

## Approach

## References

## Pre-Merge TODO


src/client/utils/url.ts
```diff
@@ -370,3 +370,7 @@ export function getAjaxProxyUrl (url: string | URL, credentials: sharedUrlUtils.
 
     return getProxyUrl(url, opts);
 }
+
+export function isNativeAutomation (): boolean {
```
miherlosev, line 374: It's not an url utility. It just a syntax sugar for settings. So, place it in the separate file.
Also, I propose using the combined solution to reduce writing the future code: `SandboxBase` still have a `nativeAutomation` property + static classes/methods uses the `isNativeAutomation` utility function.

miherlosev (changes requested):

miherlosev (approved):

PR #2899, opened 2023-05-26
fix: added moving hidden input to the form bottom

Aleksey28:
[closes DevExpress/testcafe#7489]

## Purpose
Fix the issue with getting elements in form by index. 

## Approach
Move the custom hidden input element to the form bottom each time when a new element is added to that form.

## References

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/upload/hidden-info.ts
+++ b/src/client/sandbox/upload/hidden-info.ts
@@ -3,6 +3,9 @@ import { parse as parseJSON, stringify as stringifyJSON } from '../../../utils/j
 import nativeMethods from '../native-methods';
 import ShadowUI from '../shadow-ui';
 
+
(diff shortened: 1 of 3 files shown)


src/client/sandbox/upload/hidden-info.ts
```diff
@@ -97,3 +97,11 @@ export function removeInputInfo (input) {
 
     return false;
 }
+
+export function moveInputToFormBottom (form: HTMLFormElement) {
+    const inputSelector = '[name="' + INTERNAL_ATTRS.uploadInfoHiddenInputName + '"]';
```
miherlosev, line 102: Move this to the constant on the module level.

miherlosev (changes requested):
Excellent

miherlosev (approved):

PR #2900, opened 2023-05-30
fix IE11 failing in TestCafe (closes #7741)

AlexKamaev:
we cannot use `Object.values` in IE11

### DIFF
--- a/src/processing/style.ts
+++ b/src/processing/style.ts
@@ -11,8 +11,6 @@ import { URL_ATTR_TAGS } from './dom/attributes';
 import DomProcessor from './dom';
 import { Dictionary } from '../typings/common';
 
-// NOTE: We should avoid using native object prototype methods,
-// since they can be overriden by the client code.
 const arrayJoin  = Array.prototype.join;
 const objectKeys = Object.keys;
 
--- a/src/processing/style.ts
+++ b/src/processing/style.ts
@@ -11,10 +11,10 @@ import { URL_ATTR_TAGS } from './dom/attributes';
 import DomProcessor from './dom';
 import { Dictionary } from '../typings/common';
 
-function getTagsString (tagsDict: Dictionary<string>[]) {
+function getTagsString (tagsDict: Dictionary<string[]>) {
     const tags: string[] = [];
 
-    for (const key of Object.keys(tagsDict)) {
+    for (const key in tagsDict) {
         for (const tag of tagsDict[key])
             tags.push(tag);
     }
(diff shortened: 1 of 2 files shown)


src/processing/style.ts
```diff
@@ -9,10 +9,22 @@ import INTERNAL_ATTRS from '../processing/dom/internal-attributes';
...
+            tags.push(tag);
+    }
+
+    return tags.join().replace(/,/g, '|');
+}
 
 const SOURCE_MAP_RE                       = /\/\*\s*[#@]\s*sourceMappingURL\s*=[\s\S]*?\*\/|\/\/[\t ]*[#@][\t ]*sourceMappingURL[\t ]*=.*/ig;
 const CSS_URL_PROPERTY_VALUE_RE           = /(url\s*\(\s*(['"]?))([^\s]*?)(\2\s*\))|(@import\s+(['"]))([^\s]*?)(\6)/g;
-const TAGS_STRING                         = Object.values(URL_ATTR_TAGS).join().replace(/,/g, '|');
+const TAGS_STRING                         = getTagsString(URL_ATTR_TAGS as unknown as Dictionary<string>[]);
```
Aleksey28, line 27: Just use[ `values` from lodash](https://lodash.com/docs/4.17.15#values)

Aleksey28 (changes requested):
Great. I checked and it works. Please leave a comment that it is only for IE.

src/processing/style.ts
miherlosev: It's a common code (shared between client and server). We cannot use here any libraries.
Also, need to rewrite the code to prevent crash in case if the `Object.keys` and `Array.join` methods are overridden on the client side.
```js

const objectKey = Object.prototype.keys;
const arrayJoin = Array.prototype.join;

function getTagsString (tagsDict: Dictionary<string>[]) {
    const tags: string[] = [];

    for (const key of objectKey.call(tagsDict)) {
        for (const tag of tagsDict[key])
            tags.push(tag);
    }

    return arrayJoin.call(tags.join).replace(/,/g, '|');
}
```

miherlosev (changes requested):

src/processing/style.ts
miherlosev: > Object.keys is not in prototype

Yes. However, it's necessary to replace the `array.join` method and other `Object.keys` method call in this file.

miherlosev (approved):

Aleksey28 (approved):

PR #2903, opened 2023-06-07
refactor: added websocket transport for NA

Aleksey28:
## Purpose

## Approach

## References

## Pre-Merge TODO


src/proxy/index.ts
```diff
@@ -141,6 +144,7 @@ export default class Proxy extends Router {
             content:     workerHammerheadContent,
         });
 
+        this.GET(SERVICE_ROUTES.messaging, (req: http.IncomingMessage, res: http.ServerResponse, serverInfo: ServerInfo) => this._onServiceMessage(req, res, serverInfo));
```
miherlosev, line 147: The `GET` route for the service messages is necessary only for upgrade WebSocket requests. Let's create a separate handler for this.

src/proxy/index.ts
```diff
@@ -266,6 +301,28 @@ export default class Proxy extends Router {
         this.server1.listen(port1);
         this.server2.listen(port2);
 
+        this.wss1 = new nodeWebSocket.Server({ noServer: true });
+        this.wss2 = new nodeWebSocket.Server({ noServer: true });
+
+        this.wss1.on('connection', (ws) => {
+            ws.on('error', console.error);
```
miherlosev, line 308: Log to the debug logger.

miherlosev (changes requested):

miherlosev (approved):

PR #2913, opened 2023-06-20
Fix reprocessElementsAssociatedWithIframe throwing an DOMException error

mariyanbitsov-af:
## Purpose
_This pr fixes a bug with the processing of elements associated with Iframe. When the iframe name contains special characters, the querySelector method throws an exception and removes all the elements on the page._

## Approach
_The fix try catches the querySelectorAll method, in order to prevent exceptions that will remove all elements on the page. This way the expected behaviour is preserved_

## References
_https://github.com/DevExpress/testcafe/issues/7807_

## Pre-Merge TODO

### DIFF
--- a/src/client/sandbox/node/element.ts
+++ b/src/client/sandbox/node/element.ts
@@ -1,37 +1,37 @@
-import INTERNAL_PROPS from '../../../processing/dom/internal-properties';
-import SandboxBase from '../base';
-import NodeSandbox from '../node/index';
 import DomProcessor from '../../../processing/dom';
-import nativeMethods from '../native-methods';
-import domProcessor from '../../dom-processor';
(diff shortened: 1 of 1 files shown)


AlexKamaev:
Hello. Thank you for your contribution.
We believe that wrapping the problematic code in a try/catch block is not the best option in this case. We would like to fix the cause of the issue. First of all, please try to escape special characters in a css string. You can use the following function: [https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static](https://developer.mozilla.org/en-US/docs/Web/API/CSS/escape_static)
Also, please do not add unnecessary changes like `imports`.
In addition, please add a test for the fixed behavior.

AlexKamaev:
I closed this PR since there has been no activity for a long time

PR #2946, opened 2023-08-08
fix: add id to websocket transport messages, remove injectables for NA

Artem-Babich:
## Purpose
Support NA in TestCafe Studio

### Injectables

Do not fill injactables for NA mode. We inject all scripts in the NA Resource Injector. 
By removing these injectables we can not inject our custom scripts in NA mode on TestRun level

### Transport

Add id field to messages in order to resolve the promises in the correct order. It is required for the TestCafe Studio. It can send multiple messages at the same time and do not await their answer

## References
https://github.com/DevExpress/testcafe/pull/7878

## Pre-Merge TODO

### DIFF
--- a/src/client/transport/transport-in-socket.ts
+++ b/src/client/transport/transport-in-socket.ts
@@ -1,12 +1,14 @@
 import TransportBase from './transport-base';
-import { ServiceMessage } from '../../typings/proxy';
+import { ServiceMessage, WebSocketServiceMessage } from '../../typings/proxy';
 import Promise from 'pinkie';
 import settings from '../settings';
 import { parse, stringify } from '../../utils/json';
 import nativeMethods from '../sandbox/native-methods';
+import IntegerIdGenerator from '../utils/integer-id-generator';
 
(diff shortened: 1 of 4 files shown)


src/client/transport/transport-in-socket.ts
```diff
@@ -18,17 +20,23 @@ export default class TransportInSocket extends TransportBase {
...
+
         return new Promise((resolve, reject) => {
             const handleMessage = (event) => {
-                cleanListeners();
                 this._activeServiceMsgCount--;
 
-                resolve(parse(event.data));
+                const data = parse(event.data);
+
+                if (data.id === id) {
```
Aleksey28, line 36: Decrement `this.activeServiceMsgCount` also should be under condition. And in that way will be better to use opposite condition block and just use `return` in it.

Aleksey28 (changes requested):
Please, look at my comments.

miherlosev (approved):

Aleksey28 (approved):

PR #2966, opened 2023-10-04
added optional chaining in callMethods check

PavelMor25:
## Approach
Added optional chaining in function's arguments "isFunction" which fix check.

## References
[issue 7758](https://github.com/DevExpress/testcafe/issues/7758)

### DIFF
--- a/src/client/sandbox/code-instrumentation/methods.ts
+++ b/src/client/sandbox/code-instrumentation/methods.ts
@@ -74,10 +74,14 @@ export default class MethodCallInstrumentation extends SandboxBase {
         // So, we need to define code instrumentation functions as 'configurable' so that they can be redefined.
         nativeMethods.objectDefineProperty(window, INSTRUCTION.callMethod, {
             value: (owner: any, methName: any, args: any[], optional = false) => {
-                if (isNullOrUndefined(owner) && !optional)
-                    MethodCallInstrumentation._error(`Cannot call method '${methName}' of ${inaccessibleTypeToStr(owner)}`);
+                if (isNullOrUndefined(owner)) {
+                    if (optional)
+                        return void 0;
 
-                if (!isFunction(owner?.[methName]) && !optional)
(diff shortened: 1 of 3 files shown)


src/client/sandbox/code-instrumentation/methods.ts
```diff
@@ -77,21 +77,21 @@ export default class MethodCallInstrumentation extends SandboxBase {
                 if (isNullOrUndefined(owner) && !optional)
```
AlexKamaev, line 77: I believe that we need to add null reference check here. Smth like follows:
```JS
if (isNullOrUndefined(owner)) {
    if (optional)
        return void 0;

    MethodCallInstrumentation._error(`Cannot call method '${methName}' of ${inaccessibleTypeToStr(owner)}`);
}
```

AlexKamaev: @Aleksey28 please check if proposed solution is appropriate

AlexKamaev (changes requested):

src/client/sandbox/code-instrumentation/methods.ts
Aleksey28: Yes, that will be better 👍

test/client/fixtures/sandbox/code-instrumentation/process-script-test.js
```diff
@@ -269,6 +269,14 @@ test('optional chaining', function () {
...
             expected: void 0,
         },
+        {
+            src:      'var obj = null; window.optionChainingResult = obj?.href?.["name"]?.();',
+            expected: void 0,
+        },
+        {
+            src:      'var obj = { href: "123" }; window.optionChainingResult = obj?.link?.["name"]?.();',
+            expected: void 0,
+        },
```
AlexKamaev, line 279: this tests do not work as expected =(
This code does not generates a script with a `__call$` expression. Please research these tests or write new ones to test `__call$` function implementation.

Aleksey28 (approved):

PR #2968, opened 2023-10-18
fix: support non-canonicalised srcset attribute name #2958

### LINKED ISSUE #2958: srcset Proxying doesn't work with React
### What is your Scenario?

I'm trying to load responsive images with srcsets in an integration test for a React app.

Images cannot be loaded, because the proxying does not work correctly. The proxied URL is malformed.

Images with srcsets can be loaded and rendered correctly.

### What is your public website URL? (or attach your complete example)

Here's a minimal Sandbox where the behavior can be observed: https://codesandbox.io/s/srcset-proxying-test-58w6s7

The first image is rendered with a simple React app, that just renders the image with a srcset. The second image is just rendered in static HTML but has the same properties as the first image. In the preview, the image is visible twice, as expected: 
Screenshot 2023-08-25 at 15 21 55

When running the provided Testcafe script and test, the image rendered through React fails to load, and only the statically rendered image shows up: 
Screenshot 2023-08-25 at 15 22 21
(The image itself is different in the two screenshots because I'm using an sample image API.)

### What is your TestCafe test code?

Test Runner Script: https://codesandbox.io/s/srcset-proxying-test-58w6s7?file=/test/runTest.js

Test Script: https://codesandbox.io/s/srcset-proxying-test-58w6s7?file=/test/imgTest.js

### Your complete configuration file

None other than what is specified in the Sandbox.

### Your complete test report

_No response_

### Screenshots

_No response_

### Steps to Reproduce

1. Run the Testcafe script that is provided in the Sandbox. I'm not sure if that is possible within the Sandox. You may have to paste the `runTest.js` and `imgTest.js` to local files.
2. The test will open the Sandbox and pause, showing one failed and one successfully loaded image.

### TestCafe version

2.6.2

### Node.js version

16.14.2

### Command-line arguments

node ./test/runTest.js

### Browser name(s) and version(s)

Chrome 116.0.5845.96

### Platform(s) and version(s)

macOS 13.4.1

### Other

I can't use Testcafe's new Native Automation because we run the tests in CI in Browserstack and in Chrome, Edge, Firefox and Safari.

gg-kialo:
Some frameworks, like React, use non-canonicalised attribute names. The spec and browsers allow this, but hammerhead is only checking if the attribute name without canonicalisation matches `srcset`.

By using the `loweredAttr` for the comparison, hammerhead will be more accepting of inputs and resolve srcset issues with React projects (fixes #2958).

See https://github.com/facebook/react/issues/10863 for more background.

## Purpose

Fixes #2958 to better support `srcSet` attributes in React applications.

## Approach

Canonicalises the attribute name before comparing to `srcset`.

## References
Related react issue (and there are various duplicates of this) - https://github.com/facebook/react/issues/10863

## Pre-Merge TODO
- [ ] Make sure that existing tests do not fail (unable to get the tests to run)

### DIFF
--- a/src/client/sandbox/node/element.ts
+++ b/src/client/sandbox/node/element.ts
@@ -152,7 +152,7 @@ export default class ElementSandbox extends SandboxBase {
         const tagName     = domUtils.getTagName(el);
         const isUrlAttr   = domProcessor.isUrlAttr(el, attr, ns);
         const isEventAttr = domProcessor.EVENTS.indexOf(attr) !== -1;
-        const isUrlsSet   = attr === 'srcset';
+        const isUrlsSet   = loweredAttr === 'srcset';
 
         let needToCallTargetChanged = false;
         let needToRecalcHref        = false;
--- a/test/client/fixtures/sandbox/node/methods-test.js
+++ b/test/client/fixtures/sandbox/node/methods-test.js
@@ -243,6 +243,16 @@ test('setAttribute: img src', function () {
     strictEqual(nativeMethods.getAttribute.call(img, 'src'), urlUtils.resolveUrlAsDest('/image.gif?param=value'));
 });
 
+test('setAttribute: img srcSet', function () {
+    var img = nativeMethods.createElement.call(document, 'img');
+
+    processDomMeth(img);
+
+    img.setAttribute('srcSet', '/image.gif?param=value 1x, /imagex2.gif?param=value 2x');
+
+    strictEqual(nativeMethods.getAttribute.call(img, 'srcset'), urlUtils.resolveUrlAsDest('/image.gif?param=value 1x, /imagex2.gif?param=value 2x', true));
+});
+
 test('canvasRenderingContext2D.drawImage', function () {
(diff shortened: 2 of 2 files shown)


gg-kialo:
fyi: both locally and in github codespaces I've been unable to get the tests to run. After running `gulp test-client` nothing happens after QUnit server starts to listen.

```
...
[10:41:29] Finished 'lint' after 12 s
[10:41:29] Finished 'build' after 17 s
[10:41:29] Starting 'test-client-run'...
QUnit server listens on http://localhost:2000
```

Aleksey28 (changes requested):
Hi @gg-kialo,

Thank you for your contribution. We really appreciate your help. To preserve this fix from future changes, please add a test that checks this behavior.

gg-kialo:
I've pushed with an added test. I managed to figure how to get them to run locally (you have to open the qunit server :facepalm:).

fyi, I'll be OoO until Tuesday.

Aleksey28 (changes requested):
Thank you for your contribution. Please take a look at my suggestions.

Aleksey28 (approved):

PR #2976, opened 2023-11-13
fixed: Sockets hanging (closes #7097)

PavelMor25:
## Purpose
Due to changes in Node.js, the old approach of destroying the request is no longer suitable, as the check within the method itself determines that it has already been destroyed.

## Approach
We have added a `closeConnectionOnError` method in `RequestPipelineContext` that bypasses this check.

## References
[issue 7097](https://github.com/DevExpress/testcafe/issues/7097)

## Pre-Merge TODO


AlexKamaev (approved):

src/request-pipeline/context/index.ts
```diff
@@ -426,6 +426,11 @@ export default class RequestPipelineContext extends BaseRequestPipelineContext {
         this.goToNextStage = false;
     }
 
+    closeConnectionOnError (): void {
+        if (this.req.socket)
```
aleks-pro, line 430: We discussed in private and decided to add `req.destroy()` calling in case of some changes in future node js version

src/request-pipeline/context/index.ts
```diff
@@ -426,6 +426,15 @@ export default class RequestPipelineContext extends BaseRequestPipelineContext {
         this.goToNextStage = false;
     }
 
+    closeConnectionOnError (): void {
+        this.req.destroy();
+
+        // For Node versions greater than 15.4, we check the socket. If req.destroy didn't destroy it, we call the corresponding method of the socket.
+        if (!this.req.socket?.destroyed)
+            this.req.socket.destroy();
```
aleks-pro, line 434: What if `this.req.socket` expression returns `null` or `undefined`? It looks like we will get null ref exception

src/request-pipeline/context/index.ts
```diff
@@ -426,6 +426,15 @@ export default class RequestPipelineContext extends BaseRequestPipelineContext {
         this.goToNextStage = false;
     }
 
+    closeConnectionOnError (): void {
+        this.req.destroy();
+
+        // For Node versions greater than 15.4, we check the socket. If req.destroy didn't destroy it, we call the corresponding method of the socket.
```
aleks-pro, line 432:
```suggestion
        // NOTE: For Node versions greater than 15.4, we check the socket. If req.destroy didn't destroy it, we call the corresponding method of the socket.
```

aleks-pro (changes requested):

aleks-pro (approved):

PR #2979, opened 2023-11-20
fix: processing script with private identifier (closes DevExpress/testcafe#8067)

PavelMor25:
Added new condition to `property-get transformer` which excludes from transform the expression "object.#prop" into "__get$(object, '#prop')"

`ts-ignore` was added because updating `@types/estree` is required; the last update was three years ago. Updating now would require time for refactoring.

[esotope-hammerhead PR](https://github.com/DevExpress/esotope-hammerhead/pull/28)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "bowser": "1.6.0",
     "crypto-md5": "^1.0.0",
     "debug": "4.3.1",
-    "esotope-hammerhead": "0.6.6",
+    "esotope-hammerhead": "https://github.com/PavelMor25/builds/raw/master/esotope-hammerhead-0.6.7.tgz",
     "http-cache-semantics": "^4.1.0",
     "httpntlm": "^1.8.10",
     "iconv-lite": "0.5.1",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "bowser": "1.6.0",
     "crypto-md5": "^1.0.0",
     "debug": "4.3.1",
-    "esotope-hammerhead": "https://github.com/PavelMor25/builds/raw/master/esotope-hammerhead-0.6.7.tgz",
+    "esotope-hammerhead": "0.6.7",
     "http-cache-semantics": "^4.1.0",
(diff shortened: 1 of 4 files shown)


AlexKamaev (approved):

package.json
```diff
@@ -21,7 +21,7 @@
     "bowser": "1.6.0",
     "crypto-md5": "^1.0.0",
     "debug": "4.3.1",
-    "esotope-hammerhead": "0.6.6",
+    "esotope-hammerhead": "https://github.com/PavelMor25/builds/raw/master/esotope-hammerhead-0.6.7.tgz",
```
Aleksey28, line 24: As I see your PR in `esotope-hammerhead` has already been merged, so you can publish a new version `esotope-hammerhead`, and set it here.

src/processing/script/transformers/property-get.ts
```diff
@@ -32,6 +32,11 @@ const transformer: Transformer<MemberExpression> = {
         if (node.property.type === Syntax.Identifier && !shouldInstrumentProperty(node.property.name))
             return false;
 
+        // Skip: object.#prop
+        // @ts-ignore
+        if (node.property.type === Syntax.PrivateIdentifier)
```
Aleksey28, line 37: Have you tried to update  `estree`? Perhaps there won't be any problems.

Aleksey28 (changes requested):
Please, add a link to the corresponding PR in `esotope-hammerhead`

src/processing/script/transformers/property-get.ts
PavelMor25: Yes, I have tried, and this leads to numerous conflicts in TypeScript

Aleksey28: Okay, let's remove comment `// Skip: object.#prop` or refactor it in a way: `NOTE: <describing case>`

Aleksey28 (approved):

AlexKamaev (approved):

PR #3033, opened 2024-12-09
fix(processing/script/transformers/location): handle class property definitions

sorin-davidoi:
## Purpose
The transformers generates wrong (and invalid) JavaScript if a class contains a property named `locations`.

The generated code for `class C { location }` currently is `class C { __get$Loc(location) }`.

## Approach
Extend the transformer to handle class property definitions in the same way other processing exceptions are handled.

## Pre-Merge TODO

### DIFF
--- a/src/processing/script/transformers/location-get.ts
+++ b/src/processing/script/transformers/location-get.ts
@@ -68,6 +68,11 @@ const transformer: Transformer<Identifier> = {
         if (parent.type === Syntax.MethodDefinition)
             return false;
 
+        // Skip: class X { location }
+        // @ts-ignore
+        if (parent.type === Syntax.PropertyDefinition)
+            return false;
+
         // Skip: class location { x () {} }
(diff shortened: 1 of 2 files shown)


src/processing/script/transformers/location-get.ts
```diff
@@ -68,6 +68,11 @@ const transformer: Transformer<Identifier> = {
         if (parent.type === Syntax.MethodDefinition)
             return false;
 
+        // Skip: class X { location }
+        // @ts-ignore
+        if (parent.type === Syntax.PropertyDefinition)
```
sorin-davidoi, line 73: I'm happy to add this to `esotope-hammerhead` if there is consensus on merging this.

PavelMor25:
Hello @sorin-davidoi,
 
We sincerely appreciate your contribution and the effort you put into it! Thank you for your contribution!
 
Regarding adding this to `esotope-hammerhead`, we suggest not making changes there and keeping them limited to this project instead.

PR #3078, opened 2026-03-27
fix: support postMessage options overload and handle transferable clone failures in Firefox

### LINKED ISSUE #3077: postMessage with MessagePort transfer fails in Firefox with DataCloneError
## Description

When an application running under Hammerhead's proxy calls `postMessage` with `MessagePort` objects in the transfer list, Firefox throws:

```
DataCloneError: The object could not be cloned.
```

or:

```
MessagePort object could not be cloned.
```

This happens because `MessageSandbox.postMessage` wraps the original message in a Hammerhead envelope object before forwarding it via `fastApply`. The structured clone algorithm in Firefox cannot reconcile the wrapped envelope with the transferable `MessagePort` objects in the transfer list — the ports reference the original message, not the wrapped one.

Additionally, the `postMessage(message, { targetOrigin, transfer })` overload (the modern options-based API) is not supported. When the second argument is an object, the call is silently dropped.

## Steps to Reproduce

1. Create a page with an iframe
2. Set up a `MessageChannel` between the parent and iframe
3. Use `postMessage(message, targetOrigin, [channel.port1])` or the modern `postMessage(message, { targetOrigin: '*', transfer: [channel.port1] })` to transfer a `MessagePort` to the iframe
4. Run this under TestCafe using Firefox (not native automation)

## Expected Behavior

The `MessagePort` is successfully transferred to the iframe, and both sides can communicate over the channel.

## Actual Behavior

Firefox throws `DataCloneError: MessagePort object could not be cloned.` The message is never delivered and the communication channel is broken.

This works correctly in Chrome because TestCafe uses native automation (CDP) for Chrome, which bypasses Hammerhead's proxy entirely.

## Environment

- **TestCafe version**: 3.7.2
- **testcafe-hammerhead**: 31.7.7
- **Browser**: Firefox (ESR and stable)
- **OS**: macOS, Linux (Docker)

## Root Cause Analysis

In `MessageSandbox.postMessage`, the method wraps the user message in a `{ message, originUrl, targetUrl, type }` envelope before calling `fastApply`. When the transfer list includes `MessagePort` objects, Firefox's structured clone fails because:

1. The transferable ports are associated with the original message, not the envelope
2. The envelope object cannot be structured-cloned alongside the original transferables

Additionally, when `postMessage(message, options)` is called with an object second argument, the code treats it as a non-string targetOrigin and returns `null`, silently dropping the call.

crisryantan:
Closes #3077

## Problem

`MessageSandbox.postMessage` did not support the modern `postMessage(message, { targetOrigin, transfer })` overload.
When the second argument was an object, the call was treated as invalid and returned early, so messages were not delivered.

This also caused issues for `MessageChannel`-based communication in Firefox, because the options object (which may include `transfer`) was being handled as `targetUrl` instead of a proper overload shape.

### 1. Support the `postMessage(message, options)` overload

`postMessage` now detects when the second argument is an object and delegates to `_postMessageWithOptionsOverload`.

### 2. Handle options and transferables using native semantics

`_postMessageWithOptionsOverload`:
- extracts `targetOrigin` from options (falls back to current origin when missing),
- wraps only the user message with Hammerhead metadata,
- forwards options with `targetOrigin: '*'` so transferable handling remains native.

### 3. Keep legacy signature behavior intact

The existing wrapped flow for `postMessage(message, targetOrigin, transfer?)` remains in place.

Per maintainer feedback, the additional clone-error fallback path and related receive-side handling were removed.

## Tests

- Added test for `postMessage(message, { targetOrigin })` options overload.
- Added test for `postMessage(message, { targetOrigin, transfer })` with `MessagePort`.
- Updated the existing object-second-argument test to verify message delivery via the options-overload path.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/client/sandbox/event/message.ts
+++ b/src/client/sandbox/event/message.ts
@@ -107,14 +107,22 @@ export default class MessageSandbox extends SandboxBase {
     private _onWindowMessage (e: MessageEvent, originListener): void {
         const data = MessageSandbox._getMessageData(e);
 
-        if (data.type !== MessageType.Service) {
+        if (data.type === MessageType.Service)
+            return null;
+
+        if (data.type === MessageType.User) {
             const originUrl = destLocation.get();
 
             if (data.targetUrl === '*' || destLocation.sameOriginCheck(originUrl, data.targetUrl))
                 return callEventListener(this.window, originListener, e);
+
+            return null;
         }
 
-        return null;
+        // NOTE: unwrapped messages arrive when the Hammerhead envelope could not be structured-cloned
+        // (e.g. Firefox with MessagePort transfers). The browser already validated the target origin
+        // before delivering the MessageEvent, so it is safe to pass through to the listener.
+        return callEventListener(this.window, originListener, e);
     }
 
     private static _wrapMessage (type: MessageType, message, targetUrl?: string) {
@@ -189,7 +197,10 @@ export default class MessageSandbox extends SandboxBase {
                 const target = nativeMethods.eventTargetGetter.call(this);
                 const data   = nativeMethods.messageEventDataGetter.call(this);
 
-                if (data && data.type !== MessageType.Service && isWindow(target))
+                // NOTE: only unwrap messages that carry the Hammerhead user-message envelope.
+                // Raw (unwrapped) messages — which may arrive when the envelope could not be
+                // structured-cloned alongside transferable objects — are returned as-is.
+                if (data && data.type === MessageType.User && isWindow(target))
                     return MessageSandbox._getOriginMessageData(data);
 
                 return data;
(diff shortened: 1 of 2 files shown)


Aleksey28:
Hi @crisryantan,

We appreciate your contribution. I checked your changes and came to the conclusion that it's not necessary to make an additional handling for `MessagePort`. In the `postMessage` method, it tries to wrap the entire `args[1]` with `MessageSandbox._wrapMessage` and that causes the issue. However, I see that you already fixed it. You extracted `targetOrigin` from this object and only wrapped this `targetOrigin`. So, please remove the code that handles the `MessagePort object could not be cloned` exception. Also, you can remove comments without handling that exception. Everything is clear, so no comments are required.

crisryantan:
> Hi @crisryantan,
> 
> We appreciate your contribution. I checked your changes and came to the conclusion that it's not necessary to make an additional handling for `MessagePort`. In the `postMessage` method, it tries to wrap the entire `args[1]` with `MessageSandbox._wrapMessage` and that causes the issue. However, I see that you already fixed it. You extracted `targetOrigin` from this object and only wrapped this `targetOrigin`. So, please remove the code that handles the `MessagePort object could not be cloned` exception. Also, you can remove comments without handling that exception. Everything is clear, so no comments are required.

Thanks! Can you please take a look again 🙏

Aleksey28 (changes requested):
I checked it again. Everything works correctly now on the functionality side. You can refactor the code as follows:
1. Both `_postMessageWithOptionsOverload` and `_postMessageWrapped`  wrap messages, so information about the wrapping in the name is unnecessary. The `Overload` keyword is also unnecessary because options in the name and arguments already mean that it's overloaded.
2. Move the code that gets targetUrl as a string from `args[1]` from the `postMessage` method to the  `_postMessage` method.
3. In the postMessage method, it's better to add an `if..else` block at the beginning and call `_postMessage` or `_postMessageWithOptions`, depending on the type of `args[1]`. At the end of this method, if none of `if...else` branches run, add the block with the message about the invalid targetOrigin.

crisryantan:
I appreciate your feedback @Aleksey28! I've addressed your review. Can you check again pls 🙏🏻

src/client/sandbox/event/message.ts
```diff
@@ -213,22 +213,37 @@ export default class MessageSandbox extends SandboxBase {
     }
 
     postMessage (contentWindow: Window, args) {
-        const targetUrl = args[1] || destLocation.getOriginHeader();
+        if (!args[1] || typeof args[1] === 'string')
+            return this._postMessage(contentWindow, args);
+        else if (typeof args[1] === 'object')
+            return this._postMessageWithOptions(contentWindow, args, args[1]);
```
Aleksey28:
```suggestion
            return this._postMessageWithOptions(contentWindow, args);
```

src/client/sandbox/event/message.ts
```diff
@@ -213,22 +213,37 @@ export default class MessageSandbox extends SandboxBase {
...
-            nativeMethods.consoleMeths.log(`testcafe-hammerhead: postMessage called with invalid targetOrigin; aborting call (type: ${typeof targetUrl})`);
-            return null;
-        }
+        nativeMethods.consoleMeths.log(`testcafe-hammerhead: postMessage called with invalid targetOrigin; aborting call (type: ${typeof args[1]})`);
+
+        return null;
+    }
+
+    private _postMessageWithOptions (contentWindow: Window, args, options) {
+        const resolvedTargetUrl = typeof options.targetOrigin === 'string'
```
Aleksey28:
```suggestion
    private _postMessageWithOptions (contentWindow: Window, args) {
        const options = args[1];
        const resolvedTargetUrl = typeof options.targetOrigin === 'string'
```

src/client/sandbox/event/message.ts
```diff
@@ -213,22 +213,37 @@ export default class MessageSandbox extends SandboxBase {
...
+            ? options.targetOrigin
+            : destLocation.getOriginHeader();
+
+        const originalMessage = args[0];
+
+        args[0] = MessageSandbox._wrapMessage(MessageType.User, originalMessage, resolvedTargetUrl);
```
Aleksey28:
```suggestion
        args[0] = MessageSandbox._wrapMessage(MessageType.User, args[0], resolvedTargetUrl);
```

Aleksey28:
Please rebase your PR with the main branch and fix linter issues if needed.

Aleksey28 (approved):

crisryantan:
Thanks for merging this @aleks-pro! Just wanted to check-in, is there any schedule on when is the new package bump with these updates?

aleks-pro:
Hello @crisryantan ,

Usually, we wait until enough changes have accumulated before releasing a new version. I can't provide an exact ETA, but I'd expect the next release in the next couple of weeks.

